PR #2, opened 2014-07-07
[FIX] adapt to value_date while still supporting old versions

hbrunn:


StefanRijnhart:
Hi, as someone told me recently, you are not supposed to push your feature branch to OCA/[projectt] directly. Please create a fork first and push your branch to that fork. Create pull requests from there.

PR #4, opened 2014-07-08
Add maching of invoice based on source document

bizzappdev:
Sometimes a customer references to a sale order number when paying an invoice. Now matching on this source document number is possible.


StefanRijnhart:
LGTM, thanks! :+1:

pedrobaeza:
:+1:

PR #6, opened 2014-07-08
[FIX] account_direct_debit compatibility with odoo 8.0

sergio-teruel:


bwrsandman:
I think this might be a good time to convert these models to the new API, since they are pretty small model files.
What do you think?

sergio-teruel:
You have reason.. So i will try to do it. I will search documentation about new odoo api and i will do the necessary changes to refactoring the models

rvalyi:
agree with @bwrsandman if that's okay for you.

Also guys I want to push a PR to split that module in two, ideally we do that right before telling it's v8. Will try to find time in the week end for that.

StefanRijnhart:
LGTM. Looking forward to adoption of the new API.

account_banking_payment_export/model/account_move_line.py
```diff
@@ -35,7 +35,7 @@
 class account_move_line(orm.Model):
     _inherit = 'account.move.line'
 
-    def amount_to_pay(self, cr, uid, ids, name, arg=None, context=None):
+    def _amount_to_pay(self, cr, uid, ids, name, arg=None, context=None):
         """ Return the amount still to pay regarding all the payemnt orders
```
pedrobaeza, line 39: typo error `s/payemnt/payment`

account_banking_payment_export/model/account_move_line.py
```diff
@@ -93,6 +93,6 @@ def _to_pay_search(self, cr, uid, obj, name, args, context=None):
         return [('id', 'in', map(lambda x:x[0], res))]
 
     _columns = {
-        'amount_to_pay': fields.function(amount_to_pay,
+        'amount_to_pay': fields.function(_amount_to_pay,
```
pedrobaeza, line 96: In account_payment_extension, there is the same field, and we have a lot of performance problem with this. I resolved it making this field a stored value with triggers. You can see it here:

https://github.com/OCA/account-payment/blob/7.0/account_payment_extension/account_move_line.py#L198

pedrobaeza:
Hi, @StefanRijnhart, I have made a full migration to new API in #35, including this module. Can you check it?

hbrunn:
As #35 is the more comprehensive one, I propose to focus review efforts there and close this one

PR #13, opened 2014-07-11
[IMP] only match moves if partner id matches partner ids we got

hbrunn:


StefanRijnhart:
Same as https://github.com/OCA/banking/pull/12

pedrobaeza:
:+1:

PR #18, opened 2014-07-15
[FIX] Ignore leading zeroes when looking up domestic account numbers.

### LINKED ISSUE #16: many transactions not regonized
After importing a dutch bank statements file from the ING, I got the following screen:

import_log

Many transactions are not processed because of not recognizing the correct bankaccount number. 

The system is complaining about "Statements found for unknown account 68xxx" and "Statements found for account 06xxxxxx, but no default journal was defined.

This is a little strange, because I have setup "Default Import Settings for Bank Accounts"
in the settings of the accounting section.

I also put a default journal in there.

My information entered there is the IBAN number of my account, together with the national account (with leading zero). I also tried without the leading zero in the national account, but this doesn't make any difference at all.

Why is openerp mixing up the number of the bankaccount. (with/without zero) and not matching the national number.

StefanRijnhart:
Fixes #16

### DIFF
--- a/account_banking/res_partner_bank.py
+++ b/account_banking/res_partner_bank.py
@@ -44,6 +44,27 @@ class ResPartnerBank(orm.Model):
             '''Flag an arg as term or otherwise'''
             return isinstance(arg, (list, tuple)) and len(arg) == 3
 
+        def search_domestic(account):
+            """
+            Performing raw sql search for full account numbers,
+            ignoring leading zeroes.
+            """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1: 
Not very friend of raw SQL, but needed for performance...

StefanRijnhart:
Submitter of #16 tested succesfully https://github.com/OCA/banking/issues/16#issuecomment-49027238

hbrunn:
:+1:

PR #19, opened 2014-07-15
7.0 debit usability

StefanRijnhart:
The first commit fixes the following issues with SEPA mandates:
- the mandate field is not always visible on the payment lines on debit orders
- the mandate is not copied from the invoice because of recent client changes (purging default_\* keys in form view)
- the mandate is not copied from the invoice when creating a debit order from the payment order menu item instead of the debit order menu order. Make it impossible to do so.
- payment lines can be edited from the mandate view.

The second commit is just a PEP8 refactoring of the affected modules

The third commit sets the admin language to English during testing.


pedrobaeza:
One question, Stefan: This excludes refunds? I'm not sure about that part

StefanRijnhart:
Well, I did not change anyhing in that respect. But I think I support the filter on out_invoice. I can't really imagine that you would apply direct debit to supplier refunds at this point.

hbrunn:
:+1:

pedrobaeza:
In the migration to 8.0, I can confirm that this doesn't cover correctly refunds, but it's not a question of this PR. For this, :+1: 

@StefanRijnhart, please rebase so that we can make the merge.

hbrunn:
could you merge from upstream?

pedrobaeza:
@StefanRijnhart, can you please rebase to make the merge?

StefanRijnhart:
New attempt here: https://github.com/OCA/bank-payment/pull/67

PR #20, opened 2014-07-25
[ADD] backport sepa direct debit

hbrunn:
This is a backport of account_banking_pain_base, account_banking_sepa_direct_debit and account_banking_payment_export. Creating mandates and exporting payment order to PAIN works, but up to now I couldn't look into possible side effects of installing account_banking_payment_export


StefanRijnhart:
Thanks!

Can your debit orders make the full reconciliation workflow? It looks like you may get your orders in done state after sending them out, because of line 427 of this diff. You should try looking at line 844 of
 https://code.launchpad.net/~acsone-openerp/banking-addons/ba-70-payment-export-refactoring/+merge/179543 and see if you need this change in the workflow transition.

hbrunn:
Thanks for the hint, this transition was indeed necessary

max3903:
:+1:

bwrsandman:
:+1:

account_banking_pain_base/banking_export_pain.py
```diff
@@ -0,0 +1,439 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm
+from openerp.tools.translate import _
+from openerp.tools.safe_eval import safe_eval
+from datetime import datetime
+from unidecode import unidecode
```
pedrobaeza: This should be enclosed in a try to avoid a crash in server startup if library is not present, or move it to the method where is used (and thus only used if module is installed).

account_banking_payment_export/migrations/7.0.0.1.165/pre-migration.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: This is not needed in 6.1

account_banking_sepa_direct_debit/account_banking_sdd.py
```diff
@@ -0,0 +1,440 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm, fields
+from openerp.tools.translate import _
+from openerp.addons.decimal_precision import decimal_precision as dp
+from unidecode import unidecode
```
pedrobaeza: The same as above for avoiding server crash

account_banking_sepa_direct_debit/account_banking_sdd.py
```diff
@@ -0,0 +1,440 @@
...
+            ], 'State', readonly=True),
+    }
+
+    _defaults = {
+        'state': 'draft',
+    }
+
+
+class sdd_mandate(orm.Model):
+    '''SEPA Direct Debit Mandate'''
```
pedrobaeza, line 102: For v7 it has been developed an independent mandate module, but maybe it's too much work for this old version.

pedrobaeza:
Some issues, but the rest LGTM.

account_banking_pain_base/banking_export_pain.py
StefanRijnhart: Isn't this what external_dependencies is for in **openerp**.py?

StefanRijnhart:
Thanks for taking my comments into account! :+1:

account_banking_payment_export/migrations/7.0.0.1.165/pre-migration.py
StefanRijnhart: I don't understand your comment. Isn't this just a Python best practice, not Odoo specific?

--edit-- Look at PEP 263

account_banking_pain_base/banking_export_pain.py
pedrobaeza: Not exactly, because import in headers are executed at server initialization time, avoiding to start the server if the library is not present. With this trick, you allow to start server, but don't let the module install with the declaration in `external_dependencies`.

account_banking_payment_export/migrations/7.0.0.1.165/pre-migration.py
pedrobaeza: I'm refering the whole file. Sorry for not being too much clear. As you can see, the migration script is for 7.0 version.

account_banking_pain_base/banking_export_pain.py
StefanRijnhart: I believe you are correct. It did not use to be the case but I think the problem is that the web module triggers the loading of every module in the addons path with a static folder in it (if only containing a logo icon, like this module).

account_banking_payment_export/migrations/7.0.0.1.165/pre-migration.py
StefanRijnhart: Sorry, missed that completely. Well spotted then, and :+1: for removing this file.

hbrunn:
thanks @pedrobaeza for your input! But indeed I'd like to skip backporting the mandate module to 6.1, as this is only meant enable one customer to do sepa direct debits, and they will upgrade soon, so from my side, I'd like to leave that as it is

pedrobaeza:
Yeah, it's not worth the effort. Please fix last comments and I merge this.

hbrunn:
I did with the last commit, or am I missing something?

pedrobaeza:
Sorry, I didn't see it. Let's merge then!

And thanks for all.

PR #35, opened 2014-09-10
[MIG] Migration and enhancement of all modules involved in SEPA

pedrobaeza:
This is a full migration of the modules involved in SEPA, most of it with the new API. It's a huge diff and I have made file name changes according last conventions, so GitHub does also a mess with diff. It has been tested in an environment for a week, so it's quite stable.


michaelkarrer81:
Thank you very much!

sbidoul:
@pedrobaeza in account_banking_payment_export, there seem to be duplicate code in the model and models directories. At least payment_mode.py, payment_mode_type.py, account_move_line.py, account_payment.py.

Can you check?

pedrobaeza:
@sbidoul Sorry, it was a kind of mess with git. I have removed them.

sbidoul:
Thanks!

I'm having issue with the manual payment mode. When I click on "Make Payment", then "Ok" in the wizard, I get "TypeError: create() takes exactly 2 arguments (5 given)".

Restoring the old api create() in bank_payment_manual.py seems to help. Not sure if the fault lies in the migrated code or in an odoo bug.

sbidoul:
@pedrobaeza so in my test, the new API version of bank_payment_manual does not work. Can you reproduce on your side?

pedrobaeza:
I'm going to try and fix it. I had to leave some methods with old API in several models because of the mix of new and old. Maybe it's the same case.

pedrobaeza:
Hi, @sbidoul,

I have fixed the problem, cleaning also the code, because the used method was not the best way. As I only made blindly the port, I didn't see to this optimizations.

sbidoul:
Hi @pedrobaeza,

Two other things I discovered in my tests today:
- calls to netsvc.LocalService('workflow') are deprecated. Can you replace with openerp import workflow?
- it seems there are still duplicate files in account_banking_pain_base, account_sepa_credit_transfer, account_sepa_direct_debit

sbidoul:
@pedrobaeza there is a pylint error in account_direct_debit/models/account_move_line.py:30 which makes Travis Fail

sbidoul:
For the record, the branch with the refactoring of account_banking_payment is here https://github.com/acsone/banking/tree/8.0-refactor-account_banking_payment-sbi

It is build on top of this PR and includes the port of account_banking_tests.

alexis-via:
Pedro,

Thanks for your great work. In banking/account_banking_sepa_direct_debit, you moved the python files to "models/", but the old python files are still present in account_banking_sepa_direct_debit/, which is a big trap (I just fall in this trap :)

Another suggestion (which may be for another PR) : we should move the readonly=True on the field unique_mandate_reference (of the object sdd.mandate) from the field definition to the form view, in order to make it possible to import existing mandates with RUMs (for companies who migrate from another system to OpenERP and have to migrate existing mandates in OpenERP and keep the RUMs)

pedrobaeza:
Thanks for the input, @alexis-via. I'm resolving now the duplicated files issue, and addressing @sbidoul. I'll take care also about your input.

pedrobaeza:
@alexis-via, I'm going to integrate also the 7.0 work of splitting mandates as generic ones, but concerning your issue, I have imported mandates via XML-RPC specifying a mandate number and I don't have any problem. Also you can set this field via ORM, because readonly argument is only considered as a default value for views, and also on onchanges, but not on direct creations or modifications. Have you tried before this?

alexis-via:
@pedrobaeza 
From my experience of this afternoon with Odoo v8, when the field is readonly=True in it's definition, you can't import it via the CSV import feature of the Web interface (via the module base_import).

alexis-via:
@pedrobaeza 
Are you sure you want to include the splitting of mandates as generic ones in this PR ? We really need to get this PR in the main 8.0 branch shortly... I hate empty 8.0 branches :)

pedrobaeza:
I'm doing the work tonight, so tomorrow it's going to be ready.

pedrobaeza:
I have made a first upload of the migration with the generic bank mandate, but I'm seeing that some design things doesn't like me too much, so maybe I will change it. If you want, you can try for now and it's fully functional.

alexis-via:
@pedrobaeza 
I have updated your branch, and I think I found an issue: I can see the module account_banking_mandate, but the module account_banking_sepa_direct_debit doesn't depend on account_banking_mandate (but it inherits the object account.banking.mandate).

pedrobaeza:
Yeah, sorry, I have just fixed.

Thanks for noticing.

alexis-via:
@pedrobaeza 
The file wizard/export_sdd.py still contains 3 occurrences of self.pool['sdd.mandate'], which make me think that the file wizard/export_sdd.py hasn't been "ported" to the use of account_banking_mandate.

pedrobaeza:
Oh, that's right. Let me a couple of hours to fix it.

alexis-via:
BY the way, I just ported account_direct_debit/wizard/payment_order_create.py to the new API and re-enabled it in account_direct_debit/**init**.py

I am now trying to fix other things (error when adding lines in direct debit order) and I'll send you the patch with all the fixes in a few hours.

alexis-via:
@pedrobaeza 
In addition to the mail I sent you, I found another bug: in account_banking_mandate/models/payment_line.py on line 57: if you have 2 or more mandates for the bank account, mandates.id will try to write (ID1, ID2) on the many2one field 'mandate_id' and it would fail.

So I think it should be fixed that way:

REMOVE            vals['mandate_id'] = mandates.id
ADD                    vals['mandate_id'] = mandates[0].id

pedrobaeza:
In this case, it doesn't fail, because when you access a recordset with dot notation (ActiveRecord pattern), it returns values from first record.

pedrobaeza:
I have merged some fixes from @alexis-via for some sdd.mandate old references.

alexis-via:
@pedrobaeza 
About my remark
REMOVE vals['mandate_id'] = mandates.id
ADD vals['mandate_id'] = mandates[0].id
I didn't made this change by reading the code and thinking it was a mistake, but because I made some real tests with real data and I got a crash. It's not difficult to reproduce: you have to configure 2 valid mandates on a customer and create a SEPA direct debit file : when you validate the selection of the invoices to pay, you will get a crash.

pedrobaeza:
Well, it's strange, because theory says that, but I have changed it in any case.

NL66278:
:+1:  Looks good to me. Did some testing. Entering mandates. Creating SEPA credit transfer file. The only thing so far that did not look OK, but might be a mistake on my part in configuring something, after entering mandate on partner bank account, the partner field remains empty on the mandate.

pedrobaeza:
@NL66278 Thanks for the review. I have detected and fixed the problem with that field (a typo in the argument + readonly behaviour).

pedrobaeza:
Please give the other reviewers your thumbs up to merge at last this PR.

sbidoul:
@pedrobaeza there are new flake8 errors (E123) ?

pedrobaeza:
This is due to: https://github.com/OCA/maintainer-quality-tools/issues/115, but I have fixed it anyway.

sbidoul:
:+1: 

Thanks for this effort, Pedro.

alexis-via:
@pedrobaeza 

Two more important remarks:
1) In v7.0, the module 'account_payment_sale' only depended on 'sale' (and NOT on 'stock') ; there was another module 'account_payment_sale_stock' that depended on 'sale_stock'. This allowed the use of 'account_payment_sale' WITHOUT having to install the 'stock' module.
In your branch for v8, the module 'account_payment_sale' depends on 'stock_account' and you didn't port the module 'account_payment_sale_stock'. Why ? I think it is a regression compared to the v7 branch, because with your branch for v8 we can't use the module 'account_payment_sale' without the 'stock' module.

2) In account_payment_sale/models/sale_order.py and account_payment_purchase/models/purchase_order.py, in the definition of the payment_mode_id field, you added a domain that adds a filter on 'payment_order_type'. I don't think it is a good idea because, for example:
- you can't set the payment_mode_id to a 'direct debit' on a purchase.order ; it may be usefull for the user to know that this supplier invoice will be paid via direct debit, even if the "direct debit" is not managed by OpenERP
- you can't set the payment_mode_id to 'credit transfer' on a sale.order... but it would be usefull for the user to know that this customer usually pays via 'credit transfer' and display some information about this on the sale order report (for example, display the IBAN of my company related to the 'credit transfer' payment.mode, so that he knows on which bank accout he should send the money)
  So, if I didn't put a domain on the payment_mode_id field on sale.order and purchase.order when I developped the module account_payment_sale and account_payment_purchase for v7, it was on purpose !

pedrobaeza:
I answer first the point 2) because it doesn't required further investigation:

This is also done on purpose, and I think it must remain as is now, because in my experiences with the users of the system, they got very confused about payment modes that can be selected on customer invoices and sale orders, and the same for direct debit modes that can be selected on supplier invoices and purchase orders. They selected the wrong one on one invoice type, and when they try to add the invoices to a payment order, they got confused about not finding them.

I think the use case you are describing is a misuse of this module, and you should use another method for this, or overwriting in a custom module to remove this restriction.

rvalyi:
Hello, I just wanted to mention that point 1) (not depending on stock module since the start) is very important too nevertheless. It's important not interfering with Odoo modularity.

sbidoul:
@pedrobaeza @alexis-via I'd suggest keeping the same behaviour as the 7.0 branch in this PR, and move the discussion about the payment_mode domain to another PR.

Regarding modularity I agree with @alexis-via and @rvalyi: account_payment_sale must not depend on stock.

pedrobaeza:
Well, I think this must be discussed before doing any merge, because I don't want to issue this first version with the usability problem I see. Spanish people is going to use for the first time this branch, letting _account_payment_extension_, and they expect a correct functionality to be convinced of the switch, so I prefer to not have complains about this.

alexis-via:
@pedrobaeza 

About point 2 (domain on payment_mode_id), I understand your point about Spanish users that will migrate from account_payment_extension to the banking branch. But you also have to consider the existing users of the banking branch. In fact, I didn't discover the point 2 myself, it was reported to me by one of my customers on v8 that I had recently migrated from banking/7.0 (with minor changes to make it work on v8) to the branch of this pull request ; he was surprised to see that he couldn't select "Credit transfer" as payment mode for a sale order any more (which would display the credit transfer instructions on the sale order report).

NL66278:
About the (in?) famous point 2, the domain on payment mode. I can understand Pedro from a functional perspective. We encountered the same questions as his Spanish customer. But maybe these restrictions can be put in an optional module, so we get a clear separation between migrating the 7.0 functionality and having new (and optional) functionality.

sbidoul:
@pedrobaeza @alexis-via Regarding point 2, both point of views look valid and debit|payment is not a valid filter to cover for all use cases. Perhaps should we add a new field on payment mode to let the user select explicitly to what type of transaction the payment mode applies (sale|purchase|both, or even something finer grained taking refunds into account)?

sbidoul:
@pedrobaeza there is another duplicate file to be removed: account_payment_purchase/view/purchase.xml

alexis-via:
@sbidoul 

+1 for your idea of being able to configure on payment.mode the type of transaction on which it can apply.

NL66278:
+1 as well for the proposal from @sbidoul. For me a good alternative for a separate module.

pedrobaeza:
That's a very good idea, but I haven't got time to do it. I will do it ASAP.

alexis-via:
@pedrobaeza 
If you don't have time to do it, I can try to do it tonight if you want.

pedrobaeza:
I'll try to do it today, and I tell you if I don't. Thanks for your help.

alexis-via:
I don't know if others experience this issue too: the field partner_id on account.banking.mandate is always empty. I'll try to investigate this when I have time...

NL66278:
Hello Alexis, according to Pedro this was fixed 11 days ago. Did you pull the latest code?

alexis-via:
@NL66278 My code was up-to-date, but as this partner_id related field on account.banking.mandate is store=True, it didn't solve the issue. I dropped the column and reloaded the account_banking_module, and now the partner_id field has the right value. Sorry for the noise.

pedrobaeza:
OK, I have worked today on the remaining issues:
- I have splitted again stock part of the sale process in the _account_payment_sale_stock_ module.
- I have implemented the three-valued selection for payment, debit or both export types, and it works good. Thanks @sbidoul for the suggestion.
  *I have added translation template files for all modules.
- And as an extra, I have translated all modules to spanish.

Please review and give this the final impulse to get it merged at last.

sbidoul:
Hi @pedrobaeza,

Thanks for this work!

I only had a quick look so far, but I'm pretty sure it won't work like that. Most modules expect payment_mode_type.payment_order_type to be either 'payment' or 'debit' but nothing else, so they will break.

What I had in mind is a field named 'applies_to' on _payment_mode_ (not payment_mode_type) with possible values 'sale|purchase|both'. Or even more flexible, if we want to take refunds into accounts, a set of 4 flags: applies_to_sale, applies_to_sale_refund, applies_to_purchase, applies_to_purchase_refund.

pedrobaeza:
Hi, @sbidoul, the way you propose will duplicate fields for the same purpose, because payment_mode_type is only used to filter invoices on payment orders and direct debit orders, and thus there is no impact in adding this value. Or can you point me to any other use or code that it's dependant of this?

sbidoul:
@pedrobaeza in my understanding, payment_order_type must have the same value on the payment.order, payment.mode and payment.mode.type.

See for example this filter: https://github.com/OCA/banking/pull/35/files#diff-b5d41850b1a9531fb0e8567b9c3066b2R48

To me this field really defines what the payment mode type is: a payment or a direct debit. It cannot be both at the same time.

That's why I suggested in comment https://github.com/OCA/banking/pull/35#issuecomment-60512413 that we add a new field on payment mode, to cover the use case of @alexis-via (ie saying that a sale order will be paid via credit transfer by the customer).

pedrobaeza:
I insist that the field is only used in that way, so I see no need to duplicate fields. Payment mode type only means where it applies, and it's not the same as the type of the payment orders, that can only be payment or direct debit.

But let's see what other says. @alexis-via, any opinion?

alexis-via:
I am very surprised by the code in account_payment_sale_stock/models/stock_picking.py: @pedrobaeza , you write:

```
    @api.model
    def _create_invoice_from_picking(self, picking, vals):
        if picking:
            # Search if this picking comes from a sale order
            sale_order_obj = self.env['sale.order']
            sale_order = sale_order_obj.search(
                [('picking_ids', 'in', picking.id)], limit=1)
            if sale_order and sale_order.payment_mode_id:
                vals['partner_bank_id'] = sale_order.payment_mode_id.bank_id.id
                vals['payment_mode_id'] = sale_order.payment_mode_id.id
        return super(StockPicking, self)._create_invoice_from_picking(picking,
                                                                      vals)
```

I don't understand why you don't use the "sale_id" M2O field on the picking ? I know that it's now a computed field on v8, but the O2M field 'picking_ids' on sale.order is also a computed field...

```
    @api.model
    def _create_invoice_from_picking(self, picking, vals):
        if picking and picking.sale_id:
            sale_order = picking.sale_id
            if sale_order.payment_mode_id:
                vals['partner_bank_id'] = sale_order.payment_mode_id.bank_id.id
                vals['payment_mode_id'] = sale_order.payment_mode_id.id
        return super(StockPicking, self)._create_invoice_from_picking(picking,
                                                                      vals)
```

alexis-via:
@pedrobaeza 
I have the same point of view as @sbidoul : for me, payment_order_type must have the same value on the payment.order and payment.mode.

To make things simple, I would just add 2 boolean fields on payment.mode:
- sale_ok
- purchase_ok
  and filter on these fields in the domain of the field payment_mode_id (just like for the product).

pedrobaeza:
@alexis-via, I don't know how I ended with that code. I supposed it was too late on night :stuck_out_tongue_winking_eye: I have changed it according your comments.

@sbidoul and @alexis-via, I have implemented sale/purchase duality mechanism the way you tell me. Please check it and let's move this PR to merge finally this that it is being demanded a lot here in Spain.

alexis-via:
:+1: Thanks pedro for your work.

pedrobaeza:
Rebased

pedrobaeza:
@sbidoul, do you agree with the last changes to merge the PR?

sbidoul:
@pedrobaeza working on this is on my todo for this afternoon. Sorry for the delay.

sbidoul:
Ok, I did some quick tests again. This LGTM, now. Thanks @pedrobaeza for this work!

:+1:

pedrobaeza:
OK, I proceed with the merge, and we can fix afterwards anything more we find via PR.

PR #37, opened 2014-09-17
[ADD] account banking sepa direct debit create mandates

hbrunn:
## Introduction

When switching to SEPA payments, the mandates to be created can be a
serious show stopper. Usually, they are available in one form or the other,
but not in OpenERP. This plugin adds a button on payment orders to create
mandates for the current payment order where necessary.
## Configuration

By default, this module creates one-off mandates. If you rather want to create
recurring mandates (i.e. because your available mandates are recurring ones),
set the config parameter
'account.banking.sepa.direct.debit.create.mandates' to 'recurrent'.
## Attention

Only use this if you actually have the mandates!

### DIFF
--- a/account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
+++ b/account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
@@ -36,9 +36,9 @@ mandates for the current payment order where necessary.
 Configuration
 -------------
 
-By default, this module creates one-off mandates, which seems to be the safe
-choice. If this module is to be used to slowly evolve towards using recurring
-mandates, set the config parameter
+By default, this module creates one-off mandates. If you rather want to create
+recurring mandates (i.e. because your available mandates are recurring ones),
+set the config parameter
 'account.banking.sepa.direct.debit.create.mandates' to 'recurrent'.
 
 Attention
--- a/account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
+++ b/account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2014 Therp BV (<http://therp.nl>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Create mandates for payment orders",
+    "version": "1.0",
+    "author": "Therp BV",
+    "license": "AGPL-3",
+    "complexity": "normal",
+    "description": """
+Introduction
+------------
+
+When switching to SEPA payments, the mandates to be created can be a
+serious show stopper. Usually, they are available in one form or the other,
+but not in OpenERP. This plugin adds a button on payment orders to create
+mandates for the current payment order where necessary.
+
+Configuration
+-------------
+
+By default, this module creates one-off mandates, which seems to be the safe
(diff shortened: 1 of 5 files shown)


account_banking_sepa_direct_debit_create_mandates/__openerp__.py
```diff
@@ -0,0 +1,62 @@
...
+    "name": "Create mandates for payment orders",
+    "version": "1.0",
+    "author": "Therp BV",
+    "license": "AGPL-3",
+    "complexity": "normal",
+    "description": """
+Introduction
+------------
+
+When stepping over to SEPA payments, the mandates to be created can be a
```
StefanRijnhart: s/stepping over to/switching to/

account_banking_sepa_direct_debit_create_mandates/model/payment_order.py
```diff
@@ -0,0 +1,65 @@
...
+    _inherit = 'payment.order'
+
+    def _is_missing_mandates_get(self, cr, uid, ids, field_name, arg, context=None):
+        result = {}
+        for this in self.browse(cr, uid, ids, context=context):
+            if this.payment_order_type != 'debit':
+                result[this.id] = False
+                continue
+            result[this.id] = any(
+                [not bool(l.sdd_mandate_id) for l in this.line_ids])
```
StefanRijnhart: Is the explicit conversion to bool necessary?

account_banking_sepa_direct_debit_create_mandates/model/payment_order.py
```diff
@@ -0,0 +1,65 @@
...
+        sdd_mandate = self.pool['sdd.mandate']
+        for this in self.browse(cr, uid, ids, context=context):
+            for line in this.line_ids:
+                if line.sdd_mandate_id:
+                    continue
+                mandate_id = sdd_mandate.create(
+                    cr, uid,
+                    {
+                        'type': 'oneoff',
+                        'sepa_migrated': True,
```
StefanRijnhart: The choice for 'oneoff' seems to me customer specific. Can you make the choice between recurring and one off a system parameter?

pedrobaeza:
Maybe _account_banking_sepa_direct_debit_on_the_fly_ can be a better name?

BTW, the other day we fixed in Odoo the problem with 64 chars in image path that prevents to install modules with long names like this (SQL sentence).

hbrunn:
I added Stefan's suggestions. As for the name: 'On the fly' doesn't make it clear to me what the module does, 'create mandates' does. Do you agree?

pedrobaeza:
Sorry, I must say _account_banking_sepa_direct_debit_mandate_on_the_fly_, but both are not specific enough and the name is so long...

hbrunn:
okay, this time I don't have strong feelings about the module's name and did as you said

hbrunn:
btw. also rebased and squashed the commits

StefanRijnhart:
You damn Git history rewriters. I think your squashing made GitHub lose my synchronous new inline code comment...

Thanks for the changes so far btw. Just not done nagging.

account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
```diff
@@ -0,0 +1,70 @@
...
+mandates for the current payment order where necessary.
+
+Configuration
+-------------
+
+By default, this module creates one-off mandates, which seems to be the safe
+choice. If this module is to be used to slowly evolve towards using recurring
+mandates, set the config parameter
+'account.banking.sepa.direct.debit.create.mandates' to 'recurrent'.
+
```
StefanRijnhart, line 43: Can you make this more neutral? Converting one-off mandates is not safer than converting recurrent mandates, if I understand correctly. The mandate type is dictated by the type of the converted domestic mandate.
In the same vein, 'slowly evolving towards using recurring mandates' is misleading, as the domestic mandates should be recurrent in the first place to migrate them to recurrent SDD mandates.

hbrunn:
Done. Even though I consider one-off mandates still safer because if I mandate you to debit n times, it follows that I mandate you to debit 1 time, but not the other way around...

account_banking_sepa_direct_debit_mandate_on_the_fly/__openerp__.py
```diff
@@ -0,0 +1,70 @@
...
+    "license": "AGPL-3",
+    "complexity": "normal",
+    "description": """
+Introduction
+------------
+
+When switching to SEPA payments, the mandates to be created can be a
+serious show stopper. Usually, they are available in one form or the other,
+but not in OpenERP. This plugin adds a button on payment orders to create
+mandates for the current payment order where necessary.
```
pedrobaeza, line 34: Better put `needed` instead of `necessary`, don't you think?

pedrobaeza:
A little remark, but the rest :+1:

StefanRijnhart:
Thanks! :+1:

PR #43, opened 2014-10-07
[7.0]  Add account payment blocking

adrienpeiffer:
Prevent invoices under litigation to be proposed in payment orders.
This module uses the 'blocked' flag that is present on move lines, to filter out lines proposed in payment orders. In addition it exposes this flag on the supplier invoice form so it is easier to block an invoice.


account_banking_payment_blocking/model/account_invoice.py
```diff
@@ -0,0 +1,71 @@
...
+                assert len(move_line_ids) == 1
+                move_line = move_line_obj.browse(cr, uid, move_line_ids,
+                                                 context=context)[0]
+                res[invoice.id] = move_line.blocked
+            else:
+                res[invoice.id] = False
+        return res
+
+    _columns = {
+        'move_blocked': fields.function(_get_move_blocked,
```
sbidoul: Probably a matter of taste, but I would prefer to have this field simply named 'blocked'.

account_banking_payment_blocking/model/account_invoice.py
```diff
@@ -0,0 +1,71 @@
...
+                context.update({'from_parent_object': True})
+                move_line_obj.write(cr, uid, move_line_ids,
+                                    {'blocked': field_value}, context=context)
+
+    def _get_move_blocked(self, cr, uid, ids, name, arg, context=None):
+        res = {}
+        if isinstance(ids, (int, long)):
+                ids = [ids]
+        for invoice in self.browse(cr, uid, ids, context=context):
+            if invoice.move_id.id:
```
sbidoul: nitpicking, but can you move the search for the move line in a separate method to avoid duplicate code in _get and _set?

hbrunn:
:+1:

lmignon:
apart that the headers should probably be simplified :+1:

adrienpeiffer:
Anyone willing to review this one?

pedrobaeza:
There are some comments not honoured as I have seen.

Another thing is the name: banking is not needed. You can put only _account_payment_blocking_ as module name.

sbidoul:
All my comments (and other's as far as I can see) have been cared for so :+1: 

@pedrobaeza is the module name a blocking issue for you? There are many other modules starting with account_banking in this branch.

pedrobaeza:
Well, it's better to name correctly now that this is the first entry in OCA. Later, we have to deal with migration scripts and so on.

sbidoul:
Okay, I renamed.

pedrobaeza:
:+1:

sbidoul:
rebased

adrienpeiffer:
From my point of view, this one is ready to merge

pedrobaeza:
OK, we have enough reviews and all comments have been honoured, so I merge.

Thanks for the efforts.

PR #45, opened 2014-10-08
7.0 translation update

ehdem:
- Translate missing translated files
- Fix pep8 error
- Change definition of some fields.


bwrsandman:
@pedrobaeza We use lambda so the code is called on self. If a the function reference is used directly in the declaration, then, when you overwrite it, it still won't be called, since the declaration is done with the function in that same class (the ref).

pedrobaeza:
Interesting... Have you tried that inheritance work this way?

bwrsandman:
Yes this is what I do by default now in 7.0.
8.0 has the wonderful ability of accepting a string of the function name, so I assume (did not test) it uses `object.__getattribute__(function_name_str)()` which is called on the self and can be inherited.

pedrobaeza:
Cool to know!

:+1:

max3903:
:+1:

sebastienbeau:
Good trick with lambda! :+1:

PR #46, opened 2014-10-11
add default currency in case bankstatement do not have ccy tag

bizzappdev:
Sometimes there is no ccy tag in the statement. in that case parser/processing is causing problem.
so adding default currency.
We can improve it by taking Default company currency. Bu I am not able to get user or company instance at parser level .

### DIFF
--- a/account_banking_camt/camt.py
+++ b/account_banking_camt/camt.py
@@ -150,7 +150,12 @@ CAMT Format parser
                 identifier)
             )
 
-        statement.local_currency = self.xpath(node, './ns:Acct/ns:Ccy')[0].text
+        #Assuming If there is no Ccy tag then it belongs to defaut currency
+        if not self.xpath(node, './ns:Acct/ns:Ccy'):
+            statement.local_currency = 'EUR'
+        else:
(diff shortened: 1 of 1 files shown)


account_banking_camt/camt.py
```diff
@@ -150,7 +150,12 @@ def parse_Stmt(self, cr, node):
                 identifier)
             )
 
-        statement.local_currency = self.xpath(node, './ns:Acct/ns:Ccy')[0].text
+        #Assuming If there is no Ccy tag then it belongs to defaut currency
```
pedrobaeza: Put a space after `#`

account_banking_camt/camt.py
```diff
@@ -150,7 +150,12 @@ def parse_Stmt(self, cr, node):
...
+        if not self.xpath(node, './ns:Acct/ns:Ccy'):
+            statement.local_currency = 'EUR'
```
pedrobaeza: Better to use `statement.company_id.currency_id.id`

pedrobaeza:
:+1:

hbrunn:
:+1:

PR #49, opened 2014-10-20
[FIX] Module base_iban_not_required:

NL66278:
```
- Removed module from the __unported__ directory, because
  module is no longer needed on odoo 8.0, therefore does not need
  porting.
```


pedrobaeza:
Why is not needed?

NL66278:
The module removes a constraint that in odoo 8.0 is already non existent.

pedrobaeza:
OK, thanks for the explanation :+1:

hbrunn:
:+1:

PR #57, opened 2014-10-30
[NEW] 1st draf of account_due_list integration with account_payment_partner

mileo:
Hi All,

I'm following the mail: [Banking-addons-drivers] banking refactoring next step

As Raphael said, here in Brazil "everything is paid in 10x or so".

We are thinking in a way about meet the needs of account_payment_extension.

This is a 1 hour work scratch about integration of account_due_list and account_payment_partner like as account_payment_extension do with payment.type and account.move.line.

I'll be happy to receive any feedback
Thanks in advance


.gitignore
```diff
@@ -0,0 +1,56 @@
+# Byte-compiled / optimized / DLL files
```
pedrobaeza, line 1: This file already exists. Why it appears as added?

pedrobaeza:
There are some missing final lines in py files that Travis are going to complain about.

mileo:
I will review the request. Unfortunately I used the master branch as base.

PR #59, opened 2014-11-06
fix for Paid more does not set invoice to paid status #52

### LINKED ISSUE #52: [v7] Paid more does not set invoice to paid status
I have invoice of the amount of 850. On bank statement i receive a payment of 1000. Now when i confirm the bankstatement line, the invoice does not get the paid status. On the invoice is registered a paymnet of 1000. This is not correct.

The invoice should be reconcilled with the payment and the diff (150) should stay open by default on the debtors account. It should be the same as using the pay botton on the invoice.

ghost:
Resolve #52

### DIFF
--- a/account_banking/banking_import_transaction.py
+++ b/account_banking/banking_import_transaction.py
@@ -496,7 +496,19 @@ class banking_import_transaction(orm.Model):
             'account_id': transaction.move_line_id.account_id.id,
             'type': transaction.move_line_id.credit and 'dr' or 'cr',
         }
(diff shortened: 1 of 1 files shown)


gurneyalex:
can you please add an automated test for this bug fix?

StefanRijnhart:
Agree with @gurneyalex, it should be easy to add a new test to account_banking_tests based on what's there already (which tests payment orders, not single invoices).

StefanRijnhart:
That and the flake8 errors in the Travis run.

ghost:
This is not a proper Fix So I am closing this one

PR #60, opened 2014-11-06
7.0 fix translation link partner nl

ghost:
In Link partner form view, form string is not translated. so here is a fix for that for NL lang.

### DIFF
--- a/account_banking/i18n/nl.po
+++ b/account_banking/i18n/nl.po
@@ -646,6 +646,8 @@ msgstr "Factuurbetaling"
 #: view:account.bank.statement.line:0
 #: view:banking.link_partner:0
 #: model:ir.model,name:account_banking.model_banking_link_partner
+#: code:_description:0
+#, python-format
 msgid "Link partner"
 msgstr "Koppel relatie"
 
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Is the change to the po file the result of a fresh export after the change in the code, or is it added manually?

yvaucher:
@bwrsandman I see this [1] should resolve such issue but is only available on OCB any opinion on accepting such PR ?

[1]https://code.launchpad.net/~savoirfairelinux-openerp/openobject-server/7.0_mail_thread_translate_bug1262000/+merge/208812

pedrobaeza:
Change should be put on pot file also.

yvaucher:
agreed with @pedrobaeza  > need fixing

bwrsandman:
> https://code.launchpad.net/~savoirfairelinux-openerp/openobject-server/7.0_mail_thread_translate_bug1262000/+merge/208812

@yvaucher That was also merged in Odoo

yvaucher:
@bwrsandman thx for the update

so we just need the change in pot file here

StefanRijnhart:
@yvaucher I don't understand. The server fix referred to above is applied here: https://code.launchpad.net/~savoirfairelinux-openerp/openobject-addons/7.0_mail_thread_translate_bug1262000/+merge/199374. In this code, you can see that the model's _description is wrapped in a gettext call _(). Doesn't that mean that we need both changes and the current PR is good as it is?

pedrobaeza:
As this one hasn't been answered for a long time, I close.

PR #61, opened 2014-11-06
Link Partner form view Is company visibility

ghost:
In Link Partner form view when Partner is already selected or we select partner "Is Company" field is visible. It is not needed to be visible if we have already selected a partner . So here is a fix for that.


pedrobaeza:
Seems OK.

:+1:

StefanRijnhart:
Thanks! Fast-tracking this trivial change.

PR #63, opened 2014-11-06
Invoice is linked to wrong account

ghost:
resolved https://github.com/OCA/banking/issues/62

### DIFF
--- a/account_banking/banking_import_transaction.py
+++ b/account_banking/banking_import_transaction.py
@@ -296,6 +296,7 @@ class banking_import_transaction(orm.Model):
                     (convert.str2date(trans.execution_date, '%Y-%m-%d') +
                     self.payment_window)
                     and (not _cached(x) or _remaining(x))
+                    and x.account_id.type in ('receivable', 'payable')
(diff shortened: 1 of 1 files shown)


account_banking/banking_import_transaction.py
```diff
@@ -296,6 +296,7 @@ def is_zero(move_line, total):
                     (convert.str2date(trans.execution_date, '%Y-%m-%d') +
                     self.payment_window)
                     and (not _cached(x) or _remaining(x))
+                    and x.account_id.type in ('receivable', 'payable')
```
StefanRijnhart, line 299: If I am correct, 'x' is one from the 'move_lines' argument in the call to _match_invoice. These move lines have been selected by accounts that are reconcilable ( ('account_id.reconcile', '=', True) around l.924). This would include clearing accounts, which may not necessarily be payable or receivable. My question to you is, why is your other move line account reconcilable?

-- edit --: not tax account

legalsylvain:
I close, due to long inactivity. Please reopen and answer the question if needed.

PR #64, opened 2014-11-14
[REF] auto-fix conventions

moylop260:


moylop260:
@pedrobaeza 
If just relative import fix is good then you can fix it with next script:

``` bash
sudo pip install 3to2
2to3 --no-diffs -wf import FILE_PY  #  Auto-fix relative import of a file
```

Or if you want use original script of this PR but only fix this case, you can run:

``` bash
git clone https://github.com/Vauxoo/pylint-conf.git
python pylint-conf/fix_unused_imports/replace_import.py ADDONS_PATH_TO_FIX fix_relative_import  #  Auto-fix relative import of a folder path
```

FYI other wrong auto-fixes are by autoflake and autopep8 script's.

pedrobaeza:
Fix already, so I close.

PR #65, opened 2014-11-16
8.0 migration of account_banking_payment

sbidoul:
Summary of commits here:
- move account_banking_payment out of unported
- split it in two modules: account_banking_payment_transfer, and account_banking_payment_term
- drop the part of account_banking_payment that was dealing with bank statements
- use workflow triggers to move the payment orders to done when the transfer moves are reconciled
- migrate and improve the payment roundtrip test suite
- fix a bug that was causing the transfer moves to be wrongly numbered


alexis-via:
Code review :+1:
We really need to get this into the v8 branch, in particular the module account_banking_payment_transfer which is really needed to work seriously with v8.

alexis-via:
Function test on account_banking_payment_transfer :+1:
Stéphane, I sent you a small PR on your acsone branch to improve the form view of payment.mode. It may not be perfect still, but much better than the current form view in your branch.

alexis-via:
@sbidoul After more work on account_payment_transfer and more reflexions with real-life scenario, I think that the account move generated when we validate a payment order is not the best one. I am working on a PR to propose an alternative, and I'll explain it when it's ready

alexis-via:
Here are the changes I propose on the transfer account move, with some explainations and an example of the transfer account move for a SEPA Direct Debit payment.order:
https://github.com/acsone/bank-payment/pull/4

pedrobaeza:
We need one extra option for Spain: direct payment, that allows to create payment moves (as the voucher does) for the payment order. _account_payment_extension_ had this one, and it's going to be very missed, because banks don't inform via statement of the effective payment of the order.

I'm working in this option, and I'll make a PR with it.

alexis-via:
Just a note to say that a big part of the changes and the review work for account_banking_payment_transfer in now happening here:
https://github.com/acsone/bank-payment/pull/4

sbidoul:
Ok, so I just
- merged https://github.com/acsone/bank-payment/pull/4 with Alexis' improvements
- rebased
- fixed a pylint warning
- removed a field that moved to a account_banking_payment_transfer
- fixed and enriched the test suite to exercice the new transfer_move_option

So we are ready for the (last?) round of review. @alexis-via @pedrobaeza

alexis-via:
I have successfully used this branch for a SEPA credit transfer today on a v8 in production and the account move generated was OK (I faced several other issues, but the account move of the module account_banking_payment_transfer was fine). Let's go for the merge :+1:

sbidoul:
We are also running it in production since end december for SEPA credit transfer. I'm also :+1:  to merge.

Procedurally speaking we need another review. Any volunteer? @pedrobaeza? @StefanRijnhart?

StefanRijnhart:
Great effort! Sorry for neglecting this so far. I've now asked my colleague Ronald who is on banking for 8.0 to have a look.

NL66278:
:+1:  Limited testing. Code LGTM.

account_banking_payment_transfer/workflow/account_payment.xml
```diff
@@ -2,21 +2,21 @@
 <openerp>
     <data>
         <!-- New activity for workflow payment order: sent -->
-        <record id="account_banking.act_sent" model="workflow.activity">
+        <record id="act_sent" model="workflow.activity">
```
StefanRijnhart, line 5: Don't we need a migration script for these workflow activities?

account_banking_payment_transfer/model/account_payment.py
```diff
@@ -0,0 +1,259 @@
...
+        return False
+
+    @api.multi
+    def get_transfer_move_line_ids(self, *args):
+        '''Used in the workflow for trigger_expr_id'''
+        # TODO I don't understand why self._get_transfer_move_lines()
+        # returns a single recordset and not a list of recordset
+        # I wanted to write
+        # return self._get_transfer_move_lines().ids
+        # but it doesn't work, so I wrote this below:
```
StefanRijnhart: Is this harmful? Is there any case that multiple lines are expected?

account_banking_payment_transfer/model/account_payment.py
```diff
@@ -0,0 +1,259 @@
...
+    @api.returns('account.move.line')
+    def _get_transfer_move_lines(self):
+        """
+        Get the transfer move lines (on the transfer account).
+        """
+        for pay_line in self.line_ids:
+            move_line = pay_line.transfer_move_line_id
+            if move_line:
+                return move_line
+        return False
```
StefanRijnhart: You only return the first move line? Isn't the test in test_done not misleading then?

sbidoul: @alexis-via can you look after this one? It comes from the new api conversion you did, and now that I look at it again, test_done() looks suspicious indeed especially since _get_transfer_move_lines can return False which is not iterable. 

The test suite works, though and it must go through this, I'll check.

alexis-via:
Just had a remark this morning of the accountant of the Barroux Abbey that uses SEPA direct debit + account_banking_payment_transfer from this branch: if you do a SEPA direct debit file that contains both FIRST and RECURRING mandates (sequence type) at the same date, and you have set "transfer move option" = "One move per payment date", you will have 1 account move line in the transfer account. But the bank will generate 2 statement lines: 1 for the debits with FIRST mandates, and a second one for the debits with RECURRING mandates. So it would be better to have 2 account move lines on the transfer account in this case.

I guess all bank have this behavior, because FIRST and RECURRING debits have different delays : I'll check with another customer to see if their bank has the same behavior.

account_banking_payment_transfer/model/account_payment.py
sbidoul: @adrienpeiffer could find a scenario where it broke. So he partially reverted the new api change in https://github.com/acsone/bank-payment/commit/c4aa30a6a8e73acb4bbfe20fd5fe3e0241a05d16

sbidoul:
Given the positive feedback here and in #93 I merge into #65

sbidoul:
@pedrobaeza @alexis-via @StefanRijnhart WARNING. I merged this one by mistake instead of https://github.com/acsone/bank-payment/pull/5#issuecomment-70745569.

However, given the long review period, positive reviews, and tests by us and Akretion, I think no harm is done.

The only comment not cared for yet is @StefanRijnhart's remark about migration. Well do a new PR for this soon.

alexis-via:
yes, it a good thing to merge it, we'll do the enhancements in other PRs.

StefanRijnhart:
Alright. Created #105 to track this issue.

PR #67, opened 2014-11-20
[FIX] Payment lines are editable from the mandate form view

StefanRijnhart:
[FIX] Changing the payment order type from the payment or debit order form leads to unexpected behaviour
[FIX] When running tests on an existing database, make sure the admin user has the English language during the test


StefanRijnhart:
New attempt of https://github.com/OCA/bank-payment/pull/19

pedrobaeza:
Looks good. These issues are not present in v8 thanks to the migration to new API I think, but if you can try to be sure.

:+1:

hbrunn:
:+1: (ignoring unrelated pylint errors)

PR #68, opened 2014-11-21
[8.0] account_voucher_killer

sbidoul:
This module is migrated from OCA/bank-statement-reconcile 7.0.

As it's purpose is to remove the Pay buttons from the invoice forms, it is related to payment. 
This bank-payment repository is therefore a better place for it.

Since bank statements do not use vouchers in 8.0, this module is even more useful than in 7.0.

The module migrates trivally from 7.0.


guewen:
:+1:

invoicing_voucher_killer/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+Prevent the usage of voucher from invoices
+##########################################
+This add-on will disable "Register Payment" button
+""",
+ 'author': 'Camptocamp',
+ 'website': 'http://www.camptocamp.com',
+ 'depends': ['account_voucher'],
+ 'data': ['invoice_data.xml',
+          'invoice_view.xml'],
+ 'demo_xml': [],
```
pedrobaeza: demo_xml is deprecated and not used here, so you can remove it.

invoicing_voucher_killer/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+##############################################################################
+
+{'name': 'Invoicing voucher killer',
+ 'version': '1.0.0',
+ 'category': 'other',
+ 'description': """
+Prevent the usage of voucher from invoices
+##########################################
+This add-on will disable "Register Payment" button
```
pedrobaeza: Please put it in a separate file README.rst, and explain the description indicating that only users that belongs to the group "invoicing_voucher_killer.invoice_voucher_user" can see this button.

invoicing_voucher_killer/invoice_view.xml
```diff
@@ -0,0 +1,46 @@
+<openerp>
+  <data>
+    <record id="invoice_voucher_group" model="ir.ui.view">
+      <field name="name">Hide voucher in invoice</field>
+      <field name="model">account.invoice</field>
+      <field name="inherit_id" ref="account_voucher.view_invoice_customer" />
+      <field name="arch" type="xml">
+        <xpath expr="//button[@name='invoice_pay_customer'][last()]"
+               position="replace">
```
pedrobaeza: Instead of doing this hazardous replace, I would change attribute `groups` to change only that attribute.

pedrobaeza:
One last thing: I would take the occassion to change the name to _account_invoice_voucher_killer_ or something similar, following the de facto convention of prefixing area to the module name.

sbidoul:
@pedrobaeza I implemented the improvements you suggested.

Regarding the renaming, I'd say preserving compatibility and reducing migration hassles is more important that this convention.

pedrobaeza:
Well, this is a version change, and we can implement this module rename easily on OpenUpgrade, so I don't think it's a problem. We have bad inherited names, but we don't have to perpetuate them all the life-cycle of the module.

sbidoul:
@pedrobaeza @guewen @fclementic2c while talking about renaming this module, I was wondering if it would be a good idea to rename it account_voucher_killer, and put all voucher related menus in the same group to hide them too (ie hide Sales Receipts, Customer Payments, Purchase Receipts, Supplier Payments, Journal Vouchers). 

What do you think? When the pay buttons are hidden on the invoices by this module, are there other ways to create vouchers (I don't know any)?

pedrobaeza:
Yeah, it can be a possibility, but that way. Isn't better not to install _account_voucher_ at all? It's not mandatory to have it on your system. When you install it, you have two menus, one for customer and another one for suppliers, to make directly a voucher. They are called "Customer Payments" and "Supplier payments" respectively.

sbidoul:
@pedrobaeza I just tried to uninstall account_voucher on runbot, and there are so many dependencies that it would basically uninstall Odoo :wink:

pedrobaeza:
Yeah, I see. OK for me then.

fclementic2c:
Good idea Stephane to hide all these menus at the same time.
(and yes account_voucher is a dependancy of Sale...difficult to avoid.)

sbidoul:
@pedrobaeza @guewen @fclementic2c my colleague @anthony-muschang renamed the module and implemented the menu hiding. The module should now be complete and fully 8.0 ready.

guewen:
Nice, thanks!

pedrobaeza:
Thanks! Maybe you should add yourself to the credits.

:+1:

lmignon:
:+1:

account_voucher_killer/README.rst
```diff
@@ -0,0 +1,22 @@
+Prevent the usage of voucher from invoices
+===
```
pedrobaeza: This must be of the same length of the title string, or Odoo RST parser will complain

account_voucher_killer/README.rst
```diff
@@ -0,0 +1,22 @@
...
+
+It also disables the voucher-related menus entries such as 
+"Sales Receipts", "Customer Payments", "Purchase Receipts", 
+"Supplier Payments" and "Journal Vouchers", "Sales Receipts Analysis".
+
+More precisely, this module adds a group "Use voucher in Invoices"
+and only users in that group see these buttons and menus.
+
+Credits
+---
```
pedrobaeza: Idem

account_voucher_killer/README.rst
```diff
@@ -0,0 +1,22 @@
...
+
+Author:
```
pedrobaeza: You can add `-------` below to make it a subtitle. I think we shouldn't distinguish between author and contributors, because most of times the contributions can be bigger than the original code.

sbidoul:
@pedrobaeza I fixed README.rst according to your comments.

PR #69, opened 2014-11-24
[ADD] account_payment_multicurrency module

vrenaville:
This module add the possibility to see the amount in currency and to be able to select partial paid invoice

### DIFF
--- a/account_payment_multicurrency_extension/payment.py
+++ b/account_payment_multicurrency_extension/payment.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author: Vincent Renaville
+#    Copyright 2014 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.osv import orm
+
+
+class payment_order_create(orm.TransientModel):
+
+    _inherit = 'payment.order.create'
+
+    def create_payment(self, cr, uid, ids, context=None):
+        """
+        We recreate function to be able set
+        'amount_currency': line.amount_residual_currency
+        instead of
+        'amount_currency': line.amount_to_pay
+        To be compliant with multi currency
+        Allready corrected in V8 but will not be corrected in V7
+        """
+        order_obj = self.pool.get('payment.order')
+        line_obj = self.pool.get('account.move.line')
+        payment_obj = self.pool.get('payment.line')
+        if context is None:
+            context = {}
+        data = self.browse(cr, uid, ids, context=context)[0]
+        line_ids = [entry.id for entry in data.entries]
+        if not line_ids:
+            return {'type': 'ir.actions.act_window_close'}
+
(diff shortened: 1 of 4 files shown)


vrenaville:
@pedrobaeza @sbidoul do you have any comments on this PR

Thanks for your help,

Vincent

sbidoul:
@vrenaville I think you should consider making your module depend on account_banking_payment_export which provides a hook _prepare_payment_line. That should help you avoid copying the whole create_payment method.

Actually, I wonder if this issue is not fixed in account_banking_payment_export, can you check?

sbidoul:
If not, perhaps it's even easier to put your fix in account_banking_payment_export directly.

vrenaville:
@sbidoul thanks for your comment, I don't understand on the previous PR, I will test it and if it fit the need I will see to improve the view if required.

fclementic2c:
Hi Stephane,

I agree with your comment but it is difficult to do so in v7 since we already have several modules (swiss SEPA payement for instance) which are not base on this account_banking_payment_export and in production with several customers. 

But we will keep in mind your remark for v8

sbidoul:
@fclementic2c I understand.

Since this new module overrides create_payment without calling super() I'm concerned it will break other installs using account_banking_payment_export (which also overrides create_payment without calling super()...). 

The root of the problem is in the offical account_payment, of course.

However, introducing a conflicting module in OCA/bank-payment may create confusion.

vrenaville:
Hello,

@sbidoul  thanks for the review, I have add a super instead a complete rewrite of the function

Vincent

vrenaville:
@sbidoul I have discuss with @guewen I will improve my module, on monday.
Vincent

vrenaville:
@sbidoul I have just improve my code to have correct value assignment directly to payment line instead of search it.

Vincent

account_payment_multicurrency_extension/payment.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def create_payment(self, cr, uid, ids, context=None):
+        """
+        We add is_multi_currency tag to be able set
+        'amount_currency': line.amount_residual_currency
+        instead of
+        'amount_currency': line.amount_to_pay
+        To be compliant with multi currency
+        Allready corrected in V8 but will not be corrected in V7
+        """
```
guewen, line 36: context could be `None`, add a safeguard:

```
if context is None:
    context = {}
```

account_payment_multicurrency_extension/payment.py
```diff
@@ -0,0 +1,59 @@
...
+            ids,
+            context=context)
+
+
+class payment_line(orm.Model):
+    _inherit = 'payment.line'
+
+    def create(self, cr, uid, vals, context=None):
+        """In case of multi currency
+        we use amount_residual_currency instead of amount_to_pay"""
```
guewen, line 53: context could be None, add a safeguard:

if context is None:
    context = {}

vrenaville:
@guewen Thanks for the review.

I have correct the module according to your comment

Vincent

account_payment_multicurrency_extension/payment.py
```diff
@@ -0,0 +1,63 @@
...
+    _inherit = 'payment.order.create'
+
+    def create_payment(self, cr, uid, ids, context=None):
+        """
+        We add is_multi_currency tag to be able set
+        'amount_currency': line.amount_residual_currency
+        instead of
+        'amount_currency': line.amount_to_pay
+        To be compliant with multi currency
+        Allready corrected in V8 but will not be corrected in V7
```
gurneyalex: s/Allready/Already/

account_payment_multicurrency_extension/payment.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        if context is None:
+            context = {}
+        context.update({'is_multi_currency': True})
```
gurneyalex: do we really want the modified context to be seen by the caller?

the usual idiom is to make a copy and modify the copy.

fclementic2c:
functionnal test :+1:

vrenaville:
@gurneyalex I have made the fix according to your comment

guewen:
Red due to pylint warnings in other modules :+1:

PR #70, opened 2014-11-25
[FIX] remove leftover code from refactoring

hbrunn:
that's a followup for the sepa backport, the _auto_init patch fixes a module update that goes wrong because the model also existed in account_banking and the xml id would be recreated during an --update=all


pedrobaeza:
:+1:

guewen:
:+1:

sbidoul:
:+1:

PR #72, opened 2014-11-28
[FIX] Voucher still assigns bank statement period instead of statement line period

StefanRijnhart:
In 6.1, we could not call super on create_move_from_st_line and had to incorporate a large body of core code in account_banking. In 7.0 it looked like we did because of refactorings driven by the good people of C2C but as it turns out, account_voucher's override of this method still insists in writing the statement period to the voucher that account_banking creates for each reconciled bank statement line, instead of the period based on the statement line's date itself.

I see no other way than monkeypatching this out of account_voucher but I am taking care not to mess up other databases on the same instance. Setting this file to noqa to preserve original style for easy detection of any upstream changes.

### DIFF
--- a/account_banking/__init__.py
+++ b/account_banking/__init__.py
@@ -2,6 +2,7 @@
 ##############################################################################
 #
 #    Copyright (C) 2009 EduSense BV (<http://www.edusense.nl>).
+#              (C) 2011 - 2014 Banking addons community
 #    All Rights Reserved
 #
 #    WARNING: This program as such is intended to be used by professional
@@ -34,3 +35,4 @@ from . import wizard
 from . import res_partner
 from . import res_bank
 from . import res_partner_bank
+from . import bank_statement_monkeypatch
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why don't you overwrite _create_move_from_st_line_, call super, and then rewrite again the period?

StefanRijnhart:
because the voucher gets confirmed in the same method I am monkeypatching, and I cannot change the period on a confirmed voucher (or it would have no effect).

pedrobaeza:
OK, I see, very bad practises from Odoo again...

Travis errors unrelated.

:+1:

lfreeke:
:+1:  works. But needs a rebase

StefanRijnhart:
Setting to 'Needs fixing' as per @lfreeke's comments. @sbidoul, why enhancement? Looks like a bug to me if your moves are posted in the wrong period.

StefanRijnhart:
Sorry for the delay. Now rebased.

NL66278:
:+1:  Tested and running succesfully in production instance.

StefanRijnhart:
I believe this has been ready to be merged for a year now. Maybe it was ignored because the 'bug' label made reviewers think that the PR causes a bug?

pedrobaeza:
Merging.

StefanRijnhart:
Thanks @pedrobaeza ;-)

PR #75, opened 2014-12-03
[FIX] Add CASCADE option to DROP TABLE sdd_mandate

NL66278:
Prevent exception on migrating sdd mandates.

### DIFF
--- a/account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py
+++ b/account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py
@@ -50,5 +50,5 @@ def migrate(cr, installed_version):
     cr.execute(query4)
     query5 = "ALTER TABLE payment_line DROP COLUMN IF EXISTS sdd_mandate_id"
     cr.execute(query5)
-    query6 = "DROP TABLE IF EXISTS sdd_mandate"
+    query6 = "DROP TABLE IF EXISTS sdd_mandate CASCADE"
     cr.execute(query6)


pedrobaeza:
:+1:

StefanRijnhart:
From what additional table are the entries removed? It would be disastrous if it concerned bank accounts or payment lines.

NL66278:
It where the constraints that prevented the table drop. Strange, because I would think dropping the constraints would be implicit in dropping the table.

StefanRijnhart:
No, that would defeat the purpose of the constraints, wouldn't it? How about removing the constraints manually from the schema?

NL66278:
Well, what is the use of constraints on a table about to be removed? And what the advantage of encountering an error, and then having to do a lot of manual work?

StefanRijnhart:
Maybe I don't understand the topic well enough, but I strongly fear that if a table drop does not work without cascade, it is taking down columns of unknown tables with it once you add the cascade

About the use of constraints on tables to be dropped: the constraint is to guarantee the referential integrity, which would be violated by dropping the table. So removing the constraint manually, thereby taking responsibility for the referential integrity seems the right way to do it IMHO.

NL66278:
The problem is we have with several of our customers an extra module that also refers to sdd_mandate. So while the migration moves info from sdd_mandate_id to mandate_id (without sdd_), then drops the sdd_mandate_id field from the tables, it does not take care of the extra reference. I double checked what drop cascade does. It does NOT remove any referring fields from other tables, however the constraint is dropped, basically turning sdd_mandate_id into a simple integer field. (Which could still be used to set mandate_id in the extra table, when if would be needed to migrate that one).

hbrunn:
the advantage of getting an error is to make it harder to mess up stuff. On my test database, the table looks like this:

```
\d sdd_mandate
                                                Table "public.sdd_mandate"
             Column              |            Type             |                        Modifiers                         
---------------------------------+-----------------------------+----------------------------------------------------------
 id                              | integer                     | not null default nextval('sdd_mandate_id_seq'::regclass)
 create_uid                      | integer                     | 
 create_date                     | timestamp without time zone | 
 write_date                      | timestamp without time zone | 
 write_uid                       | integer                     | 
 last_debit_date                 | date                        | 
 partner_bank_id                 | integer                     | 
 scan                            | bytea                       | 
 recurrent_sequence_type         | character varying           | 
 sepa_migrated                   | boolean                     | 
 company_id                      | integer                     | not null
 original_mandate_identification | character varying(35)       | 
 state                           | character varying           | 
 unique_mandate_reference        | character varying(35)       | 
 signature_date                  | date                        | 
 type                            | character varying           | not null
Indexes:
    "sdd_mandate_pkey" PRIMARY KEY, btree (id)
    "sdd_mandate_mandate_ref_company_uniq" UNIQUE CONSTRAINT, btree (unique_mandate_reference, company_id)
Foreign-key constraints:
    "sdd_mandate_company_id_fkey" FOREIGN KEY (company_id) REFERENCES res_company(id) ON DELETE SET NULL
    "sdd_mandate_create_uid_fkey" FOREIGN KEY (create_uid) REFERENCES res_users(id) ON DELETE SET NULL
    "sdd_mandate_partner_bank_id_fkey" FOREIGN KEY (partner_bank_id) REFERENCES res_partner_bank(id) ON DELETE SET NULL
    "sdd_mandate_write_uid_fkey" FOREIGN KEY (write_uid) REFERENCES res_users(id) ON DELETE SET NULL
Referenced by:
    TABLE "account_invoice" CONSTRAINT "account_invoice_sdd_mandate_id_fkey" FOREIGN KEY (sdd_mandate_id) REFERENCES sdd_mandate(id) ON DELETE SET NULL
    TABLE "payment_line" CONSTRAINT "payment_line_sdd_mandate_id_fkey" FOREIGN KEY (sdd_mandate_id) REFERENCES sdd_mandate(id) ON DELETE SET NULL
```

The last two lines tell us that account_invoice and payment_line depend on it, and your cascade will also drop those tables. And the ones that depend on those two. And their dependencies. Bad idea.

`query3` and `query4` should have removed the constraints above, so I think you have to analyze where the constraints you're struggling with come from, and resolve them the same way.

NL66278:
@hbrunn this is not the case. Other tables will not be dropped. I did test it.

hbrunn:
@NL66278 yes, you're right. shame on me, it's the constraint which is dropped. Thanks for bringing this up.

hbrunn:
(pathetic attempt to retain honor: cascading truncates do what I described above)

hbrunn:
:+1:

StefanRijnhart:
@NL66278 thanks for confirrming, and thank you for teaching me a thing or two about DROP TABLE ... CASCADE. :+1:

PR #76, opened 2014-12-03
[FIX] Prevent exception on showing sepa files after deletion on payment/...

NL66278:
...direct debit orders.

Withouth this, when users delete payment orders, an exception will occur when using the menu options to show the transfer files.

### DIFF
--- a/account_banking_sepa_credit_transfer/account_banking_sepa.py
+++ b/account_banking_sepa_credit_transfer/account_banking_sepa.py
@@ -34,11 +34,14 @@ class banking_export_sepa(orm.Model):
     def _generate_filename(self, cr, uid, ids, name, arg, context=None):
         res = {}
         for sepa_file in self.browse(cr, uid, ids, context=context):
-            ref = sepa_file.payment_order_ids[0].reference
-            if ref:
-                label = unidecode(ref.replace('/', '-'))
+            if not sepa_file.payment_order_ids:
+                label = 'no payment order'
             else:
-                label = 'error'
+                ref = sepa_file.payment_order_ids[0].reference
+                if ref:
+                    label = unidecode(ref.replace('/', '-'))
+                else:
+                    label = 'error'
             res[sepa_file.id] = 'sct_%s.xml' % label
         return res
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Maybe a better approach is to restrict or delete cascade these files?

NL66278:
@pedrobaeza  The current pull request will prevent exceptions from occuring if there is any inconsistency between payment orders and available sepa transfer files. I think that is an improvement anyway. 
I am reluctant to restrict throwing away files / confirmed payment orders (although it might be argued this should only be possible for aministrators). Making it impossible to delete, will make it more difficult to rectify mistakes. I would opt for cascading delete - also only for administrators. What do you think?

pedrobaeza:
OK, I agree that cascading is the best option. I would put account managers as the one with deletion permission.

sbidoul:
I'm fine with cascading delete too. 

Maybe @alexis-via's opinion would be valuable here.

hbrunn:
:+1: and I also agree that cascading deletes make sense here. But that shouldn't block merging this

alexis-via:
I agree with this change :+1:

In fact, I always thought that the models banking.export.sdd and banking.export.sepa to store the XML files where useless, and that we should just put the XML file as an attachment to the payment.order. It would avoid quite a lot of code, and I even think that it would be a bit easier for the user. But I followed the implementation of the existing module of the banking addons, and that's why I created the models banking.export.sdd and banking.export.sepa.

Maybe having these models has some advantages compared to using attachments, I would be curious to know what are the advantages...

pedrobaeza:
I'm glad to see that I'm not the only one that see both models as useless. We can make that move with the proper migration script.

alexis-via:
@pedrobaeza I'm glad to know that I'm not the only one to think this :) I'll start a discussion in the mailing-list before we start coding on this. Of course, updating the code for that it quite easy... but the migration script will not be easy to write :-(

sbidoul:
I was also wondering why there was a separate model for the generated SEPA XML files. I never dared to ask.

:+1: for storing them as attachments.

NL66278:
Maybe somebody can merge this PR?

I agree that storing the xml files as attachments is a good idea, but that might be part of a new and separate PR?

pedrobaeza:
I merge then, but the PR already exists here :https://github.com/OCA/bank-payment/pull/123

PR #78, opened 2014-12-10
[ADD] Add account_pay_draft_move addons

adrienpeiffer:


adrienpeiffer:
Thanks for the review @sbidoul

pedrobaeza:
The module name should be `account_payment_include_draft_move` for being self-explanatory.

sbidoul:
I agree with the renaming proposed by Pedro.

adrienpeiffer:
indeed it's more explicit ! Thanks @pedrobaeza @sbidoul

account_payment_include_draft_move/README.rst
```diff
@@ -0,0 +1,8 @@
+Account Payment Draft Move
+==============================================
```
pedrobaeza: Put same length for the title string and the == string. This can be `Include draft moves in account payments`

account_payment_include_draft_move/README.rst
```diff
@@ -0,0 +1,8 @@
...
+
+Add payment order line from unposted move line.
+
+With account_default_draft_move, this module allow to add move line
```
pedrobaeza: s/allow/allows

account_payment_include_draft_move/README.rst
```diff
@@ -0,0 +1,8 @@
+Account Payment Draft Move
+==============================================
+
+Add payment order line from unposted move line.
```
pedrobaeza: s/line/lines

account_payment_include_draft_move/README.rst
pedrobaeza: s/line/lines

account_payment_include_draft_move/README.rst
```diff
@@ -0,0 +1,8 @@
...
+
+With account_default_draft_move, this module allow to add move line
+before periodic processing to post all move
```
pedrobaeza: Replace with `...to payment orders before making the periodic process of posting all moves`

account_payment_include_draft_move/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Account Payment Draft Move",
+    "version": "1.0",
+    "author": "ACSONE SA/NV",
+    "maintainer": "ACSONE SA/NV",
```
pedrobaeza: Maintainer is not a key that is used in the manifest

adrienpeiffer:
Thanks @pedrobaeza !

pedrobaeza:
Sorry, closed by mistake.

:+1:

StefanRijnhart:
:+1:, adding new CLA Pending tag

sbidoul:
@StefanRijnhart, @adrienpeiffer 's contributions are covered by ACSONE's ECLA.

StefanRijnhart:
@sbidoul I always understood this differently but I must admit that it is a bit ambiguous. The contributor is also asked to sign the individual CLA 'to cover any of their contributions which are not owned by the corporation signing the Entity CLA'. Of course, apart from your intervention on threads like these, it is impossible to tell (for me as well as for the cla-bot) if the contribution was done during work hours. I'm taking it to the OCA board for clarification but I'll take your word for it now.

account_payment_include_draft_move/wizard/payment_order_create.py
```diff
@@ -0,0 +1,36 @@
...
+POSTED_MOVE_DOMAIN = ('move_id.state', '=', 'posted')
+
+
+class PaymentOrderCreate(models.TransientModel):
+    _inherit = 'payment.order.create'
+
+    @api.model
+    def extend_payment_order_domain(self, payment_order, domain):
+        if POSTED_MOVE_DOMAIN in domain:
+            domain.remove(POSTED_MOVE_DOMAIN)
```
StefanRijnhart: On second thoughts, I have a code comment. By removing domain tuples, you run the risk of creating an imbalanced domain, depending on what operators the element is governed by. Would it not be better to replace the element by the true leaf (1, '=', 1)?

sbidoul: @StefanRijnhart replacing de element by (1,'=',1) would be slightly more robust. That said, if the domain as been modified too much by other addons, the mechanisme is fragile anyway. But I see no other way. @adrienpeiffer can you make this last change?

sbidoul:
@StefanRijnhart what would be the purpose of the ECLA if all contributors need to have an individual CLA in addition? One could say that when contributors contribute as employee they do it with their company e-mail address, and when they contribute as private individuals, they contribute with their private e-mail address. 'hope we can keep paperwork low.

adrienpeiffer:
Rebased

sbidoul:
:+1:

PR #79, opened 2014-12-10
[IMP] sepa credit tfr: get BIC from bank account if bank is not specified on bank account

sbidoul:
Since the bank is optional on bank accounts and the BIC can be set
on the bank account if there is no bank, we allow the system to
get the BIC from the bank account in case it is not on the bank
or the bank account has no associated bank.

We give priority to the BIC indicated on the bank.

### DIFF
--- a/account_banking_pain_base/models/res_partner_bank.py
+++ b/account_banking_pain_base/models/res_partner_bank.py
@@ -1,9 +1,7 @@
 # -*- encoding: utf-8 -*-
 ##############################################################################
 #
-#    PAIN Base module for OpenERP
-#    Copyright (C) 2013 Akretion (http://www.akretion.com)
-#    @author: Alexis de Lattre <alexis.delattre@akretion.com>
+#    Copyright (c) ACSONE SA/NV (<http://acsone.eu>).
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


sbidoul:
If you agree with this, I can do the same for the SDD part.

alexis-via:
:+1: code review, no test.
If this is approved, it will be important to do the same on SDD of course.

pedrobaeza:
Shouldn't it be the contrary on the priority? Use case: a user finds an error on master data, and fix it at partner bank level because he/she doesn't have permission to modify bank data.

sbidoul:
@pedrobaeza I did it this way to avoid changing the behaviour with existing data. But I'm fine with your proposal.

StefanRijnhart:
Because the data model in upstream Odoo is obviously broken, containing duplicate data it seems pretty arbitrary which data is checked first. So :+1: as it is.

pedrobaeza:
OK, :+1: 

Do you think it deserves a mention in the module description?

alexis-via:
I agree with stefan that the datamodel in upstream odoo is broken.

I think it's important to explain in the help message in the fields which one has the priority (it will also be usefull for us, to remember which one is used first without looking at the code):

we could put in the module account_banking_pain_base :

```
class ResPartnerBank(models.Model):
    _inherit = 'res.partner.bank'

   bank = fields.Many2one(help="If this field is set and the related bank has a 'Bank Identifier Code', then this BIC will be used to generate the credit transfers and direct debits files.")
   bank_bic = fields.Char(help="In the generation of credit transfer and direct debit files, this BIC will be used only when the 'Bank' field is empty, or has a value but the field 'Bank Identifier Code' is not set on the related bank.")
```

Of course, feel free to put better help messages.

pedrobaeza:
Good option to put it on the help of the field! I agree with that.

sbidoul:
I 
- implemented the same logic on SDD, 
- added the help text as suggested by @alexis-via 
- rebased

Not tested yet.

pedrobaeza:
Thanks! 

:+1:

sbidoul:
I fixed a couple of bugs I introduced in the previous commit and rebased.

alexis-via:
Let's merge this one ! Anyone with the super-power to merge this ?

sbidoul:
@alexis-via done!

PR #80, opened 2014-12-14
[IMP] sepa: make it easier to extend structured communication types

sbidoul:
Implement the usual trick to make selection field extensible

### DIFF
--- a/account_banking_pain_base/models/payment_line.py
+++ b/account_banking_pain_base/models/payment_line.py
@@ -42,7 +42,9 @@ class PaymentLine(orm.Model):
             "company. Depicts 'What do you want to say to the recipient "
             "about this order ?'"),
         'struct_communication_type': fields.selection(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
With new API, we have now several methods for extending selections:
- Enclose in quotes the method name, and it will be evaluated at execution time, being the same as your lambda method.
- Use _extend_selection_ attribute to add values.

I recommend you to make the change enclosing method in quotes, or don't do any change, and extend selection in derived modules the other way.

sbidoul:
Thanks @pedrobaeza. I implemented your first suggestion.

pedrobaeza:
:+1:

guewen:
:+1:

sbidoul:
Is this one good for merging?

pedrobaeza:
Yeah, I think so.

PR #81, opened 2014-12-14
[IMP] set default value for payment.mode.type

sbidoul:
Since the type field is required and payment modes may exist when
account_banking_payment_export is installed, a default value
is necessary to let Odoo set the required database constraint.

We use the Manual Bank Transfer as default.

### DIFF
--- a/account_banking_payment_export/models/payment_mode.py
+++ b/account_banking_payment_export/models/payment_mode.py
@@ -43,9 +43,14 @@ class PaymentMode(models.Model):
             res = [t.code for t in payment_mode.type.suitable_bank_types]
         return res
 
+    def _default_type(self):
+        return self.env.ref('account_banking_payment_export.'
+                            'manual_bank_tranfer')
+
     type = fields.Many2one(
         'payment.mode.type', string='Export type', required=True,
-        help='Select the Export Payment Type for the Payment Mode.')
+        help='Select the Export Payment Type for the Payment Mode.',
+        default=_default_type)
     payment_order_type = fields.Selection(
         related='type.payment_order_type', readonly=True, string="Order Type",
         selection=[('payment', 'Payment'), ('debit', 'Debit')],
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Unfortunately this makes Travis fail. Maybe we can tweak the module to force any value on init, or in the post-hook?

account_banking_payment_export/models/payment_mode.py
```diff
@@ -43,9 +43,14 @@ def suitable_bank_types(self, cr, uid, payment_mode_id=None, context=None):
             res = [t.code for t in payment_mode.type.suitable_bank_types]
         return res
 
+    def _default_type(self):
+        return self.env.ref('account_banking_payment_export.'
+                            'manual_bank_tranfer')
+
     type = fields.Many2one(
```
lmignon, line 53: The method should return a recordset.. (even if it is empty)
It should work with the following method

``` python
@api.model
def _default_type(self):
    return self.env.ref(
        'account_banking_payment_export.'
        'manual_bank_tranfer', raise_if_not_found=False)\
        or self.env['payment.mode.type']
```

Odoo will try to apply the default value a fist time during the installation step and give a warning. It will retry again during the post-install step and this time it should work.

sbidoul, line 53: @lmignon very interesting, thanks!

Can you add a commit to the branch to try if this works?

Would `@api.returns('payment.mode.type')` help avoiding the `or self.env['payment.mode.type']` ?

lmignon, line 53: @sbidoul I'll do the changes 
Unfortunately, the @api.returns('payment.mode.type') is without effect in this case (as I remember) (a odoo bug?)

sbidoul:
:+1:

pedrobaeza:
:+1: Let's merge this then

PR #87, opened 2014-12-31
[FIX] account_banking_pain_base: Set a general rule for the initiating party identifier

pedrobaeza:
Currently, there is no general rule for each country of the SEPA area for getting the initiating party identification (only two for Belgium and Spain), so the rest of the countries cannot generate for example correct SEPA direct debit files. Also, with current rules, you need to set company vat identification for making any match, which is unneeded.

I propose to set company SEPA creditor identifier as the default value, and then each country can set another value (as Belgium does) if it's needed.

### DIFF
--- a/account_banking_pain_base/models/res_company.py
+++ b/account_banking_pain_base/models/res_company.py
@@ -44,13 +44,11 @@ class ResCompany(orm.Model):
         assert isinstance(company_id, int), 'Only one company ID'
         company = self.browse(cr, uid, company_id, context=context)
         company_vat = company.vat
-        party_identifier = False
-        if company_vat:
+        party_identifier = company.sepa_creditor_identifier
+        if not party_identifier and company_vat:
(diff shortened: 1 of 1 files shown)


sbidoul:
This makes sense to me. :+1:

StefanRijnhart:
Would it even make sense to give this field always priority if its filled in? Currently, in the case of a company in Belgian, the field is ignored which is not obvious at all from the interface.

pedrobaeza:
Good point, Stefan,

I have changed the PR according your comments.

StefanRijnhart:
Thanks! :+1:

Olivier-LAURENT:
It seems this PR introduces a dependency issue. The field `sepa_creditor_identifier` is declared in `account_banking_sepa_direct_debit` but there is no dependencies route from `account_banking_pain_base` to reach it.

pedrobaeza:
This is being discussed here: #112.

PR #91, opened 2015-01-08
[IMP] _columns deprecated

rlizana:


rvalyi:
:+1: LGTM

pedrobaeza:
Thanks for the change, Roberto. This is one of the less modules that I didn't touch for the migration, and now there is a better option for doing this: you can declare the field as usual (fields.Selection) and add the attribute `extend_selection=` with the new value/s for the selection, and there's no need to make this little hack.

rvalyi:
@pedrobaeza you are right indeed I had forgotten that one.

StefanRijnhart:
Thanks Roberto. I missed your proposal before but it is now obsolete by this commit https://github.com/OCA/bank-payment/commit/9ad1dd94a249f2e45342fa733b64bb693b0fdda6

PR #95, opened 2015-01-14
account_payment_partner : remove unused file + cleanup code

alexis-via:


account_payment_partner/wizard/payment_order_create.py
```diff
@@ -28,10 +28,10 @@ class PaymentOrderCreate(models.TransientModel):
...
     def extend_payment_order_domain(self, payment_order, domain):
-        res = super(PaymentOrderCreate, self).extend_payment_order_domain(
+        super(PaymentOrderCreate, self).extend_payment_order_domain(
             payment_order, domain)
         domain += ['|', '|',
                    ('invoice', '=', False),
                    ('invoice.payment_mode_id', '=', False),
                    ('invoice.payment_mode_id', '=', payment_order.mode.id)]
-        return res
+        return True
```
rvalyi: in general it's good to propagate the return from super (unless you really want to change it). Here super will return True too but eventually it could be changed. Is there any reason to hardcode return True in this override? Not against it, but just asking why this choice. (approve the file removal otherwise).

sbidoul: @alexis-via's change in return True mimicks the behaviour of the super method.
Actually, it's the return in the super method that should be removed as that method has no meaningful return value.

sbidoul:
:+1:

account_payment_partner/wizard/payment_order_create.py
rvalyi: Hum in fact many methods used to return True to avoid returning nothing and not be incompatible with the XMLRPC protocol, so may be this is the kind of dummy meaningless return True. Let's say I just fail to see how it's better to hardcode True than returning super. But anyway not blocking for such a detail.

rvalyi:
:+1:  so LGTM

account_payment_partner/wizard/payment_order_create.py
sbidoul: Thanks for the reminder about XMLRPC. I had forgotten that one. I agree returning the result of super is a good habit to have.

pedrobaeza:
I also don't see not to propagate return code (nevertheless it's None, True or anything meaningful). If you want to change the return value, change it on the first return of the hierarchy chain.

alexis-via:
ok, we can merge it now :+1:

Issue #96, opened 2015-01-14
Cannot create a payment line manually


alexis-via:
Scenario to reproduce the bug:
1) create a new DB with demo data and install the account_payment module
2) create a new payment.order. Then create a new payment line MANUALLY (not via the wizard)
Partner: camptocamp
AMount : 42
Communication : Test
Click "Save & close" on the form view of the payment line, then "save" on the payment order => it works
3) install the module account_banking_mandate
4) execute step 2 again : create a new payment order and create MANUALLY a new payment line, click "Save & close" and then "Save" : it blocks with the following error:

```
Integrity Error

The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: order_id - order.id]
```

The openerp logs show:

```
2015-01-14 17:02:06,044 26664 ERROR o8_test2 openerp.sql_db: bad query: INSERT INTO "payment_line" ("id", "date", "currency", "state", "name", "company_currency", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('payment_line_id_seq'), NULL, 1, 'normal', '004', 1, 1, 1, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
IntegrityError: null value in column "order_id" violates not-null constraint
DETAIL:  Failing row contains (3, null, 2015-01-14 17:02:06.022553, 1, null, null, null, 1, 2015-01-14 17:02:06.022553, null, 1, null, 1, 004, null, null, null, normal, null, null).

2015-01-14 17:02:06,047 26664 ERROR o8_test2 openerp.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 525, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 562, in dispatch
    result = self._call_function(**self.params)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 302, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/service/model.py", line 149, in wrapper
[...]

PR #97, opened 2015-01-14
[FIX] Bug #96 bad use of new API

alexis-via:
Full scenario of the bug on the bug report #96


sbidoul:
The relationship between the bug description and the fix is not obvious. You must have had a hard time to diagnose this ;)

:+1:

alexis-via:
@sbidoul yes, it was quite hard to find the cause of this bug !!!

pedrobaeza:
Indeed it's not obvious.

Thanks. :+1:

PR #99, opened 2015-01-16
[Usability] mandates: search by reference, add group_by, add seq type in tree view

alexis-via:


account_banking_mandate/views/account_banking_mandate_view.xml
```diff
@@ -68,11 +68,15 @@
 		<field name="model">account.banking.mandate</field>
 		<field name="arch" type="xml">
 			<search string="Search Banking Mandates">
```
StefanRijnhart: Thanks for the improvements! Odd spacing here, as in the other XML. Are tabs involved, in the original or new code? Maybe this is a good time to reformat on indents of 4 spaces?

pedrobaeza:
Can you please add also by type of mandate (CORE, B2B)?

account_banking_mandate/views/account_banking_mandate_view.xml
alexis-via: yes, it's caused by the fact that there are tabs everywhere in the XML file !!! And the lines I added have spaces.

alexis-via:
@pedrobaeza yes, i'll update the PR with your remark when I have 5 min

alexis-via:
This PR is finished, we can merge it. Apart from the replacement of tabs by spaces, it's just a few more lines in tree+search view of mandates. No risk.

pedrobaeza:
All is correct as said by Alexis, so I merge. :+1:

PR #92, opened 2015-01-08
[FIX] uk_hsbc statement import was not parsing detail lines

challs:
Fixed a problem with a recent change to the regexp string causing it
to no longer match against statement lines due to backslash.

Also fixed testing code when running from the command line

PR #102, opened 2015-01-19
7.0 fix uk hsbc mt940 import

challs:
Fixed a problem with a recent change to the regexp string causing it
to no longer match against statement lines due to backslash.

Also fixed testing code when running from the command line

This pull request replaces https://github.com/OCA/bank-payment/pull/92, which was missing PEP8 compliance.

@sbidoul previously commented:

> Interesting bug. The fix looks good to me.
> Can you add a space after the comma for pep8 compliance?

PEP8 fix in 5de112de8dc0e547af283a73e25d48554255dd2a

### DIFF
--- a/account_banking_uk_hsbc/mt940_parser.py
+++ b/account_banking_uk_hsbc/mt940_parser.py
@@ -50,18 +50,18 @@ class HSBCParser(object):
                            r"(?P<startingbalance>[\d,]{1,15})")
 
         # Transaction
-        recparse["61"] = r"""\
-:(?P<recordid>61):\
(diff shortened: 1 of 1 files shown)


veloutin:
:+1:  Also consider verbose (re.VERBOSE) for this kind of regexp.

challs:
> Also consider verbose (re.VERBOSE) for this kind of regexp.

Thanks for your suggestion @veloutin! I considered using `re.VERBOSE` first, but the expression shares code with the other patterns (because of the %patterns substitution), so the change would have been more intrusive. I felt this was the simplest way to express this change.

yvaucher:
:+1:

PR #103, opened 2015-01-19
8.0 use sale_ok and purchase_ok filters in partner/sale/purchase form views

alexis-via:
Use widget=selection for payment_mode_id fields
Update demo data

### DIFF
--- a/account_banking_payment_export/demo/banking_demo.xml
+++ b/account_banking_payment_export/demo/banking_demo.xml
@@ -59,6 +59,7 @@
 
 <record id="account_payment.payment_mode_1" model="payment.mode">
     <field name="type" ref="account_banking_payment_export.manual_bank_tranfer"/>
+    <field name="sale_ok" eval="False"/>
 </record>
 
 <record id="payment_mode_2" model="payment.mode">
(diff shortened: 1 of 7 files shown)


pedrobaeza:
This way, we mimick payment term behaviour.

:+1:

StefanRijnhart:
Does this mean that I cannot add a payment type to a customer refund?

sbidoul:
@StefanRijnhart not sure to understand your question. 

These domains mean that only payment modes allowed for sale (sale_ok) are visible on customer invoices, and only payment modes allowed for purchase (purchase_ok) are visible on supplier invoices.

StefanRijnhart:
If I have a customer refund, I might want to add it to a credit transfer. Customer refunds share the customer invoice form view. With this change, I cannot select a credit transfer payment mode on the invoice, if I understand correctly. Maybe the patch can take the invoice type into account.

alexis-via:
@StefanRijnhart 
You point seems valid, but the problem is already present with the current code ; it's not specific to this PR.

pedrobaeza:
Yeah, sale_ok and purchase_ok is not the problem. The problem comes from the domain apply adding invoices to the payment order. Sometime ago, Stefan, I ask you in 7.0 if refunds are included in a PR you made. As me, you didn't know because the code was not yours. Migrating the set to v8, I found that the problem exists, but I made only a port because of the hurry.

We should treat this problem, but on another issue.

StefanRijnhart:
Right! Thanks for your comments. :+1: then.

sbidoul:
So I merge this one.

@pedrobaeza would you create an issue to track your comment https://github.com/OCA/bank-payment/pull/103#issuecomment-71017335 above?

PR #107, opened 2015-01-21
[FIX] Cannot extend selection if object is not fully initialized

### LINKED ISSUE #106: Test fails with Odoo, not OCB
https://travis-ci.org/OCA/bank-payment/builds/47806067

File "/home/travis/build/OCA/bank-payment/account_direct_debit/models/account_invoice.py", line 140, in  __ init __
invoice_obj._columns['state'].selection.append(
KeyError: 'state'

StefanRijnhart:
Fixes #106

### DIFF
--- a/account_direct_debit/models/account_invoice.py
+++ b/account_direct_debit/models/account_invoice.py
@@ -126,7 +126,7 @@ from openerp.tools.translate import _
 class AccountInvoice(orm.Model):
     _inherit = "account.invoice"
 
-    def __init__(self, pool, cr):
+    def _register_hook(self, cr):
         """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In 8.0, we can use now the new API attribute extend_selection to add a new one. Maybe it's time for doing it in this module, the only one I think it's not migrated to new API. Or is too much work?

StefanRijnhart:
It's quite a bit of work. Can we merge this first, to fix the tests again?

pedrobaeza:
OK,I fast track this and we can think about it later.

PR #108, opened 2015-01-22
[FIX] Don't break inheritance chain

StefanRijnhart:
Fix inheritance chain that was broken in my PR #107


pedrobaeza:
:+1: We should fix this ASAP.

sbidoul:
Otherwise :+1: 

Thanks!

yvaucher:
:+1:

Issue #101, opened 2015-01-16
Use of Waiting accounts in payment orders (Cash discount)


chrisdec:
Hello,

I have a question regarding the process of the payment order. Normally, it’s recommended to use waiting accounts to avoid paying two times the same invoice. However, I have some customers who work with cash discount.

Let’s take an exemple :

I have an invoice of 121 € with due date of 31/01/2015 but if I pay before 15/01/2015, I only pay 119€.

If they use waiting accounts, the amount on the waiting account will be the amount they paid in the payment order. The reconciliation will be partial as the paid amount is not the total amount of the invoice and the invoice will stay open. Moreover, as the invoice is still open, Odoo will propose to pay it a second time.  

If they don’t use the waiting accounts, they can reconcile easily on the bank statement with the cash discount account but here also, Odoo will propose to pay the invoice a second time before the first bank statement and the reconciliation.

What do you suggest?

Thanks in advance for the answer,

Christelle

sbidoul:
This relates to #93. Indeed, moves to the transfer account create partial reconciles, and https://github.com/acsone/bank-payment/pull/5 works best when invoices are fully reconciled as soon as possible.

A solution could be to do the write-off of the difference at the same time as the move to the transfer account.

What do others think?

adrienpeiffer:
I agree with your proposition @sbidoul.

sbidoul:
The problem is that we would end-up re-creating a full reconcile/writeoff wizard that already exists.

So maybe adding a button on the payment order to show the unreconciled transfer journal entries would be enough? From there, the user could easily finalise the write-off using the standard wizard.

sbidoul:
https://github.com/acsone/bank-payment/commit/f6c8cd782caea6c3a49bc349128b240ab1618590 in https://github.com/acsone/bank-payment/pull/5 provides a solution for this, letting the user easily register cash discount differences just after doing the payment order:
- a button on the payment order to access the partially reconciled payments moves
- from that screen, the user can use the standard reconcile with writeoff feature to do the writeoff to the cash discount account, therefore completely reconciling the invoices which in turn will not be proposed anymore in payment orders

pedrobaeza:
This is already solved.

PR #110, opened 2015-01-22
8.0 improve selection of invoices to pay, solves #93

### LINKED ISSUE #93: Problem adding partially reconciled moves to a payment order
Hello,

When adding invoices to a payment order, the algorithm in account_banking_payment_export (amount_to_pay) takes the credit minus the existing payment lines in valid payment orders.
In a scenario where the original invoice has been partially reconciled with a credit note (and therefore not paid through a payment order), the proposed amount is still the original invoice. 

The scenario goes as follow:
- create a supplier invoice for amount 600€
- create a supplier refund for amount 200€ and reconcile partially it with the supplier invoice
- the residual amount on the supplier invoice is 400€
- create a payment order and add the invoice: the system proposes 600€ where the user expects 400€

In official 8.0, the amount to pay is the residual amount (which is now correct in the presence of partial reconciliations). This does not take into existing payment orders anymore, though. IMO, this is fine especially when people work with transfer accounts, because as soon as the payment order is confirmed, the invoice is reconciled and the residual updated.

So I would suggest 
- dropping this computation of amount_to_pay
- when searching invoices to pay, filtering on unreconciled entries where credit > 0 instead of amout_to_pay
- when adding to the payment order, using the residual amount

@alexis-via @StefanRijnhart @pedrobaeza what is your opinion?

sbidoul:
This PR has already been extensively discussed in https://github.com/acsone/bank-payment/pull/5.

It provides:
- an improvement of the criteria to select invoices to pay
- a button on payment orders with transfer moves to show the partially reconciled transfer moves in one click, therefore providing a path to reconcile and writeoff such partial payments; this is particularly useful in the presence of cash discount and fixes #93

### DIFF
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -3,6 +3,7 @@
 #
 #    Copyright (C) 2009 EduSense BV (<http://www.edusense.nl>).
 #              (C) 2011 - 2013 Therp BV (<http://therp.nl>).
+#              (C) 2014 - 2015 ACSONE SA/NV (<http://acsone.eu>).
 #
 #    All other contributions are (C) by their respective contributors
 #
@@ -45,8 +46,32 @@ class PaymentOrderCreate(models.TransientModel):
     def extend_payment_order_domain(self, payment_order, domain):
         if payment_order.payment_order_type == 'payment':
             domain += [('account_id.type', 'in', ('payable', 'receivable')),
-                       ('amount_to_pay', '>', 0)]
+                       ('credit', '>', 0)]
         return True
+
+    @api.model
+    def filter_lines(self, lines):
+        """ Filter move lines before proposing them for inclusion
+            in the payment order.
+
+        This implementation filters out move lines that are already
+        included in draft or open payment orders. This prevents the
+        user to include the same line in two different open payment
+        orders. When the payment order is sent, it is assumed that
+        the move will be reconciled soon (or immediately with
+        account_banking_payment_transfer), so it will not be
+        proposed anymore for payment.
+
+        See also https://github.com/OCA/bank-payment/issues/93.
+
+        :param lines: recordset of move lines
+        :returns: list of move line ids
+        """
+        payment_lines = self.env['payment.line'].\
+            search([('order_id.state', 'in', ('draft', 'open')),
+                    ('move_line_id', 'in', lines.ids)])
+        to_exclude = set([l.move_line_id.id for l in payment_lines])
+        return [l.id for l in lines if l.id not in to_exclude]
 
     @api.multi
     def search_entries(self):
@@ -69,7 +94,7 @@ class PaymentOrderCreate(models.TransientModel):
         # -- end account_direct_debit --
         lines = line_obj.search(domain)
         context = self.env.context.copy()
-        context['line_ids'] = lines.ids
+        context['line_ids'] = self.filter_lines(lines)
         context['populate_results'] = self.populate_results
         model_datas = model_data_obj.search(
             [('model', '=', 'ir.ui.view'),
@@ -129,7 +154,7 @@ class PaymentOrderCreate(models.TransientModel):
                 'amount_to_receive' in line):
             amount_currency = line.amount_to_receive
         else:
-            amount_currency = line.amount_to_pay
+            amount_currency = line.amount_residual_currency
         line2bank = line.line2bank(payment.mode.id)
         # -- end account banking
         res = {'move_line_id': line.id,
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -45,9 +45,22 @@ class PaymentOrderCreate(models.TransientModel):
     @api.model
     def extend_payment_order_domain(self, payment_order, domain):
         if payment_order.payment_order_type == 'payment':
-            domain += [('account_id.type', 'in', ('payable', 'receivable')),
-                       ('credit', '>', 0)]
-        return True
+            # For payables, propose all unreconciled credit lines,
+            # including partially reconciled ones.
+            # If they are partially reconciled with a supplier refund,
+            # the residual will be added to the payment order.
+            #
+            # For receivables, propose all unreconciled credit lines.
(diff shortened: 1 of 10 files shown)


account_banking_payment_export/wizard/payment_order_create.py
```diff
@@ -45,9 +46,33 @@ def default_get(self, field_list):
     def extend_payment_order_domain(self, payment_order, domain):
         if payment_order.payment_order_type == 'payment':
             domain += [('account_id.type', 'in', ('payable', 'receivable')),
-                       ('amount_to_pay', '>', 0)]
+                       ('credit', '>', 0)]
```
pedrobaeza: This is the main reason why we can't add refunds. We can redo this making:

```
...
'|',
'&',
('credit', '>', '0'),
('type', '=', 'in_invoice'),
'&',
('credit', '<', '0'),
('type', '=', 'in_refund')]
```

sbidoul: Can you elaborate the scenario that causes problems for you? In my understanding credit or debit can never be negative in Odoo.

So we have the following:
- supplier invoice: credit > 0 on payable account
- supplier refund: not something to be paid on a credit transfer order
- customer invoice: not something to be paid on a credit transfer order
- customer refund: credit > 0 on receivable account

alexis-via:
@sbidoul  I tested your PR and it works well.

But I have a question: I see you added a button "Partial reconciles counter" on payment order to see the number of lines that you need to reconcile manually. Don't you think that we should be able to do it automatically ? I mean, if I have the module account_banking_payment_transfer:
1) I create a supplier invoice amount 100 for agrolait
2) I create a supplier refund amount 20 for agrolait
3) I partially reconcile the invoice and refund (reconcile mark A1)
4) I create a payment order and I select the account move line of the invoice of agrolait: it creates a payment line of 80 => your PR does that very well.
5) when I valiade the payment order, an account move will be generated from agrolait's supplier account to the transfer account for an amount of 80. We should be able to delete the reconcile mark A1 and re-create a new reconcile mark A2 with the same lines as A1 plus the line of the transfer move: if all the lines of the reconcile mark A2 are balanced, A2 will be a "full reconcile" mark (it is the case in this scenario) ; otherwise, A2 will be a partial reconcile mark (like A1), and we will be able to select the invoice again in another payment order to pay the residual.

I think it is possible to develop that. And I think it's what accountants would expect. I can help you implement that if you think it's possible.

sbidoul:
Hi Alexis,

Yes the scenario you describe in 5) should be automatic. Actually, I thought it was the case already. So if it's not we should develop it (in another PR I suppose).

But the scenario that led us to create that button is the following (see also #101):
- accountant selects invoice to pay
- accountant modifies the amount to pay on the payment line because he can benefit from a discount for early payment (actually, we are preparting a set of modules that help doing that)
- the system does the transfer move and a partial reconcile with the invoice
- the residual amount must be put on an income account: the new button introduced in this PR is there to facilitate that by giving one click access to the partially reconciled transfer moves so the user can do the write-off easily and make sure the invoice will not be proposed again for payment

adrienpeiffer:
@alexis-via changes in b5e571f allow automatic reconciliation with the transfer move even if move line linked on the payment line is already partially reconciled

matthieuchoplin:
Hi,
thanks for your work.
I would like to add some tests in the same idea of this PR: https://github.com/OCA/bank-payment/pull/136 according to the test scenario described by @alexis-via above.
I cannot checkout this PR at the moment (because of write access). Do you know an other way to do it?
I think that with automated tests, it will be easier to decide to merge a PRs.
What do you think?

sbidoul:
@matthieuchoplin thanks a lot for proposing to add test. This is most welcome.

Since the test you propose is specific to this PR, I suggest you do it as a PR to branch acsone:8.0-bank-payment-fix93-sbi. I'll review it and merge it into this PR.

sbidoul:
I committed a more sophisticated filter on receivables to filter out partially reconciled customer credit notes. 

And rebased to get the latest pep8 fixes.

matthieuchoplin:
Test have been proposed for the use case stated in https://github.com/OCA/bank-payment/issues/93 here https://github.com/acsone/bank-payment/pull/6

sbidoul:
The tests from @matthieuchoplin are merged and I rebased.

sbidoul:
@pedrobaeza @alexis-via your comments have been cared for. This is working fine in production at several customers. Looking forward to your reviews.

account_banking_payment_transfer/model/account_payment.py
```diff
@@ -253,3 +268,22 @@ def action_sent(self):
         # State field is written by act_sent_wait
         self.write({'date_sent': fields.Date.context_today(self)})
         return True
+
+    @api.multi
+    def partial(self):
+        self.ensure_one()
+        view_id = self.env.ref('account.view_move_line_tree').id
```
matthieuchoplin: Don't you want a custom view here instead of taking the one from the account module? I say that because of the issue produced by on_create_write on view_move_line_tree. The domain that you set in reconcile_partial_domain will be replaced by the domain defined in on_create_write. Let me know if I am not clear, else, I described the problem a bit further here: https://github.com/OCA/bank-payment/issues/93
Or do you think of an other solution?

matthieuchoplin:
I would be happy to have this one merged and carry on the improvements in other PRs.
@pedrobaeza : do you still have concern regarding the selection of customer credit notes?
If so, I am not sure about the use case to test, could you please write it here?
Thanks

pedrobaeza:
Well, for now it can be merged by my part, because I need to go deeper to solve my problem, but there's not related to this PR.

:+1: 

Let's see if @alexis-via also agrees about merging.

account_banking_payment_transfer/model/account_payment.py
matthieuchoplin: This issue has been fixed in Odoo at this revision: https://github.com/odoo/odoo/commit/aaf9badbceaf683603b4667dd3d29e2784e78e6f
Then, I delete my comment which is outdated.

matthieuchoplin:
Hi,
FYI, I am also using this code in production.
@alexis-via: do you agree for merging this branch into OCA:8.0?

alexis-via:
yes, let's go :+1:

Issue #111, opened 2015-01-24
8.0 Cannot even generate a SEPA credit transfer file !


alexis-via:
Here is the scenario on v8:
1) install the module account_banking_sepa_credit_transfer and make sure that the module account_banking_sepa_direct_debit is NOT INSTALLED
2) create a new payment order, select a payment mode for SEPA credit transfer, select an invoice, confirm
3) in the wizard pop-up, leave the default parameters and generate the XML file: here you get a crash:

```
2015-01-24 22:45:13,713 8553 ERROR o8_test2 openerp.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 525, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 562, in dispatch
    result = self._call_function(**self.params)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 302, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 299, in checked_call
    return self.endpoint(*a, **kw)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 791, in __call__
    return self.method(*args, **kw)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 395, in response_wrap
    response = f(*args, **kw)
  File "/home/alexis/new_boite/dev/odoo/addons/web/controllers/main.py", line 953, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/alexis/new_boite/dev/odoo/addons/web/controllers/main.py", line 941, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/alexis/new_boite/dev/banking-80/account_banking_sepa_credit_transfer/wizard/export_sepa.py", line 153, in create_sepa
    cr, uid, pain_root, gen_args, context=context)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 241, in wrapper
[...]

PR #112, opened 2015-01-24
8.0 FIX bug #111: crash when generating SEPA credit transfer file

alexis-via:
Please refer to the bug report #111 for the explainations. In fact, the commit just goes back to the code of v7.0, which worked well.

### DIFF
--- a/account_banking_pain_base/models/res_company.py
+++ b/account_banking_pain_base/models/res_company.py
@@ -44,8 +44,8 @@ class ResCompany(orm.Model):
         assert isinstance(company_id, int), 'Only one company ID'
         company = self.browse(cr, uid, company_id, context=context)
         company_vat = company.vat
-        party_identifier = company.sepa_creditor_identifier
-        if not party_identifier and company_vat:
+        party_identifier = False
+        if company_vat:
             country_code = company_vat[0:2].upper()
             if country_code == 'BE':
                 party_identifier = company_vat[2:].replace(' ', '')


alexis-via:
Note that, without this patch, we CANNOT EVEN GENERATE A SEPA CREDIT TRANSFER FILE on v8 !!! We shouldn't be proud of that, so let's patch this rapidly !

guewen:
:+1:

nbessi:
This is delicate the revert is not a full revert of 
Https://github.com/OCA/bank-payment/commit/791370ec7d34505c48c18f48702511cbed2d5bf9

This commit will allows the generate file but will also probably break branch code that depends on it like the 'ES' one. That probably means the fields is required by many module and should be declared else where. I did not find the bug related to the commit so I want to have @pedrobaesa advice on this one. 

An other point is that this modification greenify Travis. That odd and maybe means a test case is missing. 

I would be in favour of merging it. But I would like in parallel to have  a related bug report with corrective steps or a confirmation that code is fine as is.

sbidoul:
As I read it, the version before 791370ec7d34505c4 was broken in the same way for ES but for ES only.

pedrobaeza:
Yeah, that's correct. This is inherited code from 7.0 that I assume correct and make it general taking "partial information". I have to investigate if SEPA creditor identifier is also required for SEPA credit transfer. If so, solution is to move the field from _account_direct_debit_ to _account_banking_pain_base_.

I'll tell you ASAP.

alexis-via:
@pedrobaeza Could we move forward on this ? This bug is really critical, because we can't generate a SEPA credit transfer on v8...

sbidoul:
I agree with @alexis-via, we must merge this now, and provide a more complete in another PR if necessary. We can't leave SEPA credit transfer in a broken state in this repo.

If nobody objects, I'll merge tomorrow.

pedrobaeza:
Sorry for the delay, but I query one expert, and I was waiting for the answer. Let me check tonight by myself, and I'll give a solution.

sbidoul:
Thanks, Pedro!

pedrobaeza:
As I can see in https://businessbanking.bankofireland.com/fs/doc/wysiwyg/sepa-credit-transfer-pain-001-001-03-xml-file-structure-july-2013.pdf (page 8), <InitgPty> section is also mandatory for credit transfer, so we have to make it another way. I'm going to prepare another PR with the following changes:
- Move field to account_banking_pain_base.
- Change export routine to raise an exception if this data is not filled.

I expect to have it tonight.

Stay tuned.

pedrobaeza:
I have made #131 for solving properly this issue. Please check.

dgouthiere:
Dear All, I am a new joiner, My name is David. Please correct me if this comment is not at the right place.
I arrive on this forum because impacted by non-working functionality when exporting SCT payment files.  I read your discussions and believe something is not 100% correct from the start point. 
My recommendation would be  to start from Febelfin pain standard.
InitiatingParty  exists for SCT and SDD.
InitiatingParty is only part of the groupheader and identify the payment initiator - will be a debtor in case  case of SCT and a creditor in case of SDD.  But no matter the role of the initiator because it does not affect the payment  infomation.  It is more an audit information for the receiving bank. But if not correct, the file is rejected by the bank.
The rules are clear in Febelfin  documentation (available on Febelfin website)
• InitiatingParty is mandatory  
• It can be a name (Max70text) AND / OR Identification     (yes, the remitter can mention both info)
• If identification is used, it must be :
o   BICorBEI   = BIC (if financial institution) 
OR
o   Other which consist of two mandatory informations
 Identification (a  Max35 text identification)   AND
 Issuer (a  Max35 text identification)
In Belgium, when the Initiator is a company, Febelfin request Identification to be the Enterprise nr (TVA nr in 10 digits) AND Issuer must be '"KBO-BCE"
I do not know special rules for other countries (ES, ...)
Febelfin is a standard based on ISO20022 - in fact a subset of ISO20022 and respect the same naming convention. My advice would be to keep the same naming in Odoo modules.
InitiatingParty exists for  SCT and SDD, with same layout but however small differences. (issuer is not mandatory for non BE company when Other is used).  
Managing SDD and SCT beeing two different businesses , my recommendation would be to manage the 2 applications separately.

https://www.febelfin.be/en/standards-distance-banking

Perhaps is my understanding not complete, then please let me know
Many thanks for your attention,
David

yvaucher:
#131 was superseeded by #172 

@pedrobaeza Should we close this one?

@dgouthiere Thanks for your input, the code was patched have you tried with the last version? I don't know if your remark was taken into account though.

pedrobaeza:
Yeah, I close it

PR #114, opened 2015-01-28
8.0 [Fix]Fix an error when no payment_order_ids

jlaloux:
[8.0][Fix] an error when no payment_order_ids


pedrobaeza:
In v8, you don't need this artifacts anymore, because False values are propagated through null records. The only doubt I have is about the index getting an error, but if it's the case, you only need to put: `ref = self.payment_order_ids.reference` instead.

jlaloux:
Yes you're right... old habits to forget ^^. And indeed, i encounter this problem because of the index

pedrobaeza:
:+1:

guewen:
:+1:

PR #116, opened 2015-01-29
[account_invoice_bank_usability] New addon to cover usability concepts

omar7r:
Hi,

We found an usability issue on customer invoices, because, we are not able to select customer bank accounts. I didn't know if is only common at Spain but, we should be able to create customer invoices as direct debit.

In addition, this addon contemplates the possibility of customer/supplier bank account was defined in partner parents.

Regards

### DIFF
--- a/account_invoice_bank_usability/__init__.py
+++ b/account_invoice_bank_usability/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/account_invoice_bank_usability/__openerp__.py
+++ b/account_invoice_bank_usability/__openerp__.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Pexego (<http://www.pexego.es>).
+#
+#    All other contributions are (C) by their respective contributors
+#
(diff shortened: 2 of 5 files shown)


pedrobaeza:
This is known issue for so long in the core. You should propose another patch to Odoo to see if this time Odoo accepts it (there was several attempts previously).

omar7r:
Issue report: https://github.com/odoo/odoo/issues/5085

yvaucher:
Any news about this?

I think we should use this module as almost a year later there was no answer from Odoo.

We will also need it in 9.0

yvaucher:
For the records, it's also common in switzerland to need to be able to select the bank on which you want to be paid.

yvaucher:
Ok I misunderstood the purpose of this module. I'm more concerned on the fact that partner_bank_id is not displayed on the invoice

pedrobaeza:
I think this should be included on account_banking_payment_extension and link it with the payment mode selected (payment or debit) to get the correct domain in the bank account.

pedrobaeza:
@omar7r, any advance about my last comment?

pedrobaeza:
Closing as no answer

PR #118, opened 2015-02-01
[8.0] account_voucher_killer: set group category

sbidoul:


pedrobaeza:
If you do this, as this group doesn't imply any other of the category groups, it's going to provoke an odd group in the user's permissions tab. I would put it as part of Technical features category.

:-1:

sbidoul:
Ah, right. I did not notice because the groups defined in OCA/account-financial-tools/account_credit_control do the same.

PR #117, opened 2015-01-29
8.0 Use attachments instead of creating a specific object for each kind of payment.order

alexis-via:
This is an implementation of the discussion that took place in the account-banking-drivers mailing-list, cf https://lists.launchpad.net/banking-addons-drivers/msg00145.html and the answers.

In the end, I decided NOT to create a specific object that would inherit of ir.attachment, because I didn't see any real advantages for this (I implemented this possibility, it worked, but it was more code for no real benefit).

PR #123, opened 2015-02-12
8.0 Use attachments instead of creating a specific object for each payment file

alexis-via:
This is the rescue PR of the previous PR that went out of control after a git rebase.


alexis-via:
I let the git experts do the rebase work this time, I won't do it myself. It's not easy to do, because you have to merge a patch on a file that I deleted in this branch :)

sbidoul:
Original PR was in #117

sbidoul:
I agree with the code and I tested it works fine. Before merging I assume we need the migration script.

alexis-via:
By the way, the module account_banking_fr_lcr has been ported to v8 and moved to l10n-france with the attachement approach (no dedicated object to store the file).

alexis-via:
@sbidoul I just contributed the migration scripts for SEPA Direct debit and SEPA Credit transfer, with the very kind help of @bwrsandman and @legalsylvain

pedrobaeza:
:+1:

bwrsandman:
Does this remove the unidecode requirement from travis?

https://github.com/OCA/bank-payment/blob/8.0/.travis.yml#L23

alexis-via:
@bwrsandman no, we still need unidecode() in account_banking_pain_base/models/banking_export_pain.py

bwrsandman:
ok :+1:

sebastienbeau:
:+1: thanks for the job

PR #131, opened 2015-02-27
[FIX] account_banking_pain_base: SEPA identifier issue

pedrobaeza:
Here it is the PR for resolving issue #111. Please check if migration path is correctly solved


account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -269,15 +269,18 @@ def generate_initiating_party_block(
...
-            iniparty_id = etree.SubElement(initiating_party_1_8, 'Id')
-            iniparty_org_id = etree.SubElement(iniparty_id, 'OrgId')
-            iniparty_org_other = etree.SubElement(iniparty_org_id, 'Othr')
-            iniparty_org_other_id = etree.SubElement(iniparty_org_other, 'Id')
-            iniparty_org_other_id.text = initiating_party_identifier
-            iniparty_org_other_issuer = etree.SubElement(
-                iniparty_org_other, 'Issr')
-            iniparty_org_other_issuer.text = initiating_party_issuer
+        if not initiating_party_identifier or not initiating_party_issuer:
+            raise orm.except_orm(
```
StefanRijnhart, line 273: As announced on #112, you now mandate the presence of the identifier and issuer on sepa credit transfers. But this does not seem to have been an issue before for SCT's in at least NL, BE and FR. Looking at one Dutch bank specification, it does not seem to be mandatory. So in spite of the evindence you found, maybe we should be more pragmatic about this?

sbidoul, line 273: For BE, it was there before, but it is indeed optional in several countries. So I also think raising an exception here is too abrupt.

sbidoul, line 273: And it breaks the test suite :smile:

sbidoul:
Otherwise I like the idea of renaming sepa_creditor_identifier to initiating_party_identifier and moving it to the account_banking_pain_base module.

What would be the migration strategy?

pedrobaeza:
About test suite, it can be adapted for not breaking (I made an attempt), and for the optional question, do you think it's so bad to have this data filled? Most companies will make also SEPA direct debit operations, so they will need this data as well, and for countries where it is mandatory for SEPA credit transfer, you already have it without making a special logic for them.

Another thing is about to raise the exception. I decide it to make it this way because this is the only way to have correct exported files, because previously, the SEPA DD file was generated and silently omit this part if you don't have the field filled, reporting incorrect XML file where issued to the bank.

StefanRijnhart:
Bit of a dilemma here. I'm guessing that the bank does not actually say 'Please regenerate the XML, this time with the correct party identifier for your organization' but instead simply blurts out 'INVALID XML, aborting!'

The method to compose an organization's identifier is actually well defined for the Netherlands, so I guess we can live with making this obligatory after all.

Good point about migration: you can simply rename the column in the pain_base module's pre-migrationscript because it is a dependency of the sepa dd module.

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -269,15 +269,18 @@ def generate_initiating_party_block(
...
+                _("Error"),
+                _("You must fill initiating party identifier and issuer."))
```
sbidoul: Perhaps "You must fill initiating party identifier and issuer on the company form", so the user easiky understands knows how to fix the error.

sbidoul:
At least in Luxemburg, the InitgPty/Id element is optional if the Nm field is present.
http://www.abbl.lu/download/14692/standard-xml-sdd-initiation-v1.pdf section 1.8.

With a clear enough error message, users can easily fill in the missing information and it's a one time correction, so I could live with it if it's important to have it mandatory in other countries.

pedrobaeza:
I have changed the error text and repair the test suite, so please approve and merge.

sbidoul:
@pedrobaeza I finally had the opportunity to dig into direct debit. And I now realize that the sepa creditor identifer is a totally different thing than the initiating party.

Look for instance in this document http://www.abbl.lu/download/14692/standard-xml-sdd-initiation-v1.pdf:
- initiating party is in the header: <InitgPty> tag
- sepa creditor identifier is in the PmtInf block: <CdtrSchmeId>  tag
  two very different things and we need both.

So I'm afraid this branch won't work :-1: 

We need to merge #112.

sbidoul:
@pedrobaeza I had a look at this document for spain: http://www.aebanca.es/cs/groups/public/documents/folletos/folleto-201305190.pdf

From what I understand, in spain the sepa creditor indentifier and initiating party are the same. Is it right?

But it's not the case in other countries (at least BE, LU, FR).

So perhaps you can slightly adapt this PR to leave the sepa_creditor_identifier alone.

Then we can add some helpers in the localizations to hide some fields or set default values.

What do you think?

account_banking_pain_base/models/res_company.py
```diff
@@ -62,19 +45,82 @@ def _initiating_party_issuer_default(self, cr, uid, context=None):
...
+        logger.debug(
+            "SEPA ID check after_replacement = %s" % after_replacement)
+        return int(ipi[2:4]) == (98 - (int(after_replacement) % 97))
 
-    _defaults = {
-        'initiating_party_issuer': _initiating_party_issuer_def,
-    }
+    @api.one
+    @api.constrains('initiating_party_identifier')
+    def _check_initiating_party_identifier(self):
```
sbidoul, line 121: should be _check_sepa_creditor_identifier and go into account_banking_sepa_direct_debit

account_banking_pain_base/models/res_company.py
```diff
@@ -62,19 +45,82 @@ def _initiating_party_issuer_default(self, cr, uid, context=None):
...
+        help="This will be used as the 'Initiating Party Issuer' in the "
+        "PAIN files generated by Odoo.", default=_initiating_party_issuer_def)
+    initiating_party_identifier = fields.Char(
+        string='SEPA Initiating Party Identifier', size=35,
+        help="Enter the SEPA Initiating Party Identifier that has been "
+             "attributed to your company to make SEPA operations. This "
+             "identifier is composed of :\n- your country ISO code (2 "
+             "letters)\n- a 2-digits checkum\n- a 3-letters business "
+             "code\n- a country-specific identifier",
+        oldname="sepa_creditor_identifier")
```
sbidoul, line 71: remove oldname

account_banking_sepa_direct_debit/demo/sepa_direct_debit_demo.xml
```diff
@@ -13,7 +13,8 @@
 </record>
 
 <record id="base.main_company" model="res.company">
-    <field name="sepa_creditor_identifier">FR78ZZZ424242</field>
```
sbidoul, line 16: preserve sepa_creditor_identifier

account_banking_sepa_direct_debit/models/res_company.py
```diff
@@ -20,61 +20,11 @@
...
 class ResCompany(models.Model):
     _inherit = 'res.company'
 
-    sepa_creditor_identifier = fields.Char(
-        string='SEPA Creditor Identifier', size=35,
-        help="Enter the Creditor Identifier that has been attributed to your "
-             "company to make SEPA Direct Debits. This identifier is composed "
-             "of :\n- your country ISO code (2 letters)\n- a 2-digits "
-             "checkum\n- a 3-letters business code\n- a country-specific "
-             "identifier")
```
sbidoul, line 38: preserve sepa_creditor_identifier

pedrobaeza:
Superseeded by https://github.com/OCA/bank-payment/pull/172

PR #132, opened 2015-02-27
This module will make the invoice paid when you finish the payment order workflow

matthieuchoplin:
I had a look at the module https://github.com/OCA/account-payment/tree/8.0/__unported__/account_payment_extension that has not been ported to 7 but this module is working differently.
account_payment_extension was creating a journal entry and doing the reconciliation directly there.
This module is creating the bank statement reconciliation and use the method of Odoo core.

Most of the feature happen in the method process_reconciliation and button_confirm_bank (that I call from the Odoo account_payment module).

### DIFF
--- a/account_payment_pay_invoice/__init__.py
+++ b/account_payment_pay_invoice/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Account Payment that pays invoices module for OpenERP
+#    Copyright (C) 2015 VisionDirect (http://www.visiondirect.co.uk)
+#    @author Matthieu Choplin <matthieu.choplin@visiondirect.co.uk>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


sbidoul:
Hi,

Thanks for contributing. 

So far I have just had time to read the description of your module and not the implementation.

It looks to me what you want to accomplish is obtained by installing account_banking_payment_transfer and configuring a transfer account on the payment mode. With this approach the invoice is reconciled with the payment as soon as the payment is sent (to the bank).

So at first glance I'd say it reaches the same goal.

You may also want to try PR #110 which improve the selection of invoice to pay.

What do you think?

matthieuchoplin:
I can see that travis is failing because of this error:

```
2015-02-27 13:12:39,273 23706 ERROR openerp_test openerp.tools.yaml_import: "name 'action_sent' is not defined" while evaluating
```

But the method that is called is action_sent, which is not in the 8.0 branch but in the 6.1: https://github.com/OCA/bank-payment/blob/6.1/account_banking/account_banking.py

So, how do I know if I have to fix anything? The tests are passing locally. Or maybe this is because travis installed all the modules of the repository and tries to call the method "action_sent" within account_banking_payment_transfer?

Thanks @sbidoul , yeah I feared to replicate what already exists, I have a look at account_banking_payment_transfer and the PR.

matthieuchoplin:
Yes, I just tested account_banking_payment_transfer and it seems to do what i want.
Thanks! 

I close this PR.

PR #133, opened 2015-02-27
[FIX] Voucher is created with wrong bank journal

StefanRijnhart:
For some reason, the voucher's default function to retrieve the journal from the context is triggered even when creating a voucher with a journal_id specified in its values.


lfreeke:
:+1: test

pedrobaeza:
:+1:

lmignon:
:+1:

PR #136, opened 2015-02-28
Yml test for a simple invoice to payment flow

matthieuchoplin:
I have tested the "good path" here:
- create the invoice and validate it
- create the payment order with a payment mode that have a journal and account transfer
- associate invoices (account_move_line) to the payment order
- confirm payment
- create bank statement and reconcile with the account_move_line of the transfer entry
  => check each step with state and amount

If relevant, I will do the other path (more use cases with partial payment, multiple invoices, multicurrency) as well.

Also, I do not know if you prefer python tests or yml tests. For me, the yml tests are good for the "system test" (that test real/business cases) and python unittest are more for testing each python the most isolated as possible. Please tell me what you think.

Thanks.


sbidoul:
Hi,

There is also a python test suite in module account_banking_tests. This is done in a separate module as it depends on different, possibly independent other modules.

Your yml test specific to account_banking_payment_transfer looks good too, so :+1: especially since you want to add more detailed tests!

matthieuchoplin:
This test and other tests have been proposed for the PR https://github.com/OCA/bank-payment/pull/110 through the PR https://github.com/acsone/bank-payment/pull/6 and I think it is better to have all of them in one place for now to avoid merge conflicts.
So, this is why I close that one to focus on adding test to https://github.com/OCA/bank-payment/pull/110

PR #139, opened 2015-03-09
[FIX] account_payment_partner: Fix for finding account move lines without invoice (manual entries in receivables/payables)

pedrobaeza:
This is a monkey patch for overriding the problem with the `('invoice', '=', False)` search, that doesn't work well on core. This should be handled as a patch on core, but I think this would be quicker for now, although I will do the corresponding issue.

@alexis-via, this is what you commented me that you didn't get on a past PR. Please review.

### DIFF
--- a/account_payment_partner/wizard/payment_order_create.py
+++ b/account_payment_partner/wizard/payment_order_create.py
@@ -30,8 +30,29 @@ class PaymentOrderCreate(models.TransientModel):
     def extend_payment_order_domain(self, payment_order, domain):
         res = super(PaymentOrderCreate, self).extend_payment_order_domain(
             payment_order, domain)
-        domain += ['|', '|',
-                   ('invoice', '=', False),
+        # Monkey patch for fixing problem with the core search function
+        # when args has ('invoice', '=', False), referred in the issue #4857
+        # (https://github.com/odoo/odoo/issues/4857)
+        #
+        # Original domain:
+        # domain += ['|', '|',
+        #            ('invoice', '=', False),
+        #            ('invoice.payment_mode_id', '=', False),
+        #            ('invoice.payment_mode_id', '=', payment_order.mode.id)]
+        self.env.cr.execute(
+            "SELECT l.id "
+            "FROM account_move_line l "
+            "LEFT OUTER JOIN account_invoice i "
+            "ON l.move_id = i.move_id "
+            "INNER JOIN account_account a "
+            "ON a.id = l.account_id "
+            "WHERE i.id IS NULL"
+            "  AND l.reconcile_id IS NULL"
+            "  AND a.type in ('receivable', 'payable')")
+        ids = [x[0] for x in self.env.cr.fetchall()]
+        domain += ['|',
+                   ('id', 'in', ids),
+                   '|',
                    ('invoice.payment_mode_id', '=', False),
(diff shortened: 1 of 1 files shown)


sbidoul:
@pedrobaeza is this to work around https://github.com/odoo/odoo/issues/4857 ?

I suggest putting a comment in the code with a reference to the issue, and leave the original code `domain += ['|', '|', ('invoice', '=', False),` commented out so it's easy to restore it when the upstream issue is fixed.

pedrobaeza:
Yeah, exactly!

pedrobaeza:
@sbidoul, I have addressed your comments.

account_payment_partner/wizard/payment_order_create.py
```diff
@@ -30,8 +30,25 @@ class PaymentOrderCreate(models.TransientModel):
...
+        # domain += ['|', '|',
+        #            ('invoice', '=', False),
+        #            ('invoice.payment_mode_id', '=', False),
+        #            ('invoice.payment_mode_id', '=', payment_order.mode.id)]
+        self.env.cr.execute(
+            'SELECT l.id '
+            'FROM account_move_line l '
+            'LEFT OUTER JOIN account_invoice i '
+            'ON l.move_id = i.move_id '
+            'WHERE i.id IS NULL')
```
sbidoul: Now that I read it again, I realize this will build in memory the list of all move lines not linked to an invoice. 
That would break with a very large number of moves. So I'm afraid it's a :-1:

pedrobaeza: The memory footprint is very short and only for the live of the query, so I beg you to reconsider your decission. There's no other way to make it, and remember that this breaks standard import functionality that works without this module. This should be only temporary also.

sbidoul: @pedrobaeza I've several database where this query returns more than 1 million ids. This cannot possibly work.
Perharps if you further filter on payable|receivable and reconcile_id is null this could work?

BTW, have you tried Alexis' fix for https://github.com/odoo/odoo/issues/4857?

pedrobaeza: Hi, @sbidoul, which fix do you talk about? I haven't seen any, and as I said, I see no other possibility: do this "in ids" trick here or in search method, but with the same drawbacks.

sbidoul: @pedrobaeza I was talking about this: https://github.com/OCA/OCB/pull/139

pedrobaeza: Well, this patch avoid the problem simply storing the field in DB and making regular searches. I can completely overwrite the invoice field, but that was one of the things we try to avoid in the past. Doesn't it?

sbidoul: Yes... I suggest you extend your SELECT to filter on payable/receivable accounts and reconcile_id=False. That should limit the number of id's returned to a more reasonable size.

pedrobaeza: I have done it the way you told me. Please check.

pedrobaeza:
@sbidoul, fixed!

sbidoul:
:+1: 

@alexis-via your input is welcome here.

StefanRijnhart:
Strictly speaking not a monkey patch, but :+1:

PR #140, opened 2015-03-11
Limit lenght of mandate reference, according to ISO 20022

alexis-via:
This limitation of size=35 was configured in the past in my code in account_banking_sepa_direct_debit/account_banking_sdd.py, but it was dropped by mistake later on, cf:

http://bazaar.launchpad.net/~banking-addons-drivers/banking-addons/direct-debit-refactoring-fosdem/view/head:/account_banking_sepa_direct_debit/account_banking_sdd.py

### DIFF
--- a/account_banking_sepa_direct_debit/models/account_banking_mandate.py
+++ b/account_banking_sepa_direct_debit/models/account_banking_mandate.py
@@ -70,12 +70,14 @@ class AccountBankingMandate(models.Model):
              "active.", default=True)
     original_mandate_identification = fields.Char(
         string='Original Mandate Identification', track_visibility='onchange',
+        size=35,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You should provide a migration script to convert current mandates (or generic ones) to valid numbers (if length is greater than 35)

alexis-via:
@pedrobaeza 
I don't think it is necessary, because the XSD already blocks when the RUM if more than 35 chars. With this modification, it will block earlier in the process.

pedrobaeza:
OK, let's assume that everything is set properly on the mandate :+1:

sbidoul:
:+1:

sbidoul:
Let's merge this one.

PR #83, opened 2014-12-21
7.0 nl ing mt940

NL66278:
Update mt940 import for Dutch bank ING to latest specs.

PR #142, opened 2015-03-13
7.0 rabobank reworked therp

NL66278:
Import for Rabobank MT940 was not working in most cases. This PR includes an earlier pull request to get the ING MT940 working. Furthermore added some more debugging for invalid transactions or statements.

### DIFF
--- a/account_banking_nl_rabo_mt940/account_banking_nl_rabo_mt940.py
+++ b/account_banking_nl_rabo_mt940/account_banking_nl_rabo_mt940.py
@@ -35,7 +35,9 @@ class RaboMT940Parser(MT940):
 
     tag_61_regex = re.compile(
         r'^(?P<date>\d{6})(?P<sign>[CD])(?P<amount>\d+,\d{2})N(?P<type>.{3})'
-        r'(?P<reference>\w{1,16})')
+        r'(?P<reference>MARF|EREF|PREF|NONREF)\s*'
+        r'\n?(?P<remote_account>\w{1,16})?'
+    )
 
     def parse(self, cr, data):
         """Filter Unprintable characters from file data.
@@ -53,6 +55,9 @@ class RaboMT940Parser(MT940):
         self.current_transaction.transferred_amount = (
             str2amount(parsed_data['sign'], parsed_data['amount']))
         self.current_transaction.reference = parsed_data['reference']
+        if parsed_data['remote_account']:
+            self.current_transaction.remote_account = (
+                parsed_data['remote_account'])
 
     def handle_tag_86(self, data):
         """Handle tag 86: transaction details"""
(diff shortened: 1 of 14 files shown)


NL66278:
Supersedes: https://github.com/OCA/bank-payment/pull/83

account_banking_nl_rabo_mt940/account_banking_nl_rabo_mt940.py
```diff
@@ -42,85 +39,57 @@ class RaboMT940Parser(MT940, parser):
...
+        # The alternative would be to split the file and start a new
+        # statement for each 20: tag encountered.
+        stmt = self.current_statement
+        if not stmt.local_currency:
+            stmt.local_currency = data[7:10]
+            stmt.date = datetime.strptime(data[1:7], '%y%m%d')
+            stmt.start_balance = str2amount(data[0], data[10:])
+            stmt.id = '%s-%s' % (
+                self.current_statement.date.strftime('%Y-%m-%d'),
+                self.current_statement.id)
```
StefanRijnhart: You are replacing a method to guarantee a unique identifier with something more static. Is this guaranteed to be unique?

account_banking_nl_rabo_mt940/account_banking_nl_rabo_mt940.py
```diff
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
+"""Implement parser for MT940 files - Rabobank dialect."""
 ##############################################################################
 #
 #    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2014 Therp BV (<http://therp.nl>).
+#    Copyright (C) 2015 Therp BV <http://therp.nl>.
```
StefanRijnhart: Next time don't remove the year but can add the current year, e.g.

```
Copyright (C) 2014-2015
```

Just leave it for now

StefanRijnhart:
Thanks, nice cleanup. Flake8  wants an extra line somewhere and Pylint says there are two print statements.

Curious to hear from @hbrunn as he has strong knowledge of the mt940 format.

yvaucher:
Your 2 first commits are missing the email can you correct this with a rebase and commit --amend ?

account_banking_nl_rabo_mt940/account_banking_nl_rabo_mt940.py
NL66278: The original method get_unique_account_identifier will return the same identifier for the same account in all cases. Maybe it was intended to call get_unique_statement_id instead? But in both cases the cursor is needed to query the database. I think the parser should just use the information in the parsed file itself. 
If there are problems with uniqueness, it should be solved outside of the parser. But I did not see one so far.

StefanRijnhart: Sorry, my mistake. Then there are no problems with uniqueness indeed.

However, does this mean that you are taking the account number out of the statement identifier (i.e. its display name)? Isn't that going to make it a little bit difficult for the user to quickly recognize statements for a particular account?

pedrobaeza:
This is conflicting, so it seems there has been other improvements that makes this not needed. I'm closing this, but if this is still useful, feel free to reopen fixing conflicts.

PR #146, opened 2015-03-21
8.0 port module account_banking_fr_lcr

alexis-via:
I'm using this module in production on v8 for several months via a symlink in the **unported** repo... but that's dirty :) So this small PR moves it out of **unported** :) That way, I won't receive mails any more asking me to port the module to v8... and me answering that the module for v7 works well on v8 ! :)


brother-bernard:
:+1:

sbidoul:
@alexis-via should this not move to OCA/l10n-france?

alexis-via:
@sbidoul I asked this question on the mailing-list, cf the archives:
https://lists.launchpad.net/banking-addons-drivers/msg00129.html

But no decision was taken. So we could keep the module here for v8 and move it to l10n-france for v9 if a clear decision is taken in the meantime.

sbidoul:
I'm still in favour of putting country-specific modules in localisation repos and implement international standards in OCA/bank-\* modules.

@StefanRijnhart have you had a chance to address this topic in the board as mentioned in https://lists.launchpad.net/banking-addons-drivers/msg00129.html?

StefanRijnhart:
Yes, the board did favour localization implementations of frameworks in
the localization repos.

PR #150, opened 2015-03-25
protect import of external dependencies

### LINKED ISSUE #148: unprodected import
https://github.com/OCA/bank-payment/blob/8.0/account_banking_sepa_credit_transfer/models/account_banking_sepa.py import unidecode is not protected and will break server startup

gurneyalex:
closes #148

Odoo won't install an addon if the external dependencies are not met.
However, the python modules of the addons are imported at startup, and the
lack of an external dependency for an external addon will cause a crash,
therefore the import needs to be in a try..except block.

### DIFF
--- a/__unported__/account_banking_fr_lcr/account_banking_lcr.py
+++ b/__unported__/account_banking_fr_lcr/account_banking_lcr.py
@@ -22,7 +22,11 @@
 
 from openerp.osv import orm, fields
 from openerp.addons.decimal_precision import decimal_precision as dp
-from unidecode import unidecode
+
+try:
+    from unidecode import unidecode
+except ImportError:
+    unidecode = None
 
 
 class banking_export_lcr(orm.Model):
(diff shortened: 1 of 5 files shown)


lmignon:
:+1:

PCatinean:
:+1:

nbessi:
:+1:

sbidoul:
Should we not log a warning in such cases? I'm a bit afrais we forget to install it...

lmignon:
@sbidoul, the installation will fail if the python dependecy is not statisfied since the depednecy is declared in the manifest.
in **openerp**.py

``` python
   'external_dependencies': {
        'python': ['unidecode', 'lxml'],
    },
```

lmignon:
@sbidoul I'll make a try to be sure of what I say.. I suddenly doubt

lmignon:
@sbidoul It works as explained.
image

sbidoul:
Ok, thanks for the clarification!

PR #151, opened 2015-03-25
protect import of external dependencies

### LINKED ISSUE #148: unprodected import

gurneyalex:
closes #148 

Odoo won't install an addon if the external dependencies are not met.
However, the python modules of the addons are imported at startup, and the
lack of an external dependency for an external addon will cause a crash,
therefore the import needs to be in a try..except block.

Conflicts:
    account_banking_sepa_credit_transfer/models/account_banking_sepa.py
    account_banking_sepa_direct_debit/models/banking_export_sdd.py

### DIFF
--- a/account_banking_fr_lcr/account_banking_lcr.py
+++ b/account_banking_fr_lcr/account_banking_lcr.py
@@ -22,7 +22,11 @@
 
 from openerp.osv import orm, fields
 from openerp.addons.decimal_precision import decimal_precision as dp
-from unidecode import unidecode
+
+try:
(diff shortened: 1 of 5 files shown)


lmignon:
:+1:

PCatinean:
:+1:

pedrobaeza:
:+1:

PR #153, opened 2015-03-25
move BeautifulSoup import inside the methods

gurneyalex:


lmignon:
:-1: a double 'import' sounds as a terrible hack!  And the next time we need it in an other method we'll add a new import.. plz prefer a try catch on top of the file...

gurneyalex:
@lmignon updated

lmignon:
@gurneyalex :+1:  Thank you!

pedrobaeza:
:+1:

sbidoul:
:+1:

PR #156, opened 2015-04-01
[IMP] pain: do not use InstrPrty for Direct Debit

sbidoul:
In Luxembourg, InstrPrty is not allowed for Direct Debit [1]. The pain.008 files that we generate are therefore rejected. 

The argument is that this element only makes sense for credit transfer. For what I could gather, this element is ignored in other countries.

Therefore I propose to remove it in the case of direct debit.

Any other suggestion on how to handle this are welcome.

[1] http://www.abbl.lu/download/14692/standard-xml-sdd-initiation-v1.pdf

### DIFF
--- a/account_banking_pain_base/models/banking_export_pain.py
+++ b/account_banking_pain_base/models/banking_export_pain.py
@@ -229,7 +229,7 @@ class BankingExportPain(orm.AbstractModel):
             control_sum_2_5 = etree.SubElement(payment_info_2_0, 'CtrlSum')
         payment_type_info_2_6 = etree.SubElement(
             payment_info_2_0, 'PmtTpInf')
-        if priority:
+        if priority and gen_args['payment_method'] != 'DD':
             instruction_priority_2_7 = etree.SubElement(
                 payment_type_info_2_6, 'InstrPrty')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
These banks SEPA parsers so stricts...

I think we should handle this on direct debit module instead on sepa_pain_base, overwriting method, and removing the offending XML node after calling super. This way, we keep modularity the right way.

sbidoul:
@pedrobaeza there are already several places in account_banking_pain_base where the output is adapted depending on the payment method, so I think my patch is consistent with the rest of the code.

See for instance:
- https://github.com/OCA/bank-payment/blob/8.0/account_banking_pain_base/models/banking_export_pain.py#L251
- https://github.com/OCA/bank-payment/blob/8.0/account_banking_pain_base/models/banking_export_pain.py#L314

pedrobaeza:
Yeah, I see, maybe this needs a bit of refactoring for proper isolation, but I'm not going to block this PR because of this.

Cedric-Pigeon:
:+1:

StefanRijnhart:
:+1:. Instruction priority should be safe to leave out for direct debits.

PR #171, opened 2015-06-01
Greenify runbot

sbidoul:


sbidoul:
runbot is green, travis will follow.

@gurneyalex thanks for the pattern-hack-thing.

StefanRijnhart:
As discussed with @gurneyalex, I'm thinking battling this type of warning is a dilemma. But as we do so, this is a pretty good example. Thanks! :+1:

gurneyalex:
:+1: thanks!

PR #172, opened 2015-06-01
Fix bug #111 New field initiating_party_identifier on res.company

alexis-via:
migration script and post-install script provided

Special thanks to @lmignon  for his help on the development of this PR.


lmignon:
@alexis-via with pleasure :smile:

account_banking_pain_base/post_install.py
```diff
@@ -0,0 +1,57 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+def set_default_initiating_party(cr, registry):
+    party_issuer_per_country = {
+        'BE': 'KBO-BCE',  # KBO-BCE = the registry of companies in Belgium
+    }
+    cr.execute('''
```
pedrobaeza: You can avoid SQL using registry, that works like `self.pool`

pedrobaeza:
Thanks for the PR. These are the holes I see:
- There's no guidance when creating a new company for composing this field. You can provide the calculation via onchange.
- There's no exception raise when issuing a DD and this field is not filled. I remind you that at least in Spain, this data is mandatory, and people send their files thinking that all it's correct, but the bank refuses them.

alexis-via:
@pedrobaeza Here are my answers to your remarks:
The code in post_install.py is mutualized with the code of the migration script in which I don't have "registry" as argument, that's why I don't use it.
About the "holes":
1) with @sbidoul, we think that the on_change should be provided in country-specific modules in each OCA localization, because each country has it's own method for composing the field
2) you could inherit the field to make it a required field for spain in a spain-specific module

pedrobaeza:
I see your reasoning. But you're not providing a specific hook for making that things. It's also advisable not to force to create such small modules that people maybe is not going to install.

Can you reconsider?

alexis-via:
@pedrobaeza In fact, you were right that we could use the registry in both post_install.py and in post-migration.py (it would not have been possible with pre-migration.py). So I'll change the code to use the registry, and I'll make it inheritable.

adrienpeiffer:
:+1: If CI becomes well

lmignon:
:+1:  Once travis is green

pedrobaeza:
What about my question about the proper hook to check for the existence of the field?

alexis-via:
@pedrobaeza Are u happy now ? I added a hook.

alexis-via:
@pedrobaeza Are you OK with my hook ? Can we merge ?

alexis-via:
Travis is green, so we are ready to merge !

alexis-via:
@pedrobaeza OK to merge this one ?

pedrobaeza:
Yeah, sure.

:+1:

PR #175, opened 2015-06-05
8.0 Port SEPA modules to new API + fix regression in SEPA direct debit

alexis-via:
Fix an important regression in account_banking_sepa_direct_debit: "Date of Last Debit" was not set any more (regresssion introduced by myself in the merge to move to attachements)

Write date_done on payment.order when account_banking_payment_export is installed without account_banking_payment_transfer -> code clean-up on how write on date_done is managed
Make generation of SEPA direct debit XML file work when account_banking_payment_export is installed without account_banking_payment_transfer (use date_done on payment.order in the method __get_previous_bank) AND when both modules are installed (use date_sent on payment.order)

Add post-install script for date_sent on payment.order

I made a lot of testing on script PR. This PR is even works in the following scenario:
1) install account_banking_sepa_direct_debit (without account_banking_payment_transfer)
2) do a direct debit for C2C
3) create a 2nd bank account for C2C and update the mandate to link it to the new bank account (back to FIRST)
4) install the account_banking_payment_transfer
5) make a new direct debit for C2C: you will have a "amendement indicator" to true in the XML file (thanks to the post-install script in account_banking_payment_transfer that sets the date_sent properly after install !).

### DIFF
--- a/account_banking_pain_base/models/banking_export_pain.py
+++ b/account_banking_pain_base/models/banking_export_pain.py
@@ -41,6 +41,7 @@ logger = logging.getLogger(__name__)
 class BankingExportPain(models.AbstractModel):
     _name = 'banking.export.pain'
 
+    @api.model
     def _validate_iban(self, iban):
         """if IBAN is valid, returns IBAN
         if IBAN is NOT valid, raises an error message"""
@@ -49,6 +50,7 @@ class BankingExportPain(models.AbstractModel):
         else:
             raise Warning(_("This IBAN is not valid : %s") % iban)
 
+    @api.model
     def _prepare_field(self, field_name, field_value, eval_ctx,
                        max_size=0, gen_args=None):
         """This function is designed to be inherited !"""
@@ -92,6 +94,7 @@ class BankingExportPain(models.AbstractModel):
             value = value[0:max_size]
         return value
 
+    @api.model
     def _validate_xml(self, xml_string, gen_args):
         xsd_etree_obj = etree.parse(
             tools.file_open(gen_args['pain_xsd_file']))
@@ -151,6 +154,7 @@ class BankingExportPain(models.AbstractModel):
         }
         return action
 
+    @api.model
     def generate_group_header_block(self, parent_node, gen_args):
         group_header_1_0 = etree.SubElement(parent_node, 'GrpHdr')
         message_identification_1_1 = etree.SubElement(
@@ -177,6 +181,7 @@ class BankingExportPain(models.AbstractModel):
         self.generate_initiating_party_block(group_header_1_0, gen_args)
         return group_header_1_0, nb_of_transactions_1_6, control_sum_1_7
 
+    @api.model
     def generate_start_payment_info_block(
             self, parent_node, payment_info_ident,
             priority, local_instrument, sequence_type, requested_date,
@@ -231,11 +236,13 @@ class BankingExportPain(models.AbstractModel):
         requested_date_2_17.text = requested_date
         return payment_info_2_0, nb_of_transactions_2_4, control_sum_2_5
 
+    @api.model
     def _must_have_initiating_party(self, gen_args):
         '''This method is designed to be inherited in localization modules for
         countries in which the initiating party is required'''
         return False
 
+    @api.model
     def generate_initiating_party_block(self, parent_node, gen_args):
         my_company_name = self._prepare_field(
             'Company Name',
@@ -267,6 +274,7 @@ class BankingExportPain(models.AbstractModel):
                 % self.payment_order_ids[0].company_id.name)
         return True
 
+    @api.model
     def generate_party_agent(
             self, parent_node, party_type, party_type_label,
             order, party_name, iban, bic, eval_ctx, gen_args):
@@ -307,6 +315,7 @@ class BankingExportPain(models.AbstractModel):
             # as per the guidelines of the EPC
         return True
 
+    @api.model
     def generate_party_block(
             self, parent_node, party_type, order, name, iban, bic,
             eval_ctx, gen_args):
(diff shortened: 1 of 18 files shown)


account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -38,21 +38,19 @@
...
+        if self.env['res.partner.bank'].is_iban_valid(iban):
             return iban.replace(' ', '')
         else:
-            raise orm.except_orm(
-                _('Error:'), _("This IBAN is not valid : %s") % iban)
+            raise Warning(_("This IBAN is not valid : %s") % iban)
 
-    def _prepare_field(self, cr, uid, field_name, field_value, eval_ctx,
-                       max_size=0, gen_args=None, context=None):
+    def _prepare_field(self, field_name, field_value, eval_ctx,
```
pedrobaeza, line 54: Decorate with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -38,21 +38,19 @@
 logger = logging.getLogger(__name__)
 
 
-class BankingExportPain(orm.AbstractModel):
+class BankingExportPain(models.AbstractModel):
     _name = 'banking.export.pain'
 
-    def _validate_iban(self, cr, uid, iban, context=None):
+    def _validate_iban(self, iban):
```
pedrobaeza, line 45: Decorate it with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -74,31 +72,27 @@ def _prepare_field(self, cr, uid, field_name, field_value, eval_ctx,
...
-                _('Error:'),
+            raise Warning(
                 _("The '%s' is empty or 0. It should have a non-null value.")
                 % field_name)
         if max_size and len(value) > max_size:
             value = value[0:max_size]
         return value
 
-    def _validate_xml(self, cr, uid, xml_string, gen_args, context=None):
+    def _validate_xml(self, xml_string, gen_args):
```
pedrobaeza, line 98: Decorate it with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -131,44 +124,41 @@ def finalize_sepa_file_creation(
...
             'res_model': self._name,
-            'res_id': ids[0],
+            'res_id': self.ids[0],
             'target': 'new',
         }
         return action
 
-    def generate_group_header_block(
-            self, cr, uid, parent_node, gen_args, context=None):
+    def generate_group_header_block(self, parent_node, gen_args):
```
pedrobaeza, line 158: Decorate it with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -246,26 +231,24 @@ def generate_start_payment_info_block(
         requested_date_2_17.text = requested_date
         return payment_info_2_0, nb_of_transactions_2_4, control_sum_2_5
 
-    def _must_have_initiating_party(self, cr, uid, gen_args, context=None):
+    def _must_have_initiating_party(self, gen_args):
```
pedrobaeza, line 240: Decorate it with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -246,26 +231,24 @@ def generate_start_payment_info_block(
...
         '''This method is designed to be inherited in localization modules for
         countries in which the initiating party is required'''
         return False
 
-    def generate_initiating_party_block(
-            self, cr, uid, parent_node, gen_args, context=None):
+    def generate_initiating_party_block(self, parent_node, gen_args):
```
pedrobaeza, line 246: Decorate it with @api.model for possible calls from a recordset

account_banking_pain_base/models/banking_export_pain.py
```diff
@@ -276,36 +259,33 @@ def generate_initiating_party_block(
...
                     "'Initiating Party Identifier' for the company '%s'. "
                     "Both fields must have a value.")
-                % gen_args['sepa_export'].payment_order_ids[0].company_id.name)
+                % self.payment_order_ids[0].company_id.name)
         return True
 
     def generate_party_agent(
-            self, cr, uid, parent_node, party_type, party_type_label,
-            order, party_name, iban, bic, eval_ctx, gen_args, context=None):
+            self, parent_node, party_type, party_type_label,
```
pedrobaeza, line 279: Decorate it with @api.model for possible calls from a recordset

pedrobaeza:
Thanks for bringing one of the pending modules to new API. In general, it seems good. I have mentioned some methods that can be decorated with @api.model, and you can extend the same for the rest, but it's only for caution.

pedrobaeza:
Now seems good :+1: 

Thanks for the work.

account_banking_pain_base/models/payment_line.py
```diff
@@ -20,33 +20,23 @@
...
-from openerp.osv import orm, fields
+from openerp import models, fields
 
 
-class PaymentLine(orm.Model):
+class PaymentLine(models.Model):
     _inherit = 'payment.line'
 
-    def _get_struct_communication_types(self, cr, uid, context=None):
+    def _get_struct_communication_types(self):
```
sbidoul, line 30: decorate api.model to preserve compatibility with old api overrides

sbidoul:
Thanks for the work!

I noticed one backward compatibility issue while testing. :+1: when that is fixed.

account_banking_pain_base/models/payment_line.py
pedrobaeza, line 30: We can even remove this method and "hard-code" it, because now we can make _selection_add_ in overrides.

sbidoul, line 30: @pedrobaeza that's true but we need to preserve a stable API for downstream extensions. So these kind a changes should be postponed to a next major upgrade.

pedrobaeza, line 30: :+1:

adrienpeiffer:
:+1: (code review)

sbidoul:
The runbot failure is unrelated and should be solved by now. So let's merge it.

PR #176, opened 2015-06-09
[MIG] restored account_payment_multicurrency_extension from 7.0 branch

charbeljc:


account_payment_multicurrency_extension/payment.py
```diff
@@ -0,0 +1,64 @@
...
+    _inherit = 'payment.order.create'
+
+    def create_payment(self, cr, uid, ids, context=None):
+        """
+        We add is_multi_currency tag to be able set
+        'amount_currency': line.amount_residual_currency
+        instead of
+        'amount_currency': line.amount_to_pay
+        To be compliant with multi currency
+        Already corrected in V8 but will not be corrected in V7
```
sbidoul, line 35: If this sentence is true, then this method must not be ported to V8

vrenaville:
@charbeljc @sbidoul I confirm in V8 it's correct now, no need to migrate this module

sbidoul:
According to last comment we can close this one.

PR #180, opened 2015-06-18
[IMP] account_banking_mandate: Add report SEPA mandate.

sergio-teruel:


pedrobaeza:
Excellent addition, Sergio! I have tried and all goes well, except you haven't removed the header and footer on the report definition, leading to a 2-page report. Please change that.

sergio-teruel:
If i remove the header and footer i need a new paper format...

pedrobaeza:
Or just don't put report.html_container but the empty one (I don't remember now which is)

sergio-teruel:
I tried this but it does not work..

pedrobaeza:
You have close this one making the PR of the website_slides. The key is to use new branches for each work, not the same (8.0). That is done with `git checkout -b branch_name`. Then, you can work with that branch, make commits and so on, and you can go back to the main branch with `git checkout 8.0`, and then start a new branch from it if you want. Switching between branches is done with `git checkout` (without -b).

sergio-teruel:
Oks... thanks...
Now I will make a new PR

PR #181, opened 2015-06-19
[IMP] account_banking_mandate: Add report SEPA mandate.

sergio-teruel:


pedrobaeza:
Sergio, you can inherit from report.minimal_layout to get a report without header and footer.

sergio-teruel:
Sorry, I have tried to do it with minimal layout but this layout don't work...

rlizana:
:+1:

pedrobaeza:
Checking on runbot, the report is fixed and there's no more full header, so :+1: and I merge.

Thanks for the work.

PR #182, opened 2015-06-24
[8.0] Improvements for account_banking_payment_transfer

adrienpeiffer:
- Create a function for partners move line creation
- Remove the use of `self` in `@api.model` decoreted method
- Use standard reconcile method

### DIFF
--- a/account_banking_payment_transfer/model/account_payment.py
+++ b/account_banking_payment_transfer/model/account_payment.py
@@ -171,23 +171,24 @@ class PaymentOrder(models.Model):
         return vals
 
     @api.model
-    def _prepare_move_line_partner_account(self, line, move, labels):
+    def _prepare_move_line_partner_account(self, line, move, labels,
+                                           payment_order_type):
         if line.move_line_id:
             account_id = line.move_line_id.account_id.id
         else:
-            if self.payment_order_type == 'debit':
+            if payment_order_type == 'debit':
                 account_id = line.partner_id.property_account_receivable.id
             else:
                 account_id = line.partner_id.property_account_payable.id
         vals = {
             'name': _('%s line %s') % (
-                labels[self.payment_order_type], line.name),
+                labels[payment_order_type], line.name),
             'move_id': move.id,
             'partner_id': line.partner_id.id,
             'account_id': account_id,
-            'credit': (self.payment_order_type == 'debit' and
+            'credit': (payment_order_type == 'debit' and
                        line.amount or 0.0),
-            'debit': (self.payment_order_type == 'payment' and
+            'debit': (payment_order_type == 'payment' and
                       line.amount or 0.0),
             }
         return vals
@@ -198,7 +199,8 @@ class PaymentOrder(models.Model):
         return
 
     @api.model
-    def _create_move_line_partner_account(self, line, move, labels):
+    def _create_move_line_partner_account(self, line, move, labels,
+                                          order_type):
         """This method is designed to be inherited in a custom module"""
 
         # TODO: take multicurrency into account
@@ -207,7 +209,7 @@ class PaymentOrder(models.Model):
         # create the payment/debit counterpart move line
         # on the partner account
         partner_ml_vals = self._prepare_move_line_partner_account(
-            line, move, labels)
+            line, move, labels, order_type)
         partner_move_line = aml_obj.create(partner_ml_vals)
 
         # register the payment/debit move line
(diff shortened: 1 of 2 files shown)


account_banking_payment_transfer/model/account_payment.py
```diff
@@ -197,6 +198,31 @@ def action_sent_no_move_line_hook(self, pay_line):
...
+
+        # register the payment/debit move line
+        # on the payment line and call reconciliation on it
+        line.write({'transit_move_line_id': partner_move_line.id})
+
+    def _reconcile_payment_lines(self, payment_lines):
+        pl_obj = self.env['payment.line']
+        for line in payment_lines:
+            if line.move_line_id:
+                pl_obj.debit_reconcile(line.id)
```
sbidoul: I think you can write line.debit_reconcile() in new api, so no need for pl_obj

account_banking_payment_transfer/model/payment_line.py
```diff
@@ -160,28 +159,9 @@ def is_zero(total):
...
-            }
-
-        if torec_move_line.reconcile_partial_id:
-            reconcile_obj.write(
-                cr, uid, [torec_move_line.reconcile_partial_id.id],
-                vals, context=context)
-        else:
-            reconcile_obj.create(
-                cr, uid, vals, context=context)
+        move_line_obj.reconcile_partial(cr, uid, line_ids, type='manual',
```
sbidoul: Use 'auto' ?

account_banking_payment_transfer/model/payment_line.py
```diff
@@ -160,28 +159,9 @@ def is_zero(total):
...
+                                        context=context)
+
         for line_id in line_ids:
             workflow.trg_trigger(
                 uid, 'account.move.line', line_id, cr)
```
sbidoul: I think this trigger can be removed as it is called by reconcile().
https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L1074

sbidoul:
I wonder why coverage decreases, as this refactoring removes code.

sbidoul:
This solves #184

account_banking_payment_transfer/model/payment_line.py
```diff
@@ -160,28 +159,9 @@ def is_zero(total):
                 cr, uid, transit_move_line.company_id.currency_id, total)
```
pedrobaeza: This method can also be removed

pedrobaeza:
Investigating further on this, I now see why this operation is not done with the proper method: it's because it requires an integrity of the account move before making the reconciliation, and we are calling this method while building the move. We can make 2 things:
- Defer the call to debit_reconcile to the end.
- Continue the same way as it was, documenting the reason.

I have made the code for the first option, and this means also we should revert the optional posting of the move, so if you want, I can propose the patch against this branch.

cc @sbidoul

sbidoul:
@pedrobaeza the first option you mention (_Defer the call to debit_reconcile to the end_) is done in this PR. And in our tests it works fine.

Can you explain the use case that has problems?

sbidoul:
I just made a full functional test and it works fine. 

:+1:

account_banking_payment_transfer/model/account_payment.py
adrienpeiffer: @sbidoul I think debit_reconcile isn't considered like a multi method because there is no `ids` parameter. Therefore, `line.debit_reconcile()` doesn't work. I can change `debit_reconcile` method déclaration to replace `payment_line_id` with `ids` and integrate a loop in this method

sbidoul: @adrienpeiffer would it work by decorating debit_reconcile with @api.cr_uid_id_context ?

adrienpeiffer: great! it works

hbrunn:
@adrienpeiffer @sbidoul what is the rationale behind using `@api.model` functions and then passing them values from self? Wouldn't it be more convenient to use `@api.multi` and `self.ensure_one()` where applicable?

sbidoul:
@hbrunn dunno. Question for @adrienpeiffer when he's back next week.

account_banking_payment_transfer/model/account_payment.py
```diff
@@ -197,6 +198,31 @@ def action_sent_no_move_line_hook(self, pay_line):
...
+
+        # register the payment/debit move line
+        # on the payment line and call reconciliation on it
+        line.write({'transit_move_line_id': partner_move_line.id})
+
+    @api.model
+    def _reconcile_payment_lines(self, payment_lines):
+        for line in payment_lines:
+            if line.move_line_id:
+                line.debit_reconcile()
```
hbrunn, line 222: this breaks for me with `ValueError: "debit_reconcile() takes at least 4 arguments (4 given)"`

hbrunn:
I took the liberty to fix my last comment in #197 and also did some porting to the v8 api there

sbidoul:
Superseded by #197

PR #183, opened 2015-06-24
[8.0][account_banking_payment_transfer] Avoid to post the created move if entry_posted on the journal isn't true

adrienpeiffer:
Avoid to post the created move if `entry_posted` on the journal isn't true

### DIFF
--- a/account_banking_payment_transfer/model/account_payment.py
+++ b/account_banking_payment_transfer/model/account_payment.py
@@ -260,8 +260,10 @@ class PaymentOrder(models.Model):
                     total_amount, move, lines, labels)
                 aml_obj.create(trf_ml_vals)
 
-                # post account move
-                move.post()
+                # consider entry_posted on account_journal
+                if move.journal_id.entry_posted:
+                    # post account move
+                    move.post()
 
         # State field is written by act_sent_wait
         self.write({'date_sent': fields.Date.context_today(self)})


sbidoul:
This is to follow the posting behaviour as configured on the journal, as everywere else.

:+1:

pedrobaeza:
:+1:

lmignon:
:+1:

pedrobaeza:
I have just seen that you cannot reconcile a move that it's not posted, so this patch should be reverted.

sbidoul:
@pedrobaeza I don't understand.
- you _can_ reconcile a move that is not posted
- the posting behaviour is configured on the journal so you can preserve the existing behaviour by adjusting the "skip draft move" (entry_posted) flag on the journal

pedrobaeza:
> - you _can_ reconcile a move that is not posted

No, this line prevents that:
https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L961

And as you need to post, you need that the move is completed (not filled line by line). As I said, I have a working version calling the method at the end.

sbidoul:
@pedrobaeza in my understanding the 'valid' state on move lines means that the move is balanced. It is unrelated to posting.

pedrobaeza:
It's also linked, but in any case, we need to do it at the end, not while making each line, because the balance state will be got when all lines are created.

PR #169, opened 2015-05-28
8.0 bank-payment flow improvements

pedrobaeza:
Include following improvements for giving possibilities importing payment lines:
- Add partners and accounts filters in import lines wizard.
- Rename button "Invoices" to "Journal Items" in payment order to be more coherent with its function.
- Allow to import invoices/journal items separately. Importing journal items, you can still include invoices also to allow to import everything.

@sbidoul, @alexis-via, please check.

PR #186, opened 2015-06-26
[IMP] account_banking_payment_transfer: Use payment line date as the maturity date for the new move

pedrobaeza:
In a chain of payment orders with transfer accounts, new maturity date should be the previous "payment" date.

### DIFF
--- a/account_banking_payment_transfer/model/account_payment.py
+++ b/account_banking_payment_transfer/model/account_payment.py
@@ -154,10 +154,12 @@ class PaymentOrder(models.Model):
             partner_id = payment_lines[0].partner_id.id
             name = _('%s line %s') % (
                 labels[self.payment_order_type], payment_lines[0].name)
+            date_maturity = payment_lines[0].date
         else:
             partner_id = False
             name = '%s %s' % (
                 labels[self.payment_order_type], self.reference)
+            date_maturity = False
         vals = {
             'name': name,
             'move_id': move.id,
@@ -167,7 +169,8 @@ class PaymentOrder(models.Model):
                        amount or 0.0),
             'debit': (self.payment_order_type == 'debit' and
                       amount or 0.0),
-            }
+            'date_maturity': date_maturity,
+        }
         return vals
 
     @api.model
--- a/account_banking_payment_transfer/model/account_payment.py
+++ b/account_banking_payment_transfer/model/account_payment.py
@@ -154,12 +154,11 @@ class PaymentOrder(models.Model):
             partner_id = payment_lines[0].partner_id.id
             name = _('%s line %s') % (
                 labels[self.payment_order_type], payment_lines[0].name)
-            date_maturity = payment_lines[0].date
         else:
             partner_id = False
             name = '%s %s' % (
                 labels[self.payment_order_type], self.reference)
-            date_maturity = False
+        date_maturity = payment_lines[0].date
         vals = {
             'name': name,
             'move_id': move.id,


pedrobaeza:
@sbidoul @alexis-via Again I ask you to check this ASAP, because I'm trying to refine these modules so spanish localization can use it for its financial management

sbidoul:
@pedrobaeza can you explain the use case?

alexis-via:
I'm a bit surprised to use date_maturity in an account move line for payment (usually used in the account move line of an invoice)... but why not. It is true that we are lacking in the account move line generated by the payment order the date requested for the execution of the payment.
:+1:

pedrobaeza:
What a busy day I'm having...

Let me explain the use case: in Spain it's very common what it is called a financial management called "gestión de efectos", what can be translated literally as "effects management", but I think this translation doesn't make sense.

This management sets a full cycle of the receivable and payable by steps. For example:
- You issue an invoice, and you have the debt in the customer account (in Spain, 430000), and the maturity date is the date when the customer has promised to pay.
- The payment mode is set with a "pagaré" (something similar to a check). In this case, you move the debt to another account "431000 Efectos comerciales en cartera", and you set the date maturity for this new move to the date the customer finally has said to you that it's going to be able to pay (spanish people pay very bad, hehe).
- Then you can go with the check to the bank, and move the debt to the account "431200 Efectos comerciales en gestión de cobro", with the maturity date as the expected date the income will be done by the bank.
- Finally you check with your bank statement the income of the money, and cancel this last balance with the bank account.

As you can see, the management it's very tedious and long, and the maturity date is a key data to control if all goes on time. That's why I see this very convenient. I have to refine even more, because I'm seeing the payment date behaviour is not too much consistent for the moment, so I'll make another commit in short.

The process can even be worst, because the check amount you have put in the bank can be anticipated by the bank before the due date (charging you with commissions of course), and you need to make a double move with a debt in the bank and in the customer, that must be balanced automatically when the maturity date arrives. Uf, too much things...

But for the moment, this is the first step. I think you don't have this complicated management, and that's why the module are kept this way. Or do you?

As a side question? Why transit accounts are regular ones? In this flow, it's very convenient that all involved accounts are 'payable' or 'receivable', to still see the total to pay/receive in the customer/supplier independently of the financial state of the "effect".

account_banking_payment_transfer/model/account_payment.py
```diff
@@ -154,10 +154,12 @@ def _prepare_move_line_transfer_account(
             partner_id = payment_lines[0].partner_id.id
             name = _('%s line %s') % (
                 labels[self.payment_order_type], payment_lines[0].name)
+            date_maturity = payment_lines[0].date
         else:
             partner_id = False
             name = '%s %s' % (
                 labels[self.payment_order_type], self.reference)
+            date_maturity = False
```
sbidoul: I think you can use payment_lines[0].date in this part too, because lines will be grouped by execution date.

sbidoul:
Thanks for the detailed explanation, Pedro. I don't have met such a complex process indeed. I just made a small comment above, can you have a look?

Regarding your question about the type of the transfer account, it cannot be payable/receivable otherwise the open items in these account would be proposed for payment too and we don't want that.

pedrobaeza:
The issue you mention with payable/receivable account is before of https://github.com/OCA/bank-payment/pull/169. With this improvement, you can pre-filter accounts and select indepently invoices, so you have the full control of what you want to import.

Take also in care that you are generating wrong information with your approach: if you check the total to receive/to pay in a partner, the information will not be correct, because it doesn't include the amount in transit.

account_banking_payment_transfer/model/account_payment.py
pedrobaeza: Thanks for the tip. Updated!

sbidoul:
I still need to review #169 indeed. But IMO transit accounts are not payable/receivable, they are closer to bank/cash accounts. For instance you don't want to send payment reminders for open items in transit accounts.

Anyway :+1: for this PR, thanks!

pedrobaeza:
Well, in our case (and probably yours), it's not crazy that a partner due statement/follow-up contains the amounts that are not yet received, although they're in transit, because this means that the payment is not yet completed. Next step we are going to work is to show in a pretty way these possible states.

pedrobaeza:
For now, I merge this and we can discuss the rest later.

PR #188, opened 2015-06-26
[FIX] don't expect columns to be in the right order

hbrunn:
and give default value for newly added required column


pedrobaeza:
Runbot warnings unrelated and very needed fix to not breaking migrations, so I fast track.

PR #190, opened 2015-06-30
[IMP][account_banking_payment_transfer] Consider invoice reference on move line

adrienpeiffer:
Consider invoice reference for the move line on transfer account if there is only one payment line


pedrobaeza:
I would prefer to have both references in the move name. For example: `payment_line_name (Invoice ref)`

adrienpeiffer:
@pedrobaeza Thanks ! It's done

pedrobaeza:
Thanks :+1:

hbrunn:
:+1:

PR #193, opened 2015-07-02
ADD description on payment mode and show it on invoice report

angelmoya:

PR #194, opened 2015-07-06
[ADD] module 'account_invoice_report_payment_mode'.

JordiBForgeFlow:
This module adds several fields to the invoice form, below the Fiscal Position:
- The Payment Mode,
- The Account Number associated to the selected Payment Mode. 

Hello, I'm not sure if this module would be valuable to the community, because many people choose to design their own invoice forms. I can see it's value when someone wants to re-use the existing invoice, with minimal changes. Or as a how-to reference, when someone wants to develop a full custom invoice form.


pedrobaeza:
This is exactly the same that https://github.com/OCA/bank-payment/pull/193 does.

JordiBForgeFlow:
Thanks @pedrobaeza. I'm then closing this PR.

PR #195, opened 2015-07-08
[FIX] account_banking_sepa_direct_debit: Fix error in payment order …

sergio-teruel:
…validation when there is more than one line with the same mandate.


pedrobaeza:
:+1: I merge due to the criticity of this patch.

PR #196, opened 2015-07-08
[RFR] decorate overridable methods in wizard as api.multi

hbrunn:
this way, those functions can react on fields they put into the wizard

### DIFF
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -130,10 +130,11 @@ class PaymentOrderCreate(models.TransientModel):
                 'target': 'new',
                 }
 
-    @api.model
+    @api.multi
     def _prepare_payment_line(self, payment, line):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why change that methods? They are not called with ids argument

hbrunn:
@pedrobaeza they are called in `search_entries`, which is `api.multi` itself. So just changing the decorator gives you the ids for free when calling them.

And why do this? I'm now working on allowing my users to select  some journal or analytic account in the wizard, so that they don't have to unset the populate checkbox and filter for the journal/account in question by hand. And with this refactoring, that's a lot more convenient to program.

pedrobaeza:
OK, :+1: 

Please check https://github.com/OCA/bank-payment/pull/169 that improves the pre-selection dialog in the same way as you want, but with partners and general accounts

guewen:
:+1:

adrienpeiffer:
:+1:

PR #198, opened 2015-07-12
[FIX] account_banking_payment_transfer: Use selection_add for state overwrite

rlizana:
Use `selection_add` for state overwrite in `payment.order`


sbidoul:
:+1:

pedrobaeza:
:+1: Merging as trivial

PR #200, opened 2015-07-15
[IMP] store journal_entry_ref to enable sorting

hbrunn:
journal_entry_ref only makes a difference to move_id when we also allow draft moves, so I moved this field to the respective module. I also made it a stored field in order to allow sorting


sbidoul:
It make sense to move this to account_payment_include_draft_move indeed.

I'm always a bit reluctant to add stored fields to account.move.line for performance reasons. Is it important to you to have this field sortable?

hbrunn:
@sbidoul yes, but I see no problem in moving that flag in its own customer specific module, I'll update the PR some time soon

pedrobaeza:
@hbrunn are you going to update this PR or should I close it?

hbrunn:
I'll close myself

PR #201, opened 2015-07-23
[ADD] German pain.001.003.03 format

mncoon:
However, in Germany we don't use pain.001.001.03 for SEPA-credit-files.
Our banks decided to use an other standard (pain.001.003.03) which is not described in the ISO 20022. But the needed XML-File is exactly the same as for pain.001.001.03.

To use this module in Germany I added this format.
More informations about the german standards here (only in german):
http://www.ebics.de/
http://www.die-deutsche-kreditwirtschaft.de/uploads/media/SRZ-Richtlinien_2015.pdf

If there are questions, please don't hesitate to ask.
Thanks, Mathias

### DIFF
--- a/account_banking_sepa_credit_transfer/data/payment_type_sepa_sct.xml
+++ b/account_banking_sepa_credit_transfer/data/payment_type_sepa_sct.xml
@@ -40,6 +40,14 @@
     <field name="ir_model_id" ref="model_banking_export_sepa_wizard"/>
 </record>
 
+<record id="export_sepa_sct_001_003_03" model="payment.mode.type">
+    <field name="name">SEPA Credit Transfer pain 001.003.03 (only for Germany)</field>
+    <field name="code">pain.001.003.03</field>
+    <field name="payment_order_type">payment</field>
+    <field name="suitable_bank_types"
+       eval="[(6,0,[ref('base_iban.bank_iban')])]" />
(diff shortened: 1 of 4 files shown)


sbidoul:
I can't comment on the Germany-specific aspects, but the code looks good.

:+1:

account_banking_sepa_credit_transfer/data/payment_type_sepa_sct.xml
```diff
@@ -40,6 +40,14 @@
     <field name="ir_model_id" ref="model_banking_export_sepa_wizard"/>
 </record>
 
+<record id="export_sepa_sct_001_003_03" model="payment.mode.type">
+    <field name="name">SEPA Credit Transfer pain 001.003.03 (only for Germany)</field>
```
pedrobaeza: Change text "only for Germany" with "used in Germany", because as a standard, it can be used elsewhere.

pedrobaeza:
A little comment, but :+1:

hbrunn:
:+1:
@copmn in the future, please create feature branches for PRs. If a request lingers for a while and you happen to push other commits to your master branch, the PR is messed up and not usable any more

mncoon:
@pedrobaeza thank you, it is corrected
@hbrunn thanks for hint, I know that and if needed I work with feature branches.

PR #202, opened 2015-07-28
[FIX] multicompany

vrenaville:
field related use sudo when they are called, so in multi-company,it will take the company currency of the user with id 1 that may differ
from the current user, so I replace the field related
by a field function that called the original _currency function to
keep the context

### DIFF
--- a/account_import_line_multicurrency_extension/models/bank_statement.py
+++ b/account_import_line_multicurrency_extension/models/bank_statement.py
@@ -26,7 +26,8 @@ class AccountBankStatementLine(models.Model):
 
     currency_symbol = fields.Char(
         string='Journal Currency',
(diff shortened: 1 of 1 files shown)


lepistone:
Have you tried to pass `related_sudo=False` to the field definition? If it doesn't work, we should file an issue on odoo/odoo and ping Raphaël.

Thanks!

vrenaville:
@lepistone thanks for pointing me the issue, I have change my PR to add to the field the related_sudo=false instead of redefine a new function field

pedrobaeza:
:+1:

lepistone:
:+1:

mdietrichc2c:
:+1:

PR #203, opened 2015-07-29
[8.0][FIX] According with the decorator change on super method

adrienpeiffer:
According with the decorator change on super method (@api.model -> @api.multi) in #196

### DIFF
--- a/account_payment_blocking/model/payment_order_create.py
+++ b/account_payment_blocking/model/payment_order_create.py
@@ -27,8 +27,8 @@ class payment_order_create(orm.TransientModel):
     _inherit = 'payment.order.create'
 
     def extend_payment_order_domain(
-            self, cr, uid, payment_order, domain, context=None):
+            self, cr, uid, ids, payment_order, domain, context=None):
         super(payment_order_create, self).extend_payment_order_domain(
-            cr, uid, payment_order, domain, context=context)
+            cr, uid, ids, payment_order, domain, context=context)
         domain += [('blocked', '!=', True)]
         return True
--- a/account_payment_mode_term/models/payment_order_create.py
+++ b/account_payment_mode_term/models/payment_order_create.py
@@ -31,9 +31,9 @@ class payment_order_create(orm.TransientModel):
     _inherit = 'payment.order.create'
 
     def extend_payment_order_domain(
-            self, cr, uid, payment_order, domain, context=None):
+            self, cr, uid, ids, payment_order, domain, context=None):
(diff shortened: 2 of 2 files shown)


account_payment_blocking/model/payment_order_create.py
```diff
@@ -27,8 +27,8 @@ class payment_order_create(orm.TransientModel):
     _inherit = 'payment.order.create'
 
     def extend_payment_order_domain(
-            self, cr, uid, payment_order, domain, context=None):
+            self, cr, uid, ids, payment_order, domain, context=None):
```
pedrobaeza, line 30: Isn't better to migrate this to new API?

account_payment_mode_term/models/payment_order_create.py
```diff
@@ -31,9 +31,9 @@ class payment_order_create(orm.TransientModel):
     _inherit = 'payment.order.create'
 
     def extend_payment_order_domain(
-            self, cr, uid, payment_order, domain, context=None):
+            self, cr, uid, ids, payment_order, domain, context=None):
         super(payment_order_create, self).extend_payment_order_domain(
-            cr, uid, payment_order, domain, context=context)
+            cr, uid, ids, payment_order, domain, context=context)
```
pedrobaeza, line 36: The same

account_payment_blocking/model/payment_order_create.py
adrienpeiffer, line 30: Hi @pedrobaeza,
I think it's a bit out of scope. It's just a fix.

pedrobaeza, line 30: As this change is only 2 lines, I see it very simple:

```
@api.multi
def extend_payment_order_domain(
        self, payment_order, domain):
    super(payment_order_create, self).extend_payment_order_domain(
        payment_order, domain)
...
```

adrienpeiffer, line 30: Yes but there something else in this module

pedrobaeza, line 30: You can migrate only these files, not the entire module. Old and new API can coexist, even in the same module or file

sbidoul:
New API migration is never trivial. We always have surprises. And we have no resources to do additional new API convertion in this module right now.

:+1:

lmignon:
:+1:

pedrobaeza:
Not for this case. I ask because it is very simple (I already paste the code), but no problem in letting as is.

:+1: Merging.

PR #211, opened 2015-09-01
[FIX] don't change api signature

hbrunn:
changing function signatures for no reason is a bad idea: https://github.com/odoo/odoo/blob/8.0/addons/account_payment/account_move_line.py#L68


sbidoul:
The upstream signature is confusing to say the least. But indeed, changing it in overrides can only create troubles.

:+1:

pedrobaeza:
:+1:

adrienpeiffer:
:+1:

PR #215, opened 2015-09-24
[8.0][IMP] Adapt the filter selecting invoices to pay

ecino:
**Filter as well invoices that are present in a confirmed payment_order** : in our use case, we don't immediately reconcile invoices after having them collected in a payment_order. I think it is not wrong to avoid the ability to collect them in another order before they are reconciled.

### DIFF
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -69,12 +69,9 @@ class PaymentOrderCreate(models.TransientModel):
             in the payment order.
 
         This implementation filters out move lines that are already
-        included in draft or open payment orders. This prevents the
+        included in non-cancelled payment orders. This prevents the
         user to include the same line in two different open payment
-        orders. When the payment order is sent, it is assumed that
-        the move will be reconciled soon (or immediately with
-        account_banking_payment_transfer), so it will not be
-        proposed anymore for payment.
+        orders.
 
         See also https://github.com/OCA/bank-payment/issues/93.
 
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -83,7 +83,7 @@ class PaymentOrderCreate(models.TransientModel):
         """
         self.ensure_one()
         payment_lines = self.env['payment.line'].\
-            search([('order_id.state', 'in', ('draft', 'open')),
+            search([('order_id.state', '!=', 'cancel'),
                     ('move_line_id', 'in', lines.ids)])
         to_exclude = set([l.move_line_id.id for l in payment_lines])
         return [l.id for l in lines if l.id not in to_exclude]


pedrobaeza:
Good improvements, thanks!

:+1:

account_banking_payment_export/models/account_payment.py
```diff
@@ -35,6 +35,12 @@ class PaymentOrder(models.Model):
         readonly=True, states={'draft': [('readonly', False)]})
     mode_type = fields.Many2one('payment.mode.type', related='mode.type',
                                 string='Payment Type')
+    total = fields.Float(compute='_total', store=True)
+
+    @api.depends('line_ids', 'line_ids.amount')
+    @api.one
+    def _total(self):
```
pedrobaeza: Minor thing: please call it `_compute_total` to follow conventions about compute methods names

account_banking_payment_export/wizard/payment_order_create.py
```diff
@@ -83,7 +80,7 @@ def filter_lines(self, lines):
         """
         self.ensure_one()
         payment_lines = self.env['payment.line'].\
-            search([('order_id.state', 'in', ('draft', 'open')),
+            search([('order_id.state', '!=', 'cancel'),
                     ('move_line_id', 'in', lines.ids)])
```
sbidoul, line 84: I'm not sure. What if, for instance, a payment order is sent/done, but for some reason (partially) rejected by the bank. The user will unreconcile some move lines and expect to be able to add the move line to another payment order. For reference, this domain has been defined after discussion in #110, #93 and related issues and PR. So I'd say :-1: but requires further thinking.

account_banking_payment_export/models/account_payment.py
ecino: I called it this way in order to be sure it overrides the method used for the functional field defined in the core module `account_payment`. Do you think we can define another instead of overriding this one ?

account_banking_payment_export/wizard/payment_order_create.py
pedrobaeza, line 84: If you have used the transfer account, it shouldn't be a problem, but in other cases... both filters can work, but with further modules. I'm working now on the abandoned _account_payment_return_ that allows to reflect returns in your payment orders. This module reconciles the original invoice move line, with the payment order move line and a new generated move line for the return. This way, you get the same balance, an history is created, and you can pass the new move line with another due date to a new payment order. In this flow, the new filter is the good one.

Anyway, @ecino, can you please propose both improvements in separate split so that we can get as soon as possible the computed total (which is something that I need urgently), and discuss the other also with the context of the other module?

account_banking_payment_export/models/account_payment.py
pedrobaeza: As you are redefining the method above, there's no problem about that.

account_banking_payment_export/wizard/payment_order_create.py
ecino, line 84: Then we could need one wizard developed in switzerland localization (PR open for v8).
https://github.com/OCA/l10n-switzerland/pull/162/files#diff-f006c2d78fe70d2df5e317db5b05160cR1

We could move this wizard inside this module. It is used to select some invoices and free them of a payment order (= remove them from the payment order and put them in a cancelled one).

This way, we better capture the workflow by showing the invoice was collected in a payment order but then it had a failure. If we simply allow collecting again the invoice, we cannot see that the first payment order was unsucessful.

So if you are OK with this, I can remove that functionality from the PR in swiss repository and include it in this PR.

pedrobaeza, line 84: Can you make a separate PR for the totals first?

About the other proposal, what do you think about the returns flow I have talked about?

ecino:
Store functionality moved in #216

account_banking_payment_export/wizard/payment_order_create.py
ecino, line 84: The return would not work for us, as we import the returns from a bank statement. We would thus don't need to generate other move lines.
As @sbidoul noticed, I think we should try to get this working without the need of any other module. This is why I proposed to add the ability of removing an invoice from a payment order (=> moving it to a cancelled payment order).

pedrobaeza, line 84: OK, for me let's go then for that approach.

sbidoul, line 84: Moving payment lines out of a sent payment order does not look right to me, as then the payment order would not reflect what has been sent to the bank.

@ecino I think there is a state on payment.order.line too, perhaps it can be used in the filter? 

@alexis-via you may be interested in this conversation.

sbidoul:
@ecino I took the liberty to adapt the PR title.

account_banking_payment_export/wizard/payment_order_create.py
ecino, line 84: The state on payment.lines is not the same as the payment order. It is called "Communication Type" and can have values `normal` or `structured`. I don't think we can use this field. But maybe add another field like `transaction_state` to capture whether the transaction is `pending`, `successful` or has `failed` and use that for the filter ? But then we should have a mechanism to properly update the state and that can be complex.

alexis-via:
It seems to me that, with ('order_id.state', '!=', 'cancel'), you don't handle the scenario where you do a SEPA Direct Debit for 5 customers, 4 direct debits succeed and 1 is rejected. This first direct debit order is now in 'done' I guess. Then you create a new SEPA direct debit to re-try to debit the one that was rejected.... but you can't select it ! So I would say : ('order_id.state', 'not in', ('cancel', 'done'))

But it's just a quick analysis, I would need to look deeper into this.

pedrobaeza:
I have faced again this problem, and indeed what @alexis-via is true: I'm not able to select again the returned invoice. We need to fix this, or maybe your "revolution" PR covers this?

P.S.: Please, I need you to rebase the work to finish that PR urgently.

ecino:
Rebased the pull request.

ecino:
I think this fix is not correct, after having it tested further. I close the pull request. However I found that the problem may come from field function `amount_to_receive` in account_move_line. I think its definition is not perfect. I also noticed that searching through this field is really really slow, so this should be improved.

pedrobaeza:
Exactly, I came to the same conclusion, and the last one I came again to this PR I was thinking that the change was in that field. I finally patched for now this line: https://github.com/OCA/bank-payment/blob/9.0/account_direct_debit/models/account_move_line.py#L49, putting `po.state not in ('cancel', 'sent')`

PR #217, opened 2015-09-30
[FIX] account_banking_payment_export: Correct communication type for customer invoices in payment lines

pedrobaeza:
In current version of _account_banking_payment_export_, all customer invoices are imported in payment orders with communication type = structured, and the invoice number is used for the communication, but this kind of communication follows ISO, which requires to have an specific format, and invoice number may not have the required format, so this default value is very dangerous.

In this PR, what I do is to show the selection to decide if you want to use the structured communication at invoice level, and take this into consideration importing customer invoice lines into the payment order.

I have also bumped the version number.

### DIFF
--- a/account_banking_payment_export/__openerp__.py
+++ b/account_banking_payment_export/__openerp__.py
@@ -7,7 +7,7 @@
 
 {
     'name': 'Account Banking - Payments Export Infrastructure',
-    'version': '8.0.0.2.0',
+    'version': '8.0.0.3.0',
     'license': 'AGPL-3',
     'author': "ACSONE SA/NV, "
               "Therp BV, "
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -186,20 +186,19 @@ class PaymentOrderCreate(models.TransientModel):
         state = 'normal'
         communication = line.ref or '-'
         if line.invoice:
-            if line.invoice.type in ('in_invoice', 'in_refund'):
-                if line.invoice.reference_type == 'structured':
-                    state = 'structured'
-                    communication = line.invoice.reference
-                else:
-                    if line.invoice.reference:
-                        communication = line.invoice.reference
-                    elif line.invoice.supplier_invoice_number:
-                        communication = line.invoice.supplier_invoice_number
-            else:
-                # Make sure that the communication includes the
-                # customer invoice number (in the case of debit order)
-                communication = line.invoice.number.replace('/', '')
+            if line.invoice.reference_type == 'structured':
                 state = 'structured'
+                # Fallback to invoice number to keep previous behaviour
+                communication = line.invoice.reference or line.invoice.number
+            else:
+                if line.invoice.type in ('in_invoice', 'in_refund'):
+                    communication = (
+                        line.invoice.reference or
+                        line.invoice.supplier_invoice_number or line.ref)
+                else:
+                    # Make sure that the communication includes the
+                    # customer invoice number (in the case of debit order)
+                    communication = line.invoice.number
         # support debit orders when enabled
         if line.debit > 0:
             amount_currency = line.amount_residual_currency * -1
--- a/account_banking_payment_export/models/account_invoice.py
+++ b/account_banking_payment_export/models/account_invoice.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openerp import api, models, _
+from lxml import etree
 
 
 class AccountInvoice(models.Model):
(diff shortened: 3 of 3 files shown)


pedrobaeza:
@sbidoul, @alexis-via, @lmignon please review.

alexis-via:
I don't understand why you modify this. For a customer invoice, if you have "structured" as communication type, I would prefer to use the invoice number instead of the "reference" field. For a customer invoice, it's "our" invoice number that we should show to the customer in the structured communication field of a SEPA file I think. :-1:

pedrobaeza:
@alexis-via, please read the text below. If you put structured communication, the communication field must respect the ISO standard way of creating this reference. This is: RF + 2 control digits + reference. Invoice number obviously is not compliant. Some banks are ignoring the check for ISO compliant and simply treat it like an unstructured communication, but I have found some others that don't, and this is why I'm making this change.

account_banking_payment_export/wizard/payment_order_create.py
```diff
@@ -155,22 +155,20 @@ def _prepare_payment_line(self, payment, line):
...
-                if line.invoice.reference_type == 'structured':
-                    state = 'structured'
-                    communication = line.invoice.reference
-                else:
-                    if line.invoice.reference:
-                        communication = line.invoice.reference
-                    elif line.invoice.supplier_invoice_number:
-                        communication = line.invoice.supplier_invoice_number
-            else:
-                # Make sure that the communication includes the
```
sbidoul, line 199: @pedrobaeza I suggest handling this with a `if line.invoice.reference_type == 'structured':` in this branch.
The patch will be smaller and it will probably answer to @alexis-via's comment.

alexis-via:
@pedrobaeza OK, so I guess you have another module that write in the "reference" field the special RF + checksum + <I don't know what>. Where is it ?

account_banking_payment_export/__openerp__.py
```diff
@@ -41,6 +41,7 @@
         'views/account_payment.xml',
         'views/payment_mode.xml',
         'views/payment_mode_type.xml',
+        'views/account_invoice.xml',
```
sbidoul: Why do you change the invoice view? It should not be necessary.

pedrobaeza: I change it to allow to select the communication type in customer invoices (now it's a hidden field)

account_banking_payment_export/wizard/payment_order_create.py
pedrobaeza, line 199: I have reorganized the code to compact it, because making 2 times the comparison with reference_type is not needed and the code is bigger, but I can do it the other way if you prefer.

pedrobaeza:
@alexis-via, I haven't done the automatic generation of the structured reference yet, because it's not a need using the other communication type. Your banks for sure admits unstructured communication, because it's part of the standard, so compatibility with current instances is assured, but if you need me to do it for approving the PR, I can do it.

account_banking_payment_export/__openerp__.py
sbidoul: Hm. That will probably break things if using l10n_be_invoice_bba which also does a similar thing.

https://github.com/odoo/odoo/blob/8.0/addons/l10n_be_invoice_bba/account_invoice_view.xml#L12

Not sure how to solve this...

angelmoya:
It would be possible to only change the communitacion type for customer invoices?

This way it's not necesary to change the invoice view, and don't break things if using l10n_be_invoice_bba

pedrobaeza:
Another solution I have thought is to inject the field dynamically in fields_view_get if it doesn't exist.

angelmoya:
@pedrobaeza it looks right

intero-chz:
What is the current state of progression for this pull request?

pedrobaeza:
I have added a routine in fields_view_get to dinamically check if the field is already present, and if not, add the field, so that now there's no conflict with _l10n_be_invoice_bba_.

@antespi, @rafaelbn, @sbidoul, @amoya-dx, please review.

rafaelbn:
Hi, thanks. For me is :+1:

pedrobaeza:
Can somebody review and approve this? It's very important for spanish transactions to use the correct field for communication.

carlosdauden:
:+1:

pedrobaeza:
@alexis-via, @sbidoul please review this one and give the final approval after the "big merge".

account_banking_payment_export/views/account_invoice.xml
```diff
@@ -0,0 +1,17 @@
...
+            <field name="model">account.invoice</field>
+            <field name="arch" type="xml">
+                <field name="origin" position="after">
+                    <field name="reference_type"/>
+                </field>
+            </field>
+        </record>
+
+    </data>
+</openerp>
```
sbidoul: Is not this one redundant with the field_view_get?

pedrobaeza: Sorry, I forgot to remove it with the rebase

pedrobaeza:
@sbidoul, XML view removed.

account_banking_payment_export/wizard/payment_order_create.py
```diff
@@ -186,20 +186,19 @@ def _prepare_payment_line(self, payment, line):
...
+                communication = line.invoice.reference or line.invoice.number
+            else:
+                if line.invoice.type in ('in_invoice', 'in_refund'):
+                    communication = (
+                        line.invoice.reference or
+                        line.invoice.supplier_invoice_number or line.ref)
+                else:
+                    # Make sure that the communication includes the
+                    # customer invoice number (in the case of debit order)
+                    communication = line.invoice.number
```
sbidoul, line 201: in previous version there was a `replace('/', '')`, is there a specific reason to remove it?

pedrobaeza, line 201: Yeah, because there was no need to replace that character, because it's supported on the communication text.

sbidoul, line 201: note I don't know why there was that replace before... @alexis-via any idea?

pedrobaeza, line 201: I think this comes from the old times before XML files. I personally see this very annoying to have `FV/2016/1234`, and find in the statements `FV20161234`. Harder to read and no automatic matching.

sbidoul, line 201: Agreed.

sbidoul:
:+1: 

I propose to merge after the week-end unless @alexis-via has additional remarks.

pedrobaeza:
As the time has passed, I merge this one.

PR #216, opened 2015-09-24
[8.0][IMP] Store total field on payment orders

ecino:
Store total fields in payment order to improve responsiveness of the views.

PR #218, opened 2015-09-30
[ADD] bump version, migrate total column

hbrunn:
#216 will cause trouble when your workflow on payment orders contains functions defined in not yet loaded modules - `account_banking_payment_transfer` is a candidate for that. Then computing the field will fail because the workflow conditions can't be evaluated.

This migration writes the wrong value in multicurrency situations, but works fine otherwise

### DIFF
--- a/account_banking_payment_export/__openerp__.py
+++ b/account_banking_payment_export/__openerp__.py
@@ -24,7 +24,7 @@
 
 {
     'name': 'Account Banking - Payments Export Infrastructure',
-    'version': '0.1.165',
+    'version': '8.0.0.1.166',
     'license': 'AGPL-3',
     'author': "ACSONE SA/NV, "
               "Therp BV, "
--- a/account_banking_payment_export/migrations/8.0.0.1.166/pre-migrate.py
+++ b/account_banking_payment_export/migrations/8.0.0.1.166/pre-migrate.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    This module copyright (C) 2015 Therp BV (<http://therp.nl>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+def migrate(cr, version):
+    cr.execute('alter table payment_order add column total numeric')
+    cr.execute(
+        'update payment_order '
+        'set total=totals.total '
+        'from '
+        '(select order_id, sum(amount_currency) total '
+        'from payment_line group by order_id) totals '
+        'where payment_order.id=totals.order_id')


pedrobaeza:
Which function are you finding problematic? I have already loaded this update in databases with _account_banking_payment_transfer_ and it's not the case.

hbrunn:
@pedrobaeza  then you were lucky that no payment order was in a state that has an outgoing workflow transition with a condition of a not yet loaded module

pedrobaeza:
Well, it was that, that all orders were in done state!

OK, then :+1:

hbrunn:
thinking of which, won't this column be dropped when updating account_payment? Then we get the same problem again, and migrations scripts won't help. The solution would be to rename the stored field and overwrite the total field with a related field

hbrunn:
@pedrobaeza I just checked this, and yes, the column is dropped and you'll always get this problems on --update=all

I'd propose to revert #216 and replace it by a fixed version. Meanwhile I close this PR because it's not fixing the problem

sbidoul:
@hbrunn Does this mean it's not possible to override a computed field to make it a stored field?

hbrunn:
@sbidoul when workflows with conditions are involved, it's not really possible to get this working properly. That's why OpenUpgrade disables workflows during the upgrade, we had a plethora of similar problems before.
It _should_ be simple to patch OCB to disable workflow transitions in recomputation phase (and possibly do a `trg_write` on the touched models after all modules are loaded), but I'm not very optimistic this is going to go into upstream, and modules working in OCB and not working in upstream is something we should avoid at nearly any cost in my opinion

hbrunn:
another solution would be do create and recompute the column by hand in the class' `init` (a function called by xml data or `_register_hook` is too late)

pedrobaeza:
I see the problem... Do you think there's a possible patch on Odoo that prevents the dropping of the column if there are another modules that modifies the field?

hbrunn:
I'm just reading some code, it seems to be possible to hack this way simpler: Override `step_workflow` (https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L3549), do $magic to detect if we're busy with an update ($magic could be a context key, backtrace, some other heuristics), and suppress evaluation if $magic is true. Maybe there's also some trickery possible with `recompute` and/or `_recompute`.

I'm just busy fixing stuff up for vacations, so I can't provide the code myself, but this seems to be a fair chance for a clean solution.

hbrunn:
@pedrobaeza this might be an option: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L2511 But I can't think of a general possibility currently to set this flag from this module early enough to have it set when account_payment's `_auto_init` runs. My comment before seems more promising to me

pedrobaeza:
I'm seeing a nodrop attribute. I think this attribute is intended for avoiding the removal of the field. Doesn't it?

hbrunn:
yes, but how can we set it in time? (just changing this in account_payment seems no option to me because then OCB is incompatible with odoo)

pedrobaeza:
I'm making a test setting it in account_banking_payment_export to see if it's persistent between updates...

pedrobaeza:
Yeah, it's not working as expected, because Odoo doesn't compose these attributes in the model until the module is loaded (I had wished to have some kind of prefetching for this...). Let's keep trying.

hbrunn:
if you're in trying mood, I'd try https://github.com/OCA/bank-payment/pull/218#issuecomment-144477082

hbrunn:
@pedrobaeza  my approach would be: Override `step_workflow` (https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L3549), check if `len(context) == 1 and not context.get('active_test')` (cf. https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L2697), only then do the super expensive traceback to be sure we're called by `_auto_init` and do nothing then

hbrunn:
@pedrobaeza or simpler: create an environment in `_auto_init` with `uid=SUPERUSER_ID` and `context={'active_test': False}`. Put some marker attribute in this environment. If things work as I hope, this environment will be chosen below (cf. https://github.com/OCA/OCB/blob/8.0/openerp/api.py#L717). When in `step_workflow`, just check for this marker, remove the marker after calling super in `_auto_init`, voila. This costs nearly no overhead and should be safe.

pedrobaeza:
OK, now I have to go outside. I'll try tonight. Thanks for the hints.

hbrunn:
outside is where this hot yellow thing is right?
;-)
last idea: Put a marker on cr in `_auto_init`, call super, check for this marker in `step_workflow`, remove the marker after super returns

hbrunn:
really last idea: do an unconditional patch of the field in `openerp.addons.account_payment.account_payment.payment_order._columns['tota']` where you set the `nodrop` flag on the field. I can't think of negative side effects for databases without the module installed at the moment, but it's late here... Protect against ImportError, then we should be fine

pedrobaeza:
Here it is the final solution with the last suggestion: https://github.com/OCA/bank-payment/pull/219

PR #214, opened 2015-09-21
[8.0][revolution] Add support for grouping the payments, for deducing a refund from an invoice and remove storno

alexis-via:
Here are some explanation about this important PR. This PR adds support for grouping the payments, which is a feature that I always needed and wanted to add in OpenERP, and this PR makes it possible at last !

Example1: I want to pay 2 invoices of the same supplier via SEPA credit transfer on 2015-09-30. With the current code, it will generate one SEPA Credit Transfer file with 2 payments inside. So, for the bank, it will be considered as 2 different payments ; if the bank charges per payment (most banks do that, at least in France), the cost is x2 !

Example2 (which is the scenario that motivate this PR): I have an agreement with my customers to invoice for each delivery and do one "direct debit" per month for all the invoices of the month. So I need to generate a direct debit file that has 1 payment corresponding to N invoices.

I have thought about many different ways of implementing this:

a) do the "grouping" inside the code that generate the bank file.
Advantage: no changes in "account_banking_payment_transfer", no change in the datamodel.
Disadvantage: when the user looks at the payment.line on the payment.order, it doesn't reflect the content of the bank file. The user could read the bank file itself to check the grouping, but the bank files are not easy to read !

b) have grouped "payment.line" (1 payment.line for N invoices)
Advantage : no changes in the code that generate the bank files
Disadvantage : I can't do a link from 1 payment.line to N account.move.line because the M2O field"move_line_id" goes from payment.line to account.move.line (I would need it the other way around !). That would also break account_banking_payment_transfer.

c) have 2 O2M fields on payment.order:
- 1 O2M field pointing to payment.line (current object)
- 1 O2M field pointing to bank.payment.line (new object). One bank.payment.line would group several payment.lines (and we would have a O2M field from bank.payment.line to payment.line, to keep track of the grouping)
  The bank file would be generated from the object bank.payment.line.
  Avantage : it's powerful, we can control the grouping and the user can check the content of the bank file by looking at the bank payment lines. No changes required in the module account_banking_payment_transfer.
  Disadvantages : it's an extension of the current datamodel, with the introduction of a new object.

So this PR implements option "c" !

The very good point of this PR is that there are almost no changes in the SEPA modules that generate the bank file ; the only change is that instead of looping on the payment lines, we loop on the bank payment lines.

This PR changes the dependency of the module account_banking_mandate to add a dependency on account_banking_payment_export. I don't know if it's a problem or not (I consider that "account_banking_payment_export" is the "base" module of this OCA project).

PR #220, opened 2015-10-08
[FIX] account_banking_payment_export: Use residual amount in direct debits

pedrobaeza:
When making direct debit orders, the amount taken for the order is the total of the move line, not the residual one, leading to possible invoices over-debited. This is due to the rely on _amount_to_receive_ field, which doesn't take into account the residual value, but the full one (https://github.com/OCA/bank-payment/blob/8.0/account_direct_debit/models/account_move_line.py#L39). 

This patch solves this in a fast way by using the same amount_residual_currency for both lines. Later, I plan to fix the `amount_to_receive` compute method so that we can rely on it, but this is a bigger patch that maybe will be done for v9.

### DIFF
--- a/account_banking_payment_export/wizard/payment_order_create.py
+++ b/account_banking_payment_export/wizard/payment_order_create.py
@@ -171,12 +171,7 @@ class PaymentOrderCreate(models.TransientModel):
                 # customer invoice number (in the case of debit order)
                 communication = line.invoice.number.replace('/', '')
                 state = 'structured'
-        # support debit orders when enabled
-        if (payment.payment_order_type == 'debit' and
-                'amount_to_receive' in line):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@alexis-via, you make also the same patch on you PR (https://github.com/OCA/bank-payment/pull/214/files#diff-a0f063789c0a3dbd84d4bbfb76d812feR175), so please approve this one as a fast-patch until your enhancement arrives.

alexis-via:
:+1:

carlosdauden:
:+1:

rafaelbn:
:+1:

pedrobaeza:
I self-merge it then as it's something critical

PR #226, opened 2015-10-31
[ADD] account_payment_transfer_reconcile_batch

pedrobaeza:
# 
# Batch reconciliation for transfer lines created in payment orders

This module allows to process with the connector technology the heavy task of
reconciliation of the receivable/payable journal entries of a payment order
against the created entries in transfer accounts.

This approach provides many advantages, similar to the ones we get
using that connector for e-commerce:
- Asynchronous: the operation is done in background, and users can
  continue to work.
- Dedicated workers: the queued jobs are performed by specific workers
  (processes). This is good for a long task, since the main workers are
  busy handling HTTP requests and can be killed if operations take
  too long, for example.
- Multiple transactions: this is an operation that doesn't need to be
  atomic, and if a line out of 100,000 fails, it is possible to catch
  it, see the error message, and fix the situation. Meanwhile, all
  other jobs can proceed.

Inspired on _account_move_batch_validate_ module from Camptocamp and ACSONE.
# Installation

This module requires the connector module, hosted on
`OCA/connector <https://github.com/OCA/connector>`_
# Configuration

This will only work for payment modes that have a transfer account set.
# Usage

When exporting the payment order, click on _Validate_ to generate the transfer
move. One connector job will be created for each payment line for a deferred
conciliation of this line.

### DIFF
--- a/account_payment_transfer_reconcile_batch/models/payment_order.py
+++ b/account_payment_transfer_reconcile_batch/models/payment_order.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# © 2015-2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 import logging
@@ -25,7 +25,9 @@ class PaymentOrder(models.Model):
 
     @api.multi
     def _reconcile_payment_lines(self, bank_payment_lines):
-        if config['test_enable'] or self.env.context.get('no_connector'):
+        test_condition = (config['test_enable'] and
+                          not self.env.context.get('test_connector'))
+        if test_condition or self.env.context.get('no_connector'):
             return super(PaymentOrder, self)._reconcile_payment_lines(
                 bank_payment_lines)
         session = ConnectorSession.from_env(self.env)
--- a/account_payment_transfer_reconcile_batch/models/payment_order.py
+++ b/account_payment_transfer_reconcile_batch/models/payment_order.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# © 2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 import logging
--- a/account_payment_transfer_reconcile_batch/models/payment_order.py
+++ b/account_payment_transfer_reconcile_batch/models/payment_order.py
@@ -24,25 +24,21 @@ class PaymentOrder(models.Model):
     _inherit = 'payment.order'
 
     @api.multi
-    def _reconcile_payment_lines(self, payment_lines):
-        if config['test_enable']:
+    def _reconcile_payment_lines(self, bank_payment_lines):
+        if config['test_enable'] or self.env.context.get('no_connector'):
             return super(PaymentOrder, self)._reconcile_payment_lines(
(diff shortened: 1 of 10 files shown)


rvalyi:
:+1: LGTM code review.
Feature seems okay too but may be I'm not specialized enough to judge.

That would however require to add connector in the oca_dependencies.txt
I'm also not sure until where it's reasonable to have an OCA repo depends on other before deciding that's a valid reason to split the repo. May be here not many modules will also depend on connector and may be that's a dependency that's clean enough to accept the module in this repo. But that's a question to consider before we have complexity explosion.

pedrobaeza:
Yeah, this one requires connector. I have added the dependency on oca_dependencies.txt now. If you prefer to have this module in another repository, we can talk about it. In account-financial-tools there's already a module with similar purpose/dependency.

sbidoul:
@adrienpeiffer can you review this one? I think you have done something similar.

rvalyi:
@pedrobaeza is a bit like server-tools, for me we can consider it's just Odoo server on steroid and not an important functional dependency justifying a repo split. If in another repo based on dependencies, that other repo would have no critical mass either, so okay for me until if have the 1 module = 1 repo thing...

account_payment_transfer_reconcile_batch/models/payment_order.py
```diff
@@ -0,0 +1,48 @@
...
+class PaymentOrder(models.Model):
+    _inherit = 'payment.order'
+
+    @api.multi
+    def _reconcile_payment_lines(self, payment_lines):
+        if config['test_enable']:
+            return super(PaymentOrder, self)._reconcile_payment_lines(
+                payment_lines)
+        session = ConnectorSession(
+            self.env.cr, self.env.uid, context=self.env.context)
```
adrienpeiffer: `session = ConnectorSession.from_env(self.env)` ?

account_payment_transfer_reconcile_batch/models/payment_order.py
```diff
@@ -0,0 +1,48 @@
...
+            if line.move_line_id:
+                reconcile_one_move.delay(
+                    session, 'payment.line', line.id)
+            else:
+                self.action_sent_no_move_line_hook(line)
+
+
+@job(default_channel='root.account_payment_transfer_reconcile_batch')
+def reconcile_one_move(session, model_name, payment_line_id):
+    payment_line_pool = session.pool[model_name]
```
adrienpeiffer: You can use session.env and use new API in this method.

account_payment_transfer_reconcile_batch/models/payment_order.py
```diff
@@ -0,0 +1,48 @@
...
+    if payment_line_pool.exists(session.cr, session.uid, payment_line_id):
+        payment_line_pool.debit_reconcile(
```
adrienpeiffer: I think it would be better to call `_reconcile_payment_lines` with only one line. You could use the context to give a parameter to call directly the super method in `_reconcile_payment_lines`

pedrobaeza:
@adrienpeiffer, your suggestions have been taken into account. Thanks for them.

But now this depends on https://github.com/OCA/bank-payment/pull/250

pedrobaeza:
@adrienpeiffer, this one is now ready to be reviewed, as the dependency has been merged.

pedrobaeza:
cc @Tecnativa

rafaelbn:
:+1: Long time in production enviroments

antespi:
:+1:

rafaelbn:
Hi, @sbidoul @adrienpeiffer could you here last review for merging? Thanks!

carlosdauden:
:+1:

yajo:
This one seems to have enough :+1:

pedrobaeza:
OK, we have enough positive reviews and I have added tests, so I proceed to merge.

Issue #228, opened 2015-11-26
[7.0] account_banking_sepa_direct_debit Exception in migration when table does not exist.


NL66278:
An error occurs in module account_banking_sepa_direct_debit when running the migration, if the table sdd_mandate does not exist.

2015-11-26 16:27:58,293 1386 INFO 2015_11_16_test openerp.modules.migration: module account_banking_sepa_direct_debit: Running migration [
0.2>] account.banking.sepa.direct_debit:: Move sdd_mandate data to account_banking_mandate
2015-11-26 16:27:58,294 1386 ERROR 2015_11_16_test openerp.sql_db: Programming error: relation "sdd_mandate" does not exist
LINE 1: ...ation, recurrent_sequence_type, type, 'CORE' FROM sdd_mandat...
                                                             ^
, in query INSERT INTO account_banking_mandate (id, create_uid, create_date, write_date, write_uid, partner_bank_id, last_debit_date, scan
, company_id, state, unique_mandate_reference, signature_date, sepa_migrated, original_mandate_identification, recurrent_sequence_type, ty
pe, scheme) SELECT id, create_uid, create_date, write_date, write_uid, partner_bank_id, last_debit_date, scan, company_id, state, unique_m
andate_reference, signature_date, sepa_migrated, original_mandate_identification, recurrent_sequence_type, type, 'CORE' FROM sdd_mandate
2015-11-26 16:27:58,294 1386 ERROR 2015_11_16_test openerp: Failed to initialize database `2015_11_16_test`.
Traceback (most recent call last):
  File "/home/odoocutst70/odoo/parts/odoo/openerp/cli/server.py", line 97, in preload_registry
    db, registry = openerp.pooler.get_db_and_pool(dbname,update_module=update_module)
  File "/home/odoocutst70/odoo/parts/odoo/openerp/pooler.py", line 33, in get_db_and_pool
    registry = RegistryManager.get(db_name, force_demo, status, update_module)
  File "/home/odoocutst70/odoo/parts/odoo/openerp/modules/registry.py", line 203, in get
    update_module)
  File "/home/odoocutst70/odoo/parts/odoo/openerp/modules/registry.py", line 233, in new
    openerp.modules.load_modules(registry.db, force_demo, status, update_module)
  File "/home/odoocutst70/odoo/parts/odoo/openerp/modules/loading.py", line 359, in load_modules
    force, status, report, loaded_modules, update_module)
[...]

PR #229, opened 2015-11-26
[FIX] Issue #228 Error when table to be migrated does not exist.

NL66278:
See #228 Problem when table sdd_mandate does not exit during migration.

### DIFF
--- a/account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py
+++ b/account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py
@@ -34,7 +34,7 @@ def migrate(cr, installed_version):
     if not installed_version:
         return
 
-    def table_exists(cr, table):                                                      
+    def table_exists(cr, table):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Does this still applies?

NL66278:
I think it still does, anyway I think it is always good when migrating to test wether the stuff to be migrated is actually there.

pedrobaeza:
Can you check falke8 errors and I'll merge?:

```
account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py:37:33: W291 trailing whitespace
account_banking_sepa_direct_debit/migrations/0.2/post-consolidation.py:55:79: E502 the backslash is redundant between brackets
```

NL66278:
Flake8 corrections done.

pedrobaeza:
OK, there are now a lot of other flake8 errors, but they are due to the new check that is being made, so I merge this one.

PR #233, opened 2015-12-18
[FIX] account_direct_debit: finding account move lines without invoice

ozono:
Fix for finding account move lines without invoice. ~~Same~~ similar bug fixed in https://github.com/OCA/bank-payment/pull/139 by @pedrobaeza


pedrobaeza:
But the fix is already merged, and works the same for payment orders or direct debit orders. Why do you need this?

ozono:
I made a typo. Bug is similar but is not the same. It affects to account_direct_debit. #139 is related to account_payment_partner. account_direct_debit also introduces a ('invoice', '=', False) in domain that doesn't match due to https://github.com/odoo/odoo/issues/4857

pedrobaeza:
:+1:

Please sign the CLA

ozono:
Thanks. CLA signed and sent 3 days ago. :smiley:

pedrobaeza:
Can you please rebase and see if the problem is still there with the new big change that has been merged?

PR #234, opened 2015-12-18
[FIX] account_payment_mode_term new api compatibility

sbidoul:
Without this patch we have `TypeError: extend_payment_order_domain() takes exactly 3 arguments (2 given)` when adding an invoice to a payment order.


pedrobaeza:
:+1:

adrienpeiffer:
:+1:

PR #238, opened 2016-01-25
[FIX] account_payment_*: Ensure work with dropshipping dual invoice

pedrobaeza:
When invoicing from dropshipping, the same picking serves for both invoicing sale and purchase orders. The OCA module _stock_dropshipping_dual_invoice_ makes that this happens, but as the picking is the same, it contains both the sale and the purchase reference, making that the resulting invoices don't have
the correct payment mode. This is harmless for the rest of operations, as the dictionary query is provided with the proper default value. On v9, dropshipping dual invoice has been merged into core, so this patch for sure will be welcome in that version (if the used method is the same).

### DIFF
--- a/account_payment_purchase/models/stock_picking.py
+++ b/account_payment_purchase/models/stock_picking.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 #    Account Payment Purchase module for OpenERP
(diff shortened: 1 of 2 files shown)


pedrobaeza:
cc @rafaelbn @antespi @JavierIniesta @JMODaERP

feketemihai:
:+1: Good catch...

JMODaERP:
OK. Thank you, this fix was much needed. Tested and   :+1:

JavierIniesta:
Tested in local environment :+1:

pedrobaeza:
Runbot is not working, killing instances after being one hour processing, so as all is correct and have enough positive review, I merge.

PR #239, opened 2016-02-09
[NEW] SCT for Italy based on xml standard CBI

zeroincombenze:
Italian banks do not use ISO standard but a modified standard called CBI.
Added a new protocol suffixed with -CBI.
No change to flow neither how to Odoo works.


pedrobaeza:
As commented in IRC, this should be included in an specific module inside l10n-italy, so :-1:

zeroincombenze:
Hi Pedro,
thanks for your response but .. there is a problem.
Today it is impossible separate Italian SCT from standard SCT because bank-payment module must be modified.
So the question is: you prefer one of
1. I update heavily bank-payment module to separate the localization part of software
2. Acquire this PR that works both for Italy both fo ISO
3.Do nothing, Italian people will use another banking-payyment modules different and not compatible with OCA/Odoo (this is the default)
4.Some other solution I cannot imagine now

I accept everything you and OCA will decide and I am to give you and OCA everry information about 2 different standard.

TIA

pedrobaeza:
I prefer the first option, and I don't think if we structure correctly the module to make too much modifications. There are already lots of methods that can be overridden, and we can provide more to accommodate your needs. Then, you provide this l10n_it_\* module based on this one.

What do you think, @sbidoul and @alexis-via?

zeroincombenze:
Hi Pedro,
I am avaiable to make modification, if you will. But I do not see unit test, so before any modification I prefer create unit test to certificate software
I tested it on 7.0 but I no yet tested on 8.0 neither on 9.0

pedrobaeza:
Yeah, unfortunately there's no tests for this one. Are you planning to include them?

zeroincombenze:
Yes but am newbie in Odoo and I develop very slowly

pedrobaeza:
Cool, your contribution will be very appreciated. The steps I see for this test:
- Create one test customer, with an account bank and a mandate.
- Create one test payment mode.
- Create one test customer invoice.
- Validate it.
- Create a test payment order.
- Include the move line of the invoice.
- Generate the file, and diff compare with an already generated file that you include in the module.

zeroincombenze:
Ok Pedro It good.
I have just talked eLBati and he explained me how to override method for italian version.
So I am thinking to add just some test on OCA version and builld a new localization module for Italian SCT.

alexis-via:
There is a test for SEPA credit transfer in the module "account_banking_tests", cf file "tests/test_payment_roundtrip.py".

alexis-via:
It is in a separate module for dependency reasons, because the test uses the module "account_banking_payment_transfer" and not only the module "account_banking_sepa_credit_transfer".

StefanRijnhart:
It seems the case for a lot of these modifications that you can take the resulting XML structure in an l10n-it specific inherit of this module and make your modifications at that stage. What do you think of that approach? We should really avoid any spaghetti (pun not intended) in the form of 'if country == 'IT':'

zeroincombenze:
Yes @Stefan,
italian group explained me how to do. 
I am a newbie in OCA/Odoo development.
Thanks for your hint.

StefanRijnhart:
Thanks @zeroincombenze for the follow-up!

PR #240, opened 2016-02-10
[8.0][FIX][account_banking_mandate] Allow to select bank account when creating a mandate

antespi:
Remove domain `[('partner_id', '=', partner_id)]` in `partner_bank_id` field, in order to be able to select a bank account when creating. Then on_change selects partner_id related with that bank account.


pedrobaeza:
Are you preventing with this the correct selection from partners and from payment orders?

antespi:
@pedrobaeza, yes both cases checked (Direct Debit Orders not Payment Orders). I don't know why that domain was there.

pedrobaeza:
Perfect :+1:

JavierIniesta:
:+1:

sergio-teruel:
Tested in runbot and good. :+1:

Issue #242, opened 2016-02-14
Mandates multicompany error


DarioLodeiros:
The restriction rule Banking Mandate multi-company, gives an error when editing a partner bank account from a company A when that account has created a mandate from another company B. So it is impossible to create a mandate for each company without disabling the rule.
Possible solution: Uncheck the rule writte in Banking Mandate multi-company. But i'm not sure this is correct.

pedrobaeza:
The best solution is to remove the default value for field company_id to leave the mandate unlocked by default, as the standard makes with the res.partner.bank model.

DarioLodeiros:
But I think that the mandate must is linked with a company_id (because I should have a mandate for company!).
This solution is valid if it is introduced into the module and will not be overwritten , no?

pedrobaeza:
Thinking again, you're right, but for doing it right, you have to include XML data that overwrites the rule by XML-ID, so that you don't lose changes on each update.

pedrobaeza:
Thinking again, you're right. Can you please make a PR making that change?

DarioLodeiros:
https://github.com/OCA/bank-payment/pull/243

PR #243, opened 2016-02-14
 Mandates multicompany error #242

DarioLodeiros:
The restriction rule Banking Mandate multi-company, gives an error when editing a partner bank account from a company A when that account has created a mandate from another company B. So it is impossible to create a mandate for each company without disabling the rule.
Solution: Uncheck the rule writte in Banking Mandate multi-company. But i'm not sure this is correct.


pedrobaeza:
:+1:

sergio-teruel:
:+1:

PR #250, opened 2016-02-15
[8.0] banking: Add support for grouping the payments, for deducing a refund from an invoice and remove storno

pedrobaeza:
This PR continues the one started by @alexis-via to complete it in order to be merged.

What I have done:
- Rebase for allowing to merge.
- Short headers according new standard and search for non attributed contributions.
- Remove field `label` in `payment.mode`, which is redundant with note field.
- Bump version numbers.

I copy the excerpt about the reasoning behind this improvement.

```
Here are some explanation about this important PR. This PR adds support for grouping the payments, which is a feature that I always needed and wanted to add in OpenERP, and this PR makes it possible at last !

Example1: I want to pay 2 invoices of the same supplier via SEPA credit transfer on 2015-09-30. With the current code, it will generate one SEPA Credit Transfer file with 2 payments inside. So, for the bank, it will be considered as 2 different payments ; if the bank charges per payment (most banks do that, at least in France), the cost is x2 !

Example2 (which is the scenario that motivate this PR): I have an agreement with my customers to invoice for each delivery and do one "direct debit" per month for all the invoices of the month. So I need to generate a direct debit file that has 1 payment corresponding to N invoices.

I have thought about many different ways of implementing this:

a) do the "grouping" inside the code that generate the bank file.
Advantage: no changes in "account_banking_payment_transfer", no change in the datamodel.
Disadvantage: when the user looks at the payment.line on the payment.order, it doesn't reflect the content of the bank file. The user could read the bank file itself to check the grouping, but the bank files are not easy to read !

b) have grouped "payment.line" (1 payment.line for N invoices)
Advantage : no changes in the code that generate the bank files
Disadvantage : I can't do a link from 1 payment.line to N account.move.line because the M2O field"move_line_id" goes from payment.line to account.move.line (I would need it the other way around !). That would also break account_banking_payment_transfer.

c) have 2 O2M fields on payment.order:
- 1 O2M field pointing to payment.line (current object)
- 1 O2M field pointing to bank.payment.line (new object). One bank.payment.line would group several payment.lines (and we would have a O2M field from bank.payment.line to payment.line, to keep track of the grouping)
The bank file would be generated from the object bank.payment.line.
Avantage : it's powerful, we can control the grouping and the user can check the content of the bank file by looking at the bank payment lines. No changes required in the module account_banking_payment_transfer.
Disadvantages : it's an extension of the current datamodel, with the introduction of a new object.

So this PR implements option "c" !

The very good point of this PR is that there are almost no changes in the SEPA modules that generate the bank file ; the only change is that instead of looping on the payment lines, we loop on the bank payment lines.

This PR changes the dependency of the module account_banking_mandate to add a dependency on account_banking_payment_export. I don't know if it's a problem or not (I consider that "account_banking_payment_export" is the "base" module of this OCA project).
```


pedrobaeza:
The previous PR already has 2 approvals.

@StefanRijnhart, now this PR is clean from account_banking module removal if you want to check.

alexis-via:
:+1:
No problem to use the "note" field of payment.mode.

pedrobaeza:
@sbidoul @alexis-via I have restored the check on entry_posted in this PR.

account_banking_payment_export/wizard/payment_order_create.py
```diff
@@ -56,8 +75,7 @@ def extend_payment_order_domain(self, payment_order, domain):
             # Do not propose partially reconciled credit lines,
             # as they are deducted from a customer invoice, and
             # will not be refunded with a payment.
-            domain += [('credit', '>', 0),
-                       '|',
+            domain += ['|',
```
sbidoul: why is this changing?

pedrobaeza: Because now there's the possibility of joining together refunds and invoices for example. The key now is to select the journal/s.

sbidoul: Is that consistent with the fact that further in the code we select the residual amount? At the very least, the comment just above is not correct anymore.

sbidoul:
In the select invoices to pay wizard, is there a reason for making the journal filter mandatory?

image

IMO it's better to leave it optional and propose move lines from all journals by default.

pedrobaeza:
You're right. I'll change it.

sbidoul:
The new filters look useful. Perhaps it would be better to leave "Linked to an invoice or refund" to False by default, so the default behaviour is the same as before and less surprising for users of the previous version.

account_banking_payment_export/wizard/payment_order_create.py
sbidoul: I just made a test, and I really don't understand why credit > 0 is removed from the domain.

For instance, I have a supplier with a credit of 15 and a debit of 5, partially reconciled together.

The payment order wizard proposes both, which is weird in itself, and if you select both, you get two payment lines both with an amount of 10 (which is the residual).

pedrobaeza:
About the "Linked to an invoice or refund" field, it's set at payment mode level, but I'll put this field as default=False.

account_banking_payment_export/wizard/payment_order_create.py
pedrobaeza: Indeed this is incorrect according my tests also. I have re-added the condition. Thanks for checking it deeply.

pedrobaeza:
@sbidoul @alexis-via, I have applied the same logic to the move lines domain in direct debit orders, so I think now it's all.

Please review and let's merge this long one.

alexis-via:
@pedrobaeza No problem for changing the default value of "default_invoice" on payment.mode, I don't care.

I'll check the question of the domain very soon (I hope tonight) ; I wrote all this 6 months ago, so I don't remember that kind of detail any more. I am just surprised that such a thing needed to be changed because this code has been used by several of my customers that use both payments and direct debit (and one of them has a very large volume of payments and direct debits and is using the module in production since 1/10/2015).

sbidoul:
I wanted to do a last functional test, but runbot is not cooperating.

cc/ @gurneyalex

sbidoul:
So runbot is back. I could do my last tests and say :+1: at last :smile: Thanks for this effort!

StefanRijnhart:
Thank you @alexis-via and everyone else involved! :+1: code review. Merging considering the support on the previous version of this PR.

Issue #224, opened 2015-10-20
[8.0][account_banking_sepa_direct_debit] Not supported 2 identifiers for CORE and B2B


rafaelbn:
module: account_banking_sepa_direct_debit
version: 8.0

**Steps to reproduce**
- After talking with banks they inform us that Companies must have 2 identifiers for CORE and for B2B direct_debit

**Current behavior**
- We have already only one indentifier

**Expected behavior**
- We must have 2 identifiers and choose one for CORE and the other for  B2B

I would like to know opinion of last contributors for resolving this new improvement. Thanks

cc @pedrobaeza @alexis-via @gurneyalex @sbidoul @rvalyi @rvalyi

pedrobaeza:
Solution is to provide identifier at payment mode level.

pedrobaeza:
The possible problem is if mixing both types on the same payment order.

rafaelbn:
Bank says we cannot mix in 1 file 2 kinds of debit, if we have both CORE and B2B we must generate 2 separated files.

pedrobaeza:
OK then, the solution is what I propose. I'll make a PR, although this will generate debate in the rest of the countries for sure.

alexis-via:
@rafaelbn 
Have you found a document that says you must have one identifier for Core and another one for B2B ? I read many documents on Sepa direct debit to be able to write the OCA module, and I never read such thing.

pedrobaeza:
@alexis-via, is not only a problem with CORE/B2B. All spanish banks use different suffixes for identifying you as a company, and even different suffixes by financial operation type (for example, for confirming or for credit transfer).

Do you have any problem if I make a PR moving initiating party identifier to the payment mode?

alexis-via:
@pedrobaeza I'm not sure to understand your point about suffixes. Which field are you talking about exactly ?
If it is something specific to Spain, you should create a small module in the Spanish localisation. If it's something for all SEPA countries, let's put it in the generic module.
I have several customers using SEPA direct debit in production and very happy with it. I'm not against changing the datamodel, but only if it's generic for all countries.

pedrobaeza:
The suffix is part of the initiating party identifier. For example, one of the possible form for this initiating party identifier is: `<country_code> + <vat-number> + '-' + <suffix>`.

[...]

PR #251, opened 2016-02-23
[8.0] New module to allow set bank's sepa identifier suffix in payment modes

omar7r:
Hi,

There are comments about this functionality in the issue https://github.com/OCA/bank-payment/issues/224

The initiating party identifier is configured by default in company as `<vat-number>` + `<suffix>` and the sepa creditor identifier too, as `<country_code>` + `<control_digit>` + `<suffix>` + `<vat-number>` this suffix can be variable depending on bank, then a company with several different banks, now is obligated to change this suffix for each payment order, with this module, you have the possibility to set in payment mode the bank's suffix and it will replace the default in sepa files, if you don't set this suffix the behavior is the default.

Regards.


account_banking_sepa_bank_suffix/README.rst
```diff
@@ -0,0 +1,54 @@
+Account Banking Sepa - Sepa identifier suffixes per payment mode
```
pedrobaeza, line 6: Use latest OCA's template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

account_banking_sepa_bank_suffix/README.rst
```diff
@@ -0,0 +1,54 @@
...
+================================================================
+
+* This module allows to change sepa identifiers suffixes from payment modes.
+
+Installation
```
pedrobaeza: This section is not needed as the module is in the same repository

account_banking_sepa_bank_suffix/README.rst
```diff
@@ -0,0 +1,54 @@
...
+============
+
+This module depends on : - account_banking_pain_base
+
+This module is part of the OCA/bank-payment suite.
+
+Configuration
```
pedrobaeza: Remove empty sections

account_banking_sepa_bank_suffix/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza, line 1: Use short headers: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

account_banking_sepa_bank_suffix/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Account Banking Sepa - Sepa identifier suffixes per payment mode",
+    "version": "8.0.0.0.0",
```
pedrobaeza: 8.0.1.0.0

account_banking_sepa_bank_suffix/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+    "author": "Comunitea,Odoo Community Association (OCA)",
+    "website": "http://www.comunitea.com",
+    "category": "Banking addons",
+    "description": """
```
pedrobaeza: Remove this key

account_banking_sepa_bank_suffix/i18n/account_banking_sepa_bank_suffix.pot
```diff
@@ -0,0 +1,38 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove POT file

account_banking_sepa_bank_suffix/models/payment_mode.py
```diff
@@ -0,0 +1,41 @@
...
+##############################################################################
+
+from openerp import models, fields, api, exceptions, _
+
+
+class PaymentMode(models.Model):
+
+    _inherit = "payment.mode"
+
+    suffix = fields.Char("Suffix", size=3, help='Suffix for sepa identifiers '
```
pedrobaeza: You can save lines (or at least make it more readable), using this layout:

```
    suffix = fields.Char(
        string="Suffix", size=3, 
        help='Suffix for sepa identifiers with this payment mode. If not set '
             'it will use the company configuration.')
```

account_banking_sepa_bank_suffix/models/payment_mode.py
```diff
@@ -0,0 +1,41 @@
...
+                                                'not set it will use the '
+                                                'company configuration.')
+
+    @api.one
+    @api.constrains('suffix')
+    def _check_suffix_format(self):
+        if self.suffix:
+            if len(self.suffix) != 3 or not self.suffix.isdigit():
+                raise exceptions.\
+                    Warning(_("Suffix must be compound by 3 numbers"))
```
pedrobaeza: s/numbers/digits

pedrobaeza:
For making this more general, you should put the complete SEPA identifier, instead of only the suffix. This way, you can replace by whatever you need.

omar7r:
Hi @pedrobaeza 

Changes done. About your last message, I don't finish to see clear this, because the company has three fields where affects the suffix, initiating_party_identifier, sepa_creditor_identifier and original_creditor_identifier, then we will need the three fields on payment model configuration and I think that the suffix is the same in the three cases.

pedrobaeza:
No, that's not really true. The three fields in company are:
- SEPA company identifier. This one is unique per company (VAT), and it's mandatory for all SEPA operations.
- SEPA creditor identifier. This one is supposed to be used for direct debit operations.
- SEPA transaction issuer: This one is supposed to be used for credit transfers.

What you are replacing here is the last two, that are put in the SEPA file depending on the type of operation.

omar7r:
Hi,

The original_creditor_identifier is added by account_banking_sepa_direct_debit equal that sepa_creditor_identifier (SEPA creditor identifier) but the first is used when a mandate is sent indicating the first time. I see three with the initiating_party_identifier (SEPA transaction issuer) that is added by account_banking_pain_base. I don't name the initiating_party_issuer (SEPA company identifier).

sandrafig:
Hi, 
First of all thank you for this PR, I want to share my opinion on this issue because I agree with Pedro that you should consider the complete SEPA identifier instead of only the suffix.
Also, I think we should be able to put **the three fields** (Initiating Party Identifier, Transaction issuer and SEPA creditor identifier) in the payment mode, because in my case, I work with two different banks and the situation I confront is the following:
- In **credit transfers** the suffix is the same for the two banks 
- In **direct debit** not only the suffix change, but also the first two fields (Initiating Party Identifier and Transaction issuer) are not the same that the ones used in the credit transfers.

So I (and anyone who works with more than one bank), have to constantly check and change the codes for every single transaction, depending if it is a direct debit, a credit transfer, and also it depends on what bank you are using.

I don't know if this situation happens in every country, but at least in Spain it does happen.

antespi:
@omar7r, @sandrafig, @pedrobaeza This is another technical approach to the same problem: https://github.com/OCA/bank-payment/pull/256

In this case, you can define a completely different identifiers per payment mode, but if not defined then company ones are selected. And no extra addon is needed.

Thanks for your review and opinion

pedrobaeza:
I close this one then, because the other is more generalist.

PR #252, opened 2016-02-24
[ADD]account banking not required

gonxi:
This module add a check in payment mode for do bank account optional.
We talk about this in the Code sprint from Spain

### DIFF
--- a/account_banking_not_required/README.rst
+++ b/account_banking_not_required/README.rst
@@ -1,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Thanks for this, but I think it should be a part of the main module: account_banking_paymente_export. There's not too much sense to need to install another module that makes something optional. What do you think?

Please sign the CLA.

gonxi:
i have thought that all people don't want have bank account optional, only some, but is not problem for me do the changes in: account_banking_payment_export and do the
 PR in this module.
If you view better i can do in this module.

pedrobaeza:
For those who don't want it, just keep the corresponding mark checked. Put the default to be mandatory and there won't be any difference with current situation.

Regards.

gonxi:
OK! close this PR and  i will do this changes in account_banKing_payment_export

PR #253, opened 2016-02-24
Account banking not required

gonxi:
Add a check in payment mode for do the account bank optional, for example... manual mode.

### DIFF
--- a/account_banking_payment_export/models/payment_mode.py
+++ b/account_banking_payment_export/models/payment_mode.py
@@ -104,3 +104,7 @@ class PaymentMode(models.Model):
                 aj_ids = ajo.search([
                     ('type', 'in', ('sale_refund', 'sale'))]).ids
             self.default_journal_ids = [(6, 0, aj_ids)]
+
+    bank_id = fields.Many2one(required=False)
+    check_required = fields.Boolean(default=True, string="Bank account required",
+        help="This check do bank account optional for payment mode that isn't necessary to inform")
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please rebase as there has been a merge with important changes in these set.

gonxi:
I think that now is OK

account_banking_payment_export/models/payment_mode.py
```diff
@@ -104,3 +104,7 @@ def type_on_change(self):
                 aj_ids = ajo.search([
                     ('type', 'in', ('sale_refund', 'sale'))]).ids
             self.default_journal_ids = [(6, 0, aj_ids)]
+
+    bank_id = fields.Many2one(required=False)
+    check_required = fields.Boolean(default=True, string="Bank account required",
```
pedrobaeza, line 109: You should include the check_required at payment type level, not at payment mode. This way, "Manual Direct Debit" and "Manual Bank Payment" have this check unmarked, and SEPA payment types have it checked. The default value for the field should be True to preseve the same behaviour.

StefanRijnhart:
Agreed with @pedrobaeza. Make this a field on the payment mode type as well, and make the field on payment mode an invisible, related field to the first one (set it in an on-change on the payment mode type).

sbidoul:
@gonxi since this one has not moved since one year, I close it. Feel free to reopen if you continue this work.

PR #255, opened 2016-03-02
[ADD] allow to restrict entries to some journal or analytic account

hbrunn:


pedrobaeza:
The restriction to some journals is already there in last revision. Have you checked?

About analytic accounts, is a bit strange, but if it doesn't mean a performance penalty when you don't fill it, no problem for me.

Another good filter would be by partner if you want to make it at the same time :wink:

sbidoul:
@hbrunn I'm curious about the use case to filter on analytic account. If I remember well, by default, analytic account does not propagate to payable/receivable journal items. So I'd say having this analytic filter by default could be misleading. Perhaps it should be in a separate module?

StefanRijnhart:
@hbrunn any comments on the reviews?

hbrunn:
sorry reviewers, this one drowned somewhere.

@pedrobaeza I can't find the journal filter. Could you point me to what you mean? As for performance: To the best of my knowledge, no performance impact if not selected.

@sbidoul this a quite big organization where different people handle payments on the same (bank and financial) account, but in different domains. So there is a piece of custom code that assigns analytic accounts as necessary to move lines, and this is the last piece they need so that the different people can handle their payments conveniently. If you're right that those fields are empty in a standard setup (can't check that at the moment), this is indeed not helpful for the community, then I'll ditch this PR

hbrunn:
PS: thanks @StefanRijnhart for the heads up

StefanRijnhart:
The journal filter is based on this field: https://github.com/OCA/bank-payment/blob/8.0/account_banking_payment_export/wizard/payment_order_create_view.xml#L16. The default value comes from the payment mode. It seems perfectly fine for your purposes.

The filter on analytic accounts is probably not a very common use case. To change the default of populate_results seems void, as this is overwritten by the default value from the payment mode.

hbrunn:
thanks, then this one is better for private purposes. Will close.

StefanRijnhart:
Thanks for the followup!

PR #258, opened 2016-03-10
[9.0] add german translations to (all) modules

rschnapka:


pedrobaeza:
Why doing this on 9.0 branch that is not migrated?

rschnapka:
maybe to give a head start, in case the changes are minor and not language or interface specific? Otherwise yes it's a good question...

StefanRijnhart:
:+1:

pedrobaeza:
This doesn't make sense now, as all the modules have been heavily reworked on https://github.com/OCA/bank-payment/pull/270, so I close it and when merged, it will be available to translate through Transifex.

PR #260, opened 2016-03-23
8.0 account_payment_partner Better partner_id change on invoice

alexis-via:
partner_id change on customer invoices: when partner.customer_payment_mode is False, don't set 'partner_bank_id' to False, so as to keep the default value if there is one.


pedrobaeza:
:+1:

sbidoul:
:+1:

PR #261, opened 2016-03-31
8.0 FIX selection of mandate when invoice is linked to a contact

alexis-via:
The mandate is linked to a bank account, which itself is linked to a "parent" partner.

But, if partner_id on invoice is a contact (parent_id != False), you cannot select the mandate of the parent contact. This PR fixes this.


pedrobaeza:
You're totally right. :+1:

alexis-via:
@sbidoul Could you review (and merge ?). I'd like to include this kind of fixes before v9 mig

sbidoul:
:+1:

PR #265, opened 2016-04-06
[IMP][8.0] account_banking_mandate:

sergio-teruel:
This PR move SEPA report mandate from **account_banking_mandate** to to **account_banking_sepa_direct_debit** and does a new report more standard.

This PR replaces https://github.com/OCA/bank-payment/pull/262.


account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -35,6 +35,12 @@ def _get_states(self):
                 ('expired', 'Expired'),
                 ('cancel', 'Cancelled')]
 
+    format = fields.Selection(
```
pedrobaeza, line 38: What is the purpose of this field?

pedrobaeza, line 38: It doesn't add too much, because you're making an specific report for SEPA, independently from this field.

pedrobaeza:
The new translations are not in the sepa_direct_debit module, but in account_banking_mandate, so there's no translation at all. You have to take into account also that the language to use is the partner_id language, not user's.

account_banking_mandate/models/account_banking_mandate.py
sergio-teruel, line 38: Only is to have mandates separated by format, basic and SEPA and other new formats.

sergio-teruel:
I lost the es.po file..... yes the report is printed in partner's language

pedrobaeza:
Thanks for the changes. One last thing: the B2B mandate in Spanish doesn't fit in one page, but 2. Please change the font size or remove some spaces to make it fit.

sergio-teruel:
@pedrobaeza changes done, but in runbot the font is not the same....

pedrobaeza:
:+1:

I have already seen that the wkhtmltopdf version makes the default font different, but this size, it's for sure that it will fit.

carlosdauden:
:+1:

rafaelbn:
Hi, tested in runbot :+1: (only english)

pedrobaeza:
As I have tested in Spanish also, merging.

PR #272, opened 2016-05-23
[ADD] inhibit deleting a move line if it belongs to a confirmed bank statement

hbrunn:
I propose this here rather than for the account module in general because this hurts a lot more with a statement import workflow.


sbidoul:
Is that not part of account_constraints in OCA/account-financial-tools?

hbrunn:
thanks, I was not aware of this module!

PR #273, opened 2016-05-23
[FIX] use set_done instead of plain workflow

hbrunn:
without this, `date_done` stays empty forever. In the migration, I try to restore this from `date_scheduled` or the last write date.

### DIFF
--- a/account_banking_fr_lcr/wizard/export_lcr.py
+++ b/account_banking_fr_lcr/wizard/export_lcr.py
@@ -355,7 +355,6 @@ class banking_export_lcr_wizard(orm.TransientModel):
         self.pool['banking.export.lcr'].write(
             cr, uid, lcr_export.file_id.id, {'state': 'sent'},
             context=context)
-        wf_service = netsvc.LocalService('workflow')
-        for order in lcr_export.payment_order_ids:
-            wf_service.trg_validate(uid, 'payment.order', order.id, 'done', cr)
+        self.pool['payment.order'].set_done(
+            cr, uid, [r.id for r in lcr_export.payment_order_ids])
         return {'type': 'ir.actions.act_window_close'}
--- a/account_banking_payment_export/migrations/7.0.0.1.166/post-migrate.py
+++ b/account_banking_payment_export/migrations/7.0.0.1.166/post-migrate.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    """ for done statements, set date_done if not set from date_scheduled """
+    cr.execute(
+        'update payment_order '
+        'set date_done=coalesce(date_scheduled, write_date) '
+        "where state='done' and date_done is null"
+    )
--- a/account_banking_payment_export/model/account_payment.py
+++ b/account_banking_payment_export/model/account_payment.py
@@ -85,9 +85,5 @@ class payment_order(orm.Model):
                           'type')
                     )
             # process manual payments
-            wf_service = netsvc.LocalService('workflow')
-            for order_id in ids:
-                wf_service.trg_validate(
-                    uid, 'payment.order', order_id, 'done', cr
-                )
+            self.set_done(cr, uid, ids)
         return result
--- a/account_banking_sepa_credit_transfer/wizard/export_sepa.py
+++ b/account_banking_sepa_credit_transfer/wizard/export_sepa.py
@@ -307,7 +307,6 @@ class banking_export_sepa_wizard(orm.TransientModel):
         self.pool.get('banking.export.sepa').write(
             cr, uid, sepa_export.file_id.id, {'state': 'sent'},
             context=context)
-        wf_service = netsvc.LocalService('workflow')
-        for order in sepa_export.payment_order_ids:
-            wf_service.trg_validate(uid, 'payment.order', order.id, 'done', cr)
+        self.pool['payment.order'].set_done(
+            cr, uid, [r.id for r in sepa_export.payment_order_ids])
         return {'type': 'ir.actions.act_window_close'}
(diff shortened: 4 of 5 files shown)


StefanRijnhart:
I think that rather, act_done's action should be replaced by a call to set_done.

When using the clearing account workflow, the 'done' signal is rerouted to state 'sent', and only at reconciliation act_done is reached. It seems to me that this change breaks that flow.

hbrunn:
@StefanRijnhart you mean overriding the call to `write` in https://github.com/OCA/OCB/blob/7.0/addons/account_payment/account_payment_workflow.xml#L24 with a call to `set_done` in https://github.com/OCA/bank-payment/blob/7.0/account_banking_payment/workflow/account_payment.xml#L62 ?

StefanRijnhart:
Yes, that's it.

StefanRijnhart:
Oh wait, bad idea. Account_payment's set_done _triggers_ the 'done' signal.

StefanRijnhart:
I can only think of terrible hacks at this point. Keep the workflow trigger and only call set_done if state is indeed 'done'?

hbrunn:
I'm not really familiar with the clearing account workflow in the first place, is this its own module?

StefanRijnhart:
It happens here: https://github.com/OCA/bank-payment/blob/7.0/account_banking_payment/workflow/account_payment.xml#L32

NL66278:
:+1:  @StefanRijnhart I was considering the same solution as @hbrunn  when I checked for a branch already doing this. Seems to me that the only difference implementing this PR will have is correctly filling the execution date, instead of just setting the state through the workflow. So I fail to see the problem in accepting this PR. What do I miss?

hbrunn:
I think @StefanRijnhart's concern is that this writes `date_done` a bit too early. The OCA module changes the semantics of the done signal to "it's sent" (https://github.com/OCA/bank-payment/blob/7.0/account_banking_payment/workflow/account_payment.xml#L32) while before it was "it's done" (https://github.com/OCA/OCB/blob/7.0/addons/account_payment/account_payment_workflow.xml#L46). 
But then I think it's actually only congruent with changing the semantics of `date_done` (this patch writes the field already when the order is in state sent) too, and this is what happens here.

NL66278:
@hbrunn But your PR does not change when the done signal is sent. It only makes sure it also sets date_done. In the present code date_done is certainly not set to early for the simple reason it is never set. So I still think the PR should be accepted. And if people want to have a different moment to set the date, then they should make a PR for that.

hbrunn:
I think we agree and I worded my comment poorly. My patch changes the semantics of `date_done`, because it already sets `date_done` when the order is in state 'sent'. So any code that selects for `date_done is not null` and expects the order to be in state done might break, but then select for the state in the first place.
Also if there's an override of set_done that expects the order to be actually done afterwards, this might be problematic. But then this override is wrong anyways too, because the function is defined in the module that changes the signal.
I consider both quite theoretical, so I'd go on with this patch...

StefanRijnhart:
Initially I thought that this change would undermine the rerouting of the 'done' trigger to the 'sent' state but that does not seem to be the case. About orders in sent state that with this change already get their done date set, would you consider overriding the action of the account_payment.act_done workflow state? https://github.com/odoo/odoo/blob/7.0/addons/account_payment/account_payment_workflow.xml#L27

pedrobaeza:
Why CIs are red?

hbrunn:
because of new checks that this repo doesn't adhere to yet: https://travis-ci.org/OCA/bank-payment/jobs/132231730#L446

pedrobaeza:
I think they are only warnings. The main problem is about module version number I think.

NL66278:
@pedrobaeza It is because of these (not related to merge request) pylint errors:

account_banking_sepa_direct_debit/__init__.py:1: [W7910(wrong-tabs-instead-of-spaces), ] sdd_mandate_view.xml:127 Use wrong tabs indentation instead of four spaces

account_banking_sepa_direct_debit/__init__.py:1: [W7910(wrong-tabs-instead-of-spaces), ] sdd_mandate_view.xml:128 Use wrong tabs indentation instead of four spaces

************* Module account_banking_sepa_direct_debit.__openerp__

account_banking_sepa_direct_debit/__openerp__.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`

account_banking_sepa_direct_debit/__openerp__.py:22: [C8106(manifest-version-format), ] Wrong Version Format "0.2" in manifest file. Regex to match: "(\d+.\d+.\d+.\d+.\d+)"

************* Module account_banking_sepa_direct_debit.account_banking_sdd

account_banking_sepa_direct_debit/account_banking_sdd.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`

************* Module account_banking_sepa_direct_debit.company

account_banking_sepa_direct_debit/company.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`

************* Module account_banking_sepa_direct_debit.wizard.export_sdd

account_banking_sepa_direct_debit/wizard/export_sdd.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`

Found 4 errors in modules changed.

pylint expected errors 0, found 4!

It is on my todo list to fix these as well.

NL66278:
@hbrunn If you rebase on latest 7.0, I have good hope this will be all green.

StefanRijnhart:
@hbrunn curious to know what you think of my suggestion in https://github.com/OCA/bank-payment/pull/273#issuecomment-284710393

hbrunn:
@NL66278 thanks, rebase done
@StefanRijnhart I agree that changing this in odoo core would be best imho, but I can't commit time to this at the moment, and given this is 7, probably never will

StefanRijnhart:
@hbrunn You don't have to do that in the core, you can do it in this PR by changing the code in the workflow activity.

pedrobaeza:
Closing this old and without inactivity. If still interested, please reopen.

PR #270, opened 2016-04-29
[9.0][2nd revolution] port to v9 with several improvements (sorrento code sprint)

alexis-via:
I started to port bank-payment to v9 (with a lot of improvements) during the Sorrento Code sprint 2016.

Improvements include:
- full re-organisation of modules and big re-organisation of the code
- simplification of the code related to the fact that support for direct debit is now in t
  he base module, not added by an optional module
- new design of the wizard to select move lines to pay
- support for non-SEPA file transfer-
- support for German direct debit SEPA files (fixes #129)
- remove workflow

This port to v9 is not finished... there is still a lot of work:
- finish the code of account_payment_order/wizard/account_payment_line_create.py
- port account_banking_payment_transfer and integrate it inside account_payment_order
- fix bugs
- clean-up code, remove dead code
- test in several complex scenarios

PR #274, opened 2016-05-26
Add account_payment_mode

guewen:
This is an extract of `account_payment_mode` from #270 so we can already merge
this part (needed for other addons such as
https://github.com/OCA/sale-workflow/pull/297 and connector-ecommerce).

ping @alexis-via

### DIFF
--- a/account_payment_mode/models/account_payment_method.py
+++ b/account_payment_mode/models/account_payment_method.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
+
+
+class AccountPaymentMethod(models.Model):
+    _inherit = 'account.payment.method'
+    _rec_name = 'display_name'
+
+    code = fields.Char(
+        string='Code (Do Not Modify)',
+        help="This code is used in the code of the Odoo module that handle "
+        "this payment method. Therefore, if you change it, "
+        "the generation of the payment file may fail.")
+    active = fields.Boolean(string='Active', default=True)
+    # In the pain_base module, we will add a default_pain_version
+    display_name = fields.Char(
+        compute='compute_display_name',
+        store=True, string='Display Name')
+
+    @api.multi
+    @api.depends('code', 'name', 'payment_type')
+    def compute_display_name(self):
+        for method in self:
+            method.display_name = u'[%s] %s (%s)' % (
+                method.code, method.name, method.payment_type)
+
+    _sql_constraints = [(
+        'code_payment_type_unique',
+        'unique(code, payment_type)',
+        'A payment method of the same type already exists with this code'
+        )]
--- a/account_payment_mode/models/account_payment_mode.py
+++ b/account_payment_mode/models/account_payment_mode.py
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*-
+# © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api, _
+from openerp.exceptions import ValidationError
+
+
+class AccountPaymentMode(models.Model):
+    """This corresponds to the object payment.mode of v8 with some
+    important changes. It also replaces the object payment.method
+    of the module sale_payment_method of OCA/e-commerce"""
+    _name = "account.payment.mode"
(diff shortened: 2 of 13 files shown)


guewen:
I extracted the module with `git filter-branch` so the commit descriptions might seem inaccurate. @alexis-via let me know if you would prefer that I squash the commits into one more descriptive commit (still authored to you of course).

account_payment_mode/models/account_payment_method.py
```diff
@@ -0,0 +1,34 @@
...
+    _rec_name = 'display_name'
+
+    code = fields.Char(
+        string='Code (Do Not Modify)',
+        help="This code is used in the code of the Odoo module that handle "
+        "this payment method. Therefore, if you change it, "
+        "the generation of the payment file may fail.")
+    active = fields.Boolean(string='Active', default=True)
+    # In the pain_base module, we will add a default_pain_version
+    display_name = fields.Char(
```
pedrobaeza, line 19: The problem with these computed fields is that the translatable nature of the fields is lost (in this case, name is translatable). Is this really needed to be stored?

account_payment_mode/models/account_payment_mode.py
```diff
@@ -0,0 +1,100 @@
...
+        "to suppliers), you should choose 'Variable', which means that you "
+        "will select the bank account on the payment order. If your company "
+        "only has one bank account, you should always select 'Fixed'.")
+    fixed_journal_id = fields.Many2one(
+        'account.journal', string='Fixed Bank Journal',
+        domain=[('type', '=', 'bank')], ondelete='restrict')
+    # I need to use the old definition, because I have 2 M2M fields
+    # pointing to account.journal
+    variable_journal_ids = fields.Many2many(
+        'account.journal',
```
pedrobaeza, line 40: Although using the full definition, you can put keyword arguments to be more verbose about the arguments (comodel_name, relation, column1 and column2).

account_payment_mode/models/account_payment_mode.py
```diff
@@ -0,0 +1,100 @@
...
+
+class AccountPaymentMode(models.Model):
+    """This corresponds to the object payment.mode of v8 with some
+    important changes. It also replaces the object payment.method
+    of the module sale_payment_method of OCA/e-commerce"""
+    _name = "account.payment.mode"
+    _description = 'Payment Modes'
+    _order = 'name'
+
+    name = fields.Char(string='Name', required=True)
```
pedrobaeza, line 17: This one should be also translatable for having different names depending on UI language.

guewen:
@alexis-via what's your opinion on @pedrobaeza points? I can make the changes if you agree.

alexis-via:
I must say I'm not very happy with this PR. We should have merged the big PR https://github.com/OCA/bank-payment/pull/270 instead... it is ready to merge indeed and Travis is green.

pedrobaeza:
OK, do you agree on making optimistic merging here and then make several PRs for patching things? The only thing that we should be flexible then is to relax migration scripts in case of changes on DB layout.

pedrobaeza:
@sbidoul should give also his opinion.

guewen:
> I must say I'm not very happy with this PR. We should have merged the big PR #270 instead... it is ready to merge indeed and Travis is green.

I thought #270 was not yet ready, as the description says "So don't use this PR for the moment... please wait a few days !" and the PR has a work in progress flag. I'm not against an optimistic merge on #270 but it was not clear that it was ready enough for that.

account_payment_mode/models/account_payment_method.py
alexis-via, line 19: The 'name' field of account.payment.method is not translatable, see https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_payment.py#L24

account_payment_mode/models/account_payment_mode.py
alexis-via, line 17: If you need to display the information on invoices to customers in several languages, you should use the note field, which is translatable.

guewen:
I close this PR as we should focus on #270

account_payment_mode/models/account_payment_method.py
pedrobaeza, line 19: OK, I see, then we should turn it into translatable in this module, because IMO this is basic for multi-language UI.

account_payment_mode/models/account_payment_mode.py
pedrobaeza, line 17: I'm not talking about invoices. I'm talking about UI. I have several customers with employees of several languages (due to several franchises along several countries), and each one would one to see this in their language.

alexis-via, line 17: OK, you won ;) I changed it.

PR #276, opened 2016-06-17
[RFR] Stateful amendment info

StefanRijnhart:
I'm proposing to refactor the handling of amendments on SEPA mandates differently, by storing the amendment state on the mandate explicitely. This fixes amendments for switching bank accounts within the same bank, and it allows for a reject trigger so that we can send the same amendment at the next try after a debit reject (note that currently these rejection flows are not otherwise supported in this project).

This change comes complete with a migration script for mandates that had their bank accounts switched out since their last debit order, and a couple of unit tests.

### DIFF
--- a/account_banking_sepa_direct_debit/README.rst
+++ b/account_banking_sepa_direct_debit/README.rst
@@ -19,6 +19,13 @@ Payments Council (http://http://www.europeanpaymentscouncil.eu) use PAIN
 version 008.001.02. So if you don't know which version your bank supports, you
 should try version 008.001.02 first.
 
+Amendments
+----------
+Changes to previously used mandates are passed on to the bank using so-called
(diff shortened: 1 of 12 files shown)


alexis-via:
This seems to be much better than the current implementation. I need to find time to read it carefully...

pedrobaeza:
Although not exactly the same, have you seen about the handle of returned payments in https://github.com/OCA/account-payment/tree/8.0/account_payment_return and derived modules?

pedrobaeza:
@StefanRijnhart, are you going to work on this?

StefanRijnhart:
@pedrobaeza I see there are conflicts here now, I will have a look at those.

As for the integration with returned payments, that seems an obvious extension but not strictly necessary for this change.

StefanRijnhart:
Obsoleted by changes in the SEPA rule books after 2016.

PR #280, opened 2016-07-06
[IMP] Add english lang when a different lang is printed

antespi:
When a mandate is printed in a different language than english, banks often required (as Spanish case) mandate in two languages: native and english.

This PR adds english language when printed language is not english.

### DIFF
--- a/account_banking_sepa_direct_debit/i18n/pt_BR.po
+++ b/account_banking_sepa_direct_debit/i18n/pt_BR.po
@@ -1,7 +1,7 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
 # * account_banking_sepa_direct_debit
-# 
+#
 # Translators:
 # OCA Transbot <transbot@odoo-community.org>, 2016
 msgid ""
@@ -90,16 +90,12 @@ msgstr ""
 
 #. module: account_banking_sepa_direct_debit
 #: model:ir.ui.view,arch_db:account_banking_sepa_direct_debit.sepa_direct_debit_mandate_document
-msgid ""
-"<span class=\"col-xs-12 text-right\" style=\"font-size:8px;\">To be "
-"completed by the creditor</span>"
+msgid "To be completed by the creditor"
 msgstr ""
 
 #. module: account_banking_sepa_direct_debit
 #: model:ir.ui.view,arch_db:account_banking_sepa_direct_debit.sepa_direct_debit_mandate_document
-msgid ""
-"<span class=\"col-xs-12 text-right\" style=\"font-size:8px;\">To be "
-"completed by the debtor</span>"
+msgid "To be completed by the debtor"
 msgstr ""
 
 #. module: account_banking_sepa_direct_debit
--- a/account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
+++ b/account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
@@ -6,6 +6,10 @@
         <t t-set="is_not_english" t-value="doc.partner_id.lang[:2] != 'en'" />
         <div class="page">
             <style type="text/css">
+                h4 {
+                     margin-top: 2px;
+                     margin-bottom: 2px;
+                }
                 .under-line{
                     border-bottom:1px solid;
                 }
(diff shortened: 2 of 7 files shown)


account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
```diff
@@ -17,41 +18,104 @@
...
+                    </h4>
                 </div>
             </div>
             <div class="row mt8">
                 <div class="col-xs-12">
                     <div class="panel panel-default">
-                        <span class="col-xs-12 text-right" style="font-size:8px;">To be completed by the creditor</span>
+                        <div class="col-xs-12 text-right" style="font-size:8px;">
+                            To be completed by the creditor
+                            <t t-if="doc.partner_id.lang not in native_langs">
```
pedrobaeza: Better to just check `<t t-if="doc.partner_id.lang[:2] != 'en'">`. This way, it works for any English variant.

pedrobaeza: As you are using the same t-if everytime, just make a t-set to have a variable that contains this check:

```
<t t-set="is_not_english" t-value="doc.partner_id.lang[:2] != 'en'"/>
```

account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
```diff
@@ -17,41 +18,104 @@
...
+                        </t>
+                    </h4>
                 </div>
             </div>
             <div class="row mt8">
                 <div class="col-xs-12">
                     <div class="panel panel-default">
-                        <span class="col-xs-12 text-right" style="font-size:8px;">To be completed by the creditor</span>
+                        <div class="col-xs-12 text-right" style="font-size:8px;">
+                            To be completed by the creditor
```
pedrobaeza, line 62: This line is not translated to Spanish because you have changed it. Please don't make only cosmetic changes.

pedrobaeza:
Tried on runbot and the problem when having 2 languages is that now the report needs 2 pages. We need to stretch all the layout to accommodate in one page the same for one language or 2.

account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
antespi, line 62: This is not a cosmetic change, I've changed from span to div in order to add inside another span.
Teorically translator dicards spaces before and after string.

Anyway I'm going to check how string is included in .pot and update it in all .po files

antespi, line 62: Strings are captured with HTML tags inside. This is not a good practice, because HTML tagging could change in future.
But I've only changed the two "To be completed ..." sentences

pedrobaeza, line 62: I think there's a reason for that, but I don't remember. Anyways, please change it in the translation.

account_banking_sepa_direct_debit/i18n/pt_BR.po
```diff
@@ -90,16 +90,12 @@ msgstr ""
 
 #. module: account_banking_sepa_direct_debit
 #: model:ir.ui.view,arch_db:account_banking_sepa_direct_debit.sepa_direct_debit_mandate_document
-msgid ""
-"<span class=\"col-xs-12 text-right\" style=\"font-size:8px;\">To be "
-"completed by the creditor</span>"
+msgid "To be completed by the creditor"
```
pedrobaeza, line 93: This is done automatically by Transifex OCA bot, so no need to do it manually if you don't want.

pedrobaeza:
Checked in runbot and everything OK.

:+1: (code and functional review)

rafaelbn:
It would be very nice if you set same font size (the bigger one) in the informative text. It would be much more easy to read.

antespi:
The size is the same, the only difference is that the first one is in lower case and the second one in upper case

rafaelbn:
:+1:

carlosdauden:
:+1:

pedrobaeza:
We can leave the font size/upper case as is, because it's a secondary language, so I merge.

pedrobaeza:
One last remark: the double language is not only for Spain, but for all European Union countries

PR #281, opened 2016-07-06
[ADD] account_payment_restrict_mandate

hbrunn:
# 
# Restrict debits on mandates

This module was written to allow a debit workflow where a customer announces
a maximum amount that can be debited within a given period.
# Usage

To use this module, you need to:
# . configure a banking mandate for your customer with a maximum amount per date
# . adjust the repetition rule, the default is every month
# . fill in a start and end date for which the repetition rule applies
# . when selecting move lines in a payment order, amounts will be capped to the

   maximum amount given in the mandate.
# . in case you have missed debit dates according to the repetition rule, the

   maximum amount is multiplied by the number of dates missed

If this does not work as expected, check the following
- is the invoice the move is originating from linked to the mandate?
- is the mandate in state 'valid'?
- is there an occurrence in the repetition rule falling between today (or the
  scheduled date if it's a fixed order) and the mandate's starting date or
  the mandate's last date of debit?

.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/173/8.0
# Known issues / Roadmap
- we need a widget for FieldRrule
- there should be an extra column when selecting move lines showing how much
  can be debited according to the rule (this is going to be a performance killer)
- when payment lines are adjusted or altogether removed, there should be a warning

This is work in progress until at least the horrible UI for repetition rules is fixed. For the time being, consult https://labix.org/python-dateutil#head-cf004ee9a75592797e076752b2a889c10f445418 for possible values in the repetition text field. For the freq parameter, use the integer value instead of the symbolic name. (0=yearly, 1=monthly, 2=weekly, etc)


account_payment_restrict_mandate/tests/test_account_payment_restrict_mandate.py
```diff
@@ -0,0 +1,42 @@
...
+from openerp.tests.common import TransactionCase
+
+
+class TestAccountPaymentRestrictMandate(TransactionCase):
+    def test_account_payment_restrict_mandate(self):
+        mandate = self.env['account.banking.mandate'].create({
+            'partner_bank_id': self.env.ref('base.res_partner_2').bank_ids
+            .filtered(lambda x: x.state == 'iban').id,
+            'max_amount_per_date': 600,
+            'date_start': '2016-01-01 00:00:00',
```
StefanRijnhart: I strongly suspect that this test will break 6 months from now. Can you take current year instead?

StefanRijnhart:
~~:+1: apart from the test nag~~  edit: just noticed WIP label

account_payment_restrict_mandate/tests/test_account_payment_restrict_mandate.py
hbrunn: I used fixed dates on purpose to exclude any date dependent issues. Note that the code that looks for the end date prefers `date_scheduled` over `datetime.now()`: https://github.com/OCA/bank-payment/pull/281/files/9cad92470583a29415463b676b5a6ae359d27e12#diff-e1403f56fc522add18775d3d1b68ca1cR23
Or where do you expect it to break?

account_payment_restrict_mandate/README.rst
```diff
@@ -0,0 +1,81 @@
...
+  the mandate's last date of debit?
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+    :alt: Try me on Runbot
+    :target: https://runbot.odoo-community.org/runbot/173/8.0
+
+Known issues / Roadmap
+======================
+
+* we need a widget for FieldRrule
```
hbrunn: I'm busy with this right now in https://github.com/OCA/server-tools/issues/480

account_payment_restrict_mandate/tests/test_account_payment_restrict_mandate.py
StefanRijnhart: I was thinking about the missing fiscal year, but now it occurred to me
that this probably is not a problem because no moves are actually
created in the test.

hbrunn: this will not fail, but now I realize the test will fail next year because it presupposes the moves from the invoice have a fixed date, but that's not true: https://github.com/OCA/OCB/blob/8.0/addons/account/demo/account_invoice_demo.xml#L5 - thanks for this point, will save us some pain and misery in January

StefanRijnhart:
What's with the runbot? Do we need to add the server-tools repo there?

account_payment_restrict_mandate/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+{
+    "name": "Restrict debits on mandates",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Accounting & Finance",
+    "summary": "Set maximum amount per period on a mandate",
+    "depends": [
+        'account_banking_mandate',
+        'field_rrule',
```
pedrobaeza, line 13: Please add in the installation section of the README that you need this module from server-tools

account_payment_restrict_mandate/models/account_banking_mandate.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+from openerp import api, fields, models
+from openerp.addons import decimal_precision
+from openerp.addons.field_rrule import FieldRRule
```
pedrobaeza: Please protect this import

pedrobaeza:
I have added server-tools repo to runbot and launch a rebuild.

StefanRijnhart:
@pedrobaeza thanks! Green now.

account_payment_restrict_mandate/__openerp__.py
hbrunn: @pedrobaeza done

hbrunn:
rebased to make codecov happy

PR #282, opened 2016-07-12
[RFR] Always post entries

StefanRijnhart:
Autopost setting on journal considered harmful: https://github.com/odoo/odoo/pull/12014

### DIFF
--- a/account_banking_payment_transfer/model/account_payment.py
+++ b/account_banking_payment_transfer/model/account_payment.py
@@ -245,9 +245,7 @@ class PaymentOrder(models.Model):
                     total_amount, move, blines, labels)
                 aml_obj.create(trf_ml_vals)
                 self._reconcile_payment_lines(blines)
-                # consider entry_posted on account_journal
-                if move.journal_id.entry_posted:
-                    move.post()
+                move.post()
 
         # State field is written by act_sent_wait
         self.write({'date_sent': fields.Date.context_today(self)})


pedrobaeza:
I think @sbidoul bases some logic on the flag in the journals, so it can't be easily removed.

sbidoul:
@StefanRijnhart @pedrobaeza yes we do have customers relying on this logic.

Why do you want to change it? it's easy to configure on the journal.

StefanRijnhart:
Well, check the linked issue: https://github.com/odoo/odoo/pull/12014. It is allowed to edit posted moves in journals with this option set. Therefore, using this option is not acceptable to my customers. Is it for yours?

sbidoul:
@StefanRijnhart I don't have the exact use case in mind right now, I'd need to check with the project manager. In any case it's a question of workflow and user rights.

But since it's just a matter of enabling entry_posted on the journal to have it posted automatically it's quite easy to configure.

In general we have two approaches to handle error correction and/or entries verification:
- working with Odoo account_cancel and allowing cancelling entries on a per journal basis
- working with OCA account_default_draft_moves (which boils down to honoring the entry_posted flag everywhere) and not installing account_cancel at all

In both cases it's possible to mess things up for a user with sufficient permissions.

StefanRijnhart:
No, it should not be so simple to mess it up _that_ simple as to be able to inadvertently change amounts or delete lines from the move line editable list view when it concerns posted moves. The requirements of my customer are really quite simple: moves from payment orders should be posted automatically, and posted moves should not be editable. As it does not seem to be possible, there has to be a bug _somewhere_.

I do see that the usage here mimics the usage of the field in parts of core Odoo such as hr_payroll_account, so it would be great if you could support the case of https://github.com/odoo/odoo/pull/12014 instead and we'll close this.

sbidoul:
@StefanRijnhart Its indeed a serious bug you raised there. I pushed as much as I can.

StefanRijnhart:
Thanks!

PR #285, opened 2016-07-19
[IMP] account_payment_order: extensibility of _prepare_move

sbidoul:
Pass the list of bank lines to _prepare_move so it is possible to customize the move (eg it's name) based on the lines being paid.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -330,7 +330,7 @@ class AccountPaymentOrder(models.Model):
         return True
 
     @api.multi
-    def _prepare_move(self):
+    def _prepare_move(self, bank_lines=None):
         if self.payment_type == 'outbound':
             ref = _('Payment order %s') % self.name
         else:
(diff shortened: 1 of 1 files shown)


account_payment_order/models/account_payment_order.py
```diff
@@ -330,7 +330,7 @@ def generated2uploaded(self):
         return True
 
     @api.multi
-    def _prepare_move(self):
+    def _prepare_move(self, bank_lines=None):
```
pedrobaeza, line 333: Where is bank_lines used?

sbidoul, line 333: It's not used in this module, but it is useful when overriding `_prepare_move' so some parts of the move can be customized in extension modules.

pedrobaeza, line 333: Then is the purpose to change bank payment lines? Then that should be done in another method I think

sbidoul, line 333: No, I don't want to change the bank payment lines. I want to compute `ref` from the bank payment lines.

pedrobaeza, line 333: OK, understood, thanks.

pedrobaeza:
:+1:

alexis-via:
ok, why not :+1:

sbidoul:
@alexis-via in scenarios where we don't group payment lines, it's good to generate more precise refs and labels in account moves.

pedrobaeza:
Merging as harmless and a default value has been provided.

PR #289, opened 2016-07-21
[7.0] [TEST] added new test to account_banking_pain_base

zeroincombenze:
Fixed some PEP8 and added a simple test.
Test is designed to be used by other local country groups.
Please, notice this is a first limited test.
I would like to add more tests in next time.


pedrobaeza:
Travis is red and please clean commit history.

PR #290, opened 2016-07-22
9.0 FIX crash when selecting a customer that has no mandate + small improvements

alexis-via:
This PR changes 3 things:
- FIX crash when selecting a customer that has a direct debit payment mode but no mandate
- Hide many fields when payment_order_ok is False on a payment_mode (usability improvement)
- Fix payment line communication when the move line is not linked to an invoice (problem found when porting donation_debit_debit to v9)

This PR is needed to make the unittests of my PR to port the donation modules to v9 work https://github.com/OCA/donation/pull/13 (requires the last change in my list of 3 changes)

### DIFF
--- a/account_banking_mandate/models/account_invoice.py
+++ b/account_banking_mandate/models/account_invoice.py
@@ -59,7 +59,8 @@ class AccountInvoice(models.Model):
                 ('state', '=', 'valid'),
                 ('partner_id', '=', self.commercial_partner_id.id),
                 ])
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1: with the change you have just made fixing Travis.

sbidoul:
:+1:

pedrobaeza:
I merge because the fix is clear

PR #292, opened 2016-08-04
[8.0][FIX]account_payment_purchase: payment mode in purchase created from procurement

jesusVMayor:
in multicompany enviroments.


pedrobaeza:
Indeed being a property field and the procurement being run by administrator we need this.

Thanks.

:+1:

alexis-via:
:+1:

PR #294, opened 2016-08-24
[IMP] set partner on offsetting account move line if possible

### LINKED ISSUE #277: [9.0] account_payment_order: propagate partner to offsetting account when possible
When the transfer move has one move line line per partner, it is useful to propagate the partner to the move line on the offsetting account.

@alexis-via do you agree?

https://github.com/OCA/bank-payment/blob/d2e87f74c04f413a793c7c12ec3d440bda6daf41/account_payment_order/models/account_payment_order.py#L358

sbidoul:
fixes #277


pedrobaeza:
:+1:

sbidoul:
@alexis-via ok with this one?

rafaelbn:
:+1:

pedrobaeza:
As @alexis-via is not answering, I merge for now and if there's any problem, he can tell us later.

PR #296, opened 2016-08-26
[ADD][8.0]'payment_order_create' add suppliers refunds payment lines

daviddiz:
When create a payment.order of type payment and launch the wizard to propose lines,
propose not only the credit lines, add too the refunds from the suppliers.

The sense of adding this refund invoices is to take care of them and not to pay invoices from the same supplier with out decreasing the amount of the refund.

To do that i only extend the function "**extend_payment_order_domain**" of the model "**payment.order.create**" in the module "**account_banking_payment_export**", 
in the domain i change the expression ('credit', '>', 0) with :

('credit', '>', 0) 
or 
( ('debit', '>', 0) and ('account_id.type', '=', 'payable') and ('journal_id.type', '=', 'purchase_refund') )


pedrobaeza:
Isn't better to include this in the root module instead of creating a new module?

daviddiz:
If you think that must be this way, there's no problem for me to include it in the account_banking_payment_export module

daviddiz:
Do it, only change a few lines in the base module as you said

pedrobaeza:
Thanks for the changes, but as stated in the list, you have to include an option in the selection dialog (a boolean field) called "Include refunds" to activate or not this option, so it's actually a bit more code. With this option (disabled by default), you don't alter current behavior, but allowing also the new one marking the check.

daviddiz:
Ok i will make that option

daviddiz:
I include that option in the selection dialog but for activate this option i cannot do it in the same way that other options like "date_type" or "invoice related", because i have to change the previous existing conditions in the domain, i must include this option with an if in the "extend_payment_order_domain" function

daviddiz:
I have to complete with the translations

daviddiz:
I think this is the better approach:
- Always visible
- Position just before "populate_results"

Translations  are proposed but only the spanish has been done.

astirpe:
One of our Customers asked us to fix module `account_banking_payment_export`, by including the refund invoices in the filter domain. This PR seems to fix exactly this case. Thank you @daviddiz for this.

However I think that adding the selection dialog "Include refunds" is not necessary. The reason is that the filter already takes in consideration the journal tags (field `journal_ids`) specified in the wizard. For example, if you include the journal called "Purchase Refund Journal" in the list of journals, you implicitly expect that the refunds are taken into account. So the filter domain should always allow the refund invoices. @pedrobaeza what do you think?

@daviddiz could you rebase the branch and update the module version according to https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers ?

pedrobaeza:
This one is complex as if you include refunds, you have to deal also with inverse amounts to be put on payment lines, and the possibility of having lines of both types for the same customer/supplier. That's why refund support was discarded in the beginning. Any way, if all the aspects of including them are taken into account, go ahead.

PR #284, opened 2016-07-16
[FIX] travis

hbrunn:

PR #297, opened 2016-08-29
[FIX] Check that move lines have an invoice in _match_invoice()

StefanRijnhart:
This fixes a crash under rare circumstances where move lines are selected without an invoice, ending with

```
account_banking/banking_import_transaction.py", line 812, in _get_move_info
    retval['type'] = type_map[move_lines[0].invoice.type]
KeyError: None
```

### DIFF
--- a/account_banking/banking_import_transaction.py
+++ b/account_banking/banking_import_transaction.py
@@ -255,7 +255,7 @@ class banking_import_transaction(orm.Model):
         if partner_ids:
             candidates = [
                 x for x in move_lines
-                if x.partner_id.id in partner_ids and
+                if x.invoice and x.partner_id.id in partner_ids and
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You should check pylint errors.

StefanRijnhart:
The pylint errors and warnings are unrelated to this change. I would do it in another PR, but don't think modernizing this deprecated module is very useful.

StefanRijnhart:
@pedrobaeza: @hbrunn already updated the code in https://github.com/OCA/bank-payment/pull/284. If you review it, I will rebase and we can check the current state of affairs.

pedrobaeza:
You can rebase now

StefanRijnhart:
And Travis is green. Thanks @hbrunn for that #284!

pedrobaeza:
:+1:

PR #298, opened 2016-09-07
[9.0] account_payment_sale : change debit account on invoice

mdietrichc2c:
Small PR to update the account on the invoice depending on the payment mode.

The goal here is that, if an invoice is set with a payment mode that uses a fixed journal, the entries created for payment will use the default debit/credit accounts, and to reconcile them with the invoice, they should use the same account.

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -28,6 +28,8 @@ class SaleOrder(models.Model):
         if self.payment_mode_id:
             vals['payment_mode_id'] = self.payment_mode_id.id
             if self.payment_mode_id.bank_account_link == 'fixed':
-                vals['partner_bank_id'] =\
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Is this still needed with latest changes? (see also opened PRs)

leemannd (approved):

leemannd:
Hello @pedrobaeza I'm coming back at you. I wonder if we can merge this.
I'm currently cleaning code we have in production and this one is present since end 2016.
Do you remember to wich PR you were refering to?

pedrobaeza:
@leemannd FYI we have removed this part in v10/v11 as it was conflicting when using SEPA DD. See https://github.com/OCA/bank-payment/pull/458/files#r175775484

leemannd:
@pedrobaeza I didn't knew. I'm going to take a look at it.

PR #302, opened 2016-09-21
[FIX] account_banking_payment_export: Change sequence and noupdate

pedrobaeza:
The new bank.payment.line sequence has the same prefix as the old payment.line, 
so this makes that installations coming from the previous version where there were
no bank.payment.line, in an effective way this provokes to restart the numbering
of the references sent to the bank and thus the possibility of duplicating numbers.

The implicit noupdate="0" also does the same effect of restarting the numbering.

### DIFF
--- a/account_banking_payment_export/data/bank_payment_line_seq.xml
+++ b/account_banking_payment_export/data/bank_payment_line_seq.xml
@@ -6,7 +6,7 @@
 -->
 
 <openerp>
-<data>
+<data noupdate="1">
 
 <record id="bank_payment_line_seq_type" model="ir.sequence.type">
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):
👍

pedrobaeza:
@sbidoul @alexis-via your feedback and fast-tracking here is welcome.

sbidoul:
:+1: cc/ @adrienpeiffer

alexis-via:
:+1:

adrienpeiffer:
:+1:

pedrobaeza:
Can you merge then or should I?

Thanks for the speed!

PR #303, opened 2016-09-23
8 .0 Update to new SEPA rules that start on 2016-11-20

alexis-via:
Update to EPC Rulebook v9.2 that start to apply on 2016-11-20, cf http://www.europeanpaymentscouncil.eu/index.cfm/knowledge-bank/epc-documents/epc-introductory-note-on-changes-incorporated-into-the-sepa-direct-debit-sdd-core-rulebook-version-90-and-sdd-business-to-business-rulebook-version-70/

Remove old fields that were only usefull during SEPA migration in 2014 (these stuff have already been removed in v9)

I think this should be merged on 2016-11-20, not before.


pedrobaeza:
The new EPC Rulebook v9.2 is itself the removal of the obsolete fields, or is there something more?

alexis-via:
The removal of the obsolete fields is unrelated to the new rulebook.

The new rulebook is about the scenario when the customer has a new bank account and how it translates in the SEPA DD XML file.

pedrobaeza:
I t would be good then to split both changes in separate PRs and maybe we can add a new export type with the new rulebook so that we can incorporate it as soon as we want.

alexis-via:
I made the 2 changes in the same PR because it impacts exactly the same piece of code in wizard/export_sdd.py

In fact, I initially only wanted to make the change for the new rulebook, and as I was modifying the code in wizard/export_sdd.py, I felt that it was time to remove sepa_migrated because it impacted the same piece of code and it was a unique opportunity to simplify it.

StefanRijnhart:
Take this as an approval to merge after 2016-11-20. @alexis-via with regards to amendments, did you ever consider https://github.com/OCA/bank-payment/pull/276, to keep amendments alive across possible rejects?

aetna-softwares:
Hi, I think that the SMNDA change should also have a PR for 9.0 branch (it is already commited in 10.0 branch but not in 9.0)

alexis-via:
@aetna-softwares yes, feel free to propose the PR for v9.

pedrobaeza (approved):

pedrobaeza:
Merging as the date has come.

PR #306, opened 2016-10-28
Fix error on amendment

aetna-softwares:
When enter in amendment case, the code tried to access a non existent "date_done" field and to use a non existant _validate_iban function

It seems that the bug is present since 9.0 port


alexis-via:
If you up-port the PR "Update to new SEPA rules that start on 2016-11-20" https://github.com/OCA/bank-payment/pull/303 (or better: backport it from v10), all the code that contain the bug you found will be removed anyway...

aetna-softwares:
I agree but there is still one month to go until new rules applied so I need the fix in the meantime.

I did the PR in the case someone else also need it

PR #307, opened 2016-10-28
10.0 Remove method which is not called anymore

alexis-via:


pedrobaeza (approved):

hbrunn:
runbot error unrelated

PR #308, opened 2016-10-31
[FIX] use setattr, otherwise we can't set the attribute

hbrunn:
fields use slots, so we can't naively set the attribute


pedrobaeza (approved):
I haven't found any problem with this for now, but the proposed changed seems OK for me.

adrienpeiffer (approved):

PR #313, opened 2016-11-14
[ADD] set account type in demo data

hbrunn:
the story here is that this module relies on `base_iban` to set the correct account type, but this doesn't happen if the module is installed before `base_iban`, making the tests in `account_banking_sepa_credit_transfer`. This seems to be the least invasive fix for this.


pedrobaeza (approved):
I agree with the solution

sbidoul (approved):

alexis-via (approved):

PR #317, opened 2016-12-01
[IMP] Define SEPA category purpose on payment mode

sandrafig:
This PR defines the **SEPA category purpose** code on the payment mode, according to **ISO 20022**:
 https://www.iso20022.org/standardsrepository/public/wqt/Description/mx/dico/codesets/_ZzX8ldp-Ed-ak6NoX_4Aeg_-191357927

Basically what this code does, is that enables you to add a category on a credit transfer, so you can specify if the transaction corresponds to a payment of salaries, taxes, etc. 

I want to credit @pedrorgil from Otherway Creatives, because I just adapted the code I found on one of his repositories.

### DIFF
--- a/account_banking_pain_base/models/banking_export_pain.py
+++ b/account_banking_pain_base/models/banking_export_pain.py
@@ -210,7 +210,17 @@ class BankingExportPain(models.AbstractModel):
             sequence_type_2_14 = etree.SubElement(
                 payment_type_info_2_6, 'SeqTp')
             sequence_type_2_14.text = sequence_type
-
+        # Set CtgyPurp for pain.001.001.002 and pain.001.001.003
+        if gen_args.get('pain_flavor') == 'pain.001.001.02' or gen_args.get(
+                'pain_flavor') == 'pain.001.001.03' or gen_args.get(
+                'pain_flavor') == 'pain.001.001.04' or gen_args.get(
+                'pain_flavor') == 'pain.001.001.05':
+            category_purpose_2_14 = etree.SubElement(
+                payment_type_info_2_6, 'CtgyPurp')
+            category_purpose_code_2_14 = etree.SubElement(
+                category_purpose_2_14, 'Cd')
+            category_purpose_code_2_14.text = gen_args.get(
+                'category_purpose_code')
         if gen_args['payment_method'] == 'DD':
             request_date_tag = 'ReqdColltnDt'
         else:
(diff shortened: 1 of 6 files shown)


alexis-via:
I don't think this property belongs to the payment mode... you are supposed to use the same payment mode for many different kinds of payment. I would say it's a property of a payment line...

For curiosity, it this field required in your country ? If yes, which country is it ?

pedrobaeza:
Can you mix several purposes in the same payment order? If so, the approach of defining at payment mode level doesn't serve.

Please check Travis status.

pedrobaeza:
@alexis-via She is from Spain, my country (you know :smile: ), and it's mandatory when sending payment orders for salaries (for deductions and so on).

alexis-via:
ok, very interesting. I'm just surprised that this requirement appears only now... how did they handle that before ? In a localisation module ?

sandrafig:
This tag is at head level on the payment order xml file, so I don't think you can define it by payment line; but correct me if I'm wrong.

Yes, I'm from Spain. This field is required just as Pedro said, when the transaction is for the payment of a salary. If you don't specify the category purpose, the bank takes it as if it were a simple credit transfer and does not recognize it as a salary.

What I have been doing until now, is to write directly the tag in the xml file.

pedrobaeza:
Yeah, this has been handled with this module https://github.com/OCA/l10n-spain/tree/8.0/l10n_es_account_banking_sepa_fsdd or manually indicating the purpose in the moment of uploading the file (most of the bank online platforms have the ability to indicate via a selector or in a separate menu).

pedrobaeza:
Sorry, I have mixed things. The localization module is for other purpose. The method has been the manual one (we are doing this way).

sandrafig:
Now Travis is green but runbot fails, I don't know why

pedrobaeza:
Runbot failure is due to something unrelated, so no worry about that: https://github.com/OCA/runbot-addons/issues/111

sandrafig:
I just did a rebase and squashed all of the commits into one

alexis-via:
As I explained in my first comment, I disagree with the datamodel proposed in this PR: category purpose is a property of payment line, not a property of payment mode.
To be constructive, I made a PR with what I think is the right datamodel for category purpose: https://github.com/OCA/bank-payment/pull/366  @sandrafig Could you have a look at this other PR ?

PR #319, opened 2016-12-03
[9.0][FIX] account_payment_order: unbind payment transfer moves without deleting them

sbidoul:
When unreconciling a bank statement line that is matched with a bank
account move lines linked to a bank.payment.line, it must be unbound,
not deleted as the standard Odoo method would do.

Note the standard Odoo method has such protection for their native
payment lines.

### DIFF
--- a/account_payment_order/models/__init__.py
+++ b/account_payment_order/models/__init__.py
@@ -8,3 +8,4 @@ from . import account_move
 from . import account_move_line
 from . import account_invoice
 from . import res_bank
+from . import bank_statement_line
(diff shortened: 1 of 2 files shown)


sbidoul:
Without this, transfer moves are easily deleted by the user in the bank reconciliation process.

@alexis-via @adrienpeiffer

susport (commented):
Can you add the test data please. Thank.

sbidoul:
In v10 this works fine, so this PR is only required on 9.0

alexis-via (approved):

adrienpeiffer:
:+1:

PR #320, opened 2016-12-06
[ADD] account_payment_mode: Migrate from 8.0 if possible in init_hook

hbrunn:


alexis-via (approved):
code review, no test

pedrobaeza:
But the pre-init hook is only run when installing the module, isn't it? Shouldn't we do this on a regular migration script?

hbrunn:
It's a new module, so no migration scripts are run

pedrobaeza:
Do you plan to activate in OpenUpgrade the installation of the module? Possibly in the migration of account_bank_payment_export, that has to be renamed to account_payment_order?

account_payment_mode/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+    migrate_from_8(cr)
+
+
+def migrate_from_8(cr):
+    """If we're installed on a database which has the payment_mode table
+    from 8.0, move its table so that we use the already existing modes"""
+    cr.execute("SELECT 1 FROM pg_class WHERE relname = 'payment_mode'")
+    if not cr.fetchone():
+        return
+    cr.execute('ALTER TABLE payment_mode RENAME TO account_payment_mode')
```
pedrobaeza: We should use openupgradelib method for renaming models, so that we don't lose for example possible attachments linked to these records (rare, but possible).

hbrunn: do we have a policy of making this a hard dependency (bad) or do we copy code (awful) or should we try to import openupgradelib, if so, use this, if not use sql (not so much better)?

pedrobaeza: I think we should put a hard dependency, as this is only executed when migrating from v8 to v9, and openupgradelib is expected in that environment.

pedrobaeza: Remember that now these scripts are not executed on module install.

pedrobaeza: Are you going to change this, @hbrunn ?

hbrunn: hey, the day only has so many hours...

pedrobaeza: Sorry, it was not a reproach. Only to know if I have to make a PR to your branch.

pedrobaeza (approved):

pedrobaeza:
As you verify your commits, can you please squash them before merging for not me doing that?

hbrunn:
rebased. But I'd be happy to have you squashing commits too, only if something goes wrong afterwards, I would claim that it's not my code lacking my signature ;-)

PR #321, opened 2016-12-09
[9.0][MIG]account_banking_mandate: Scritp migration to v9

sergio-teruel:
cc @tecnativa

### DIFF
--- a/account_banking_mandate/migrations/9.0.1.0.0/post-migration.py
+++ b/account_banking_mandate/migrations/9.0.1.0.0/post-migration.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    # Copy mandate_id to account_move_line
+    sql = """
+    UPDATE account_move_line aml
+    SET mandate_id = ai.mandate_id
+    FROM account_invoice ai
+    WHERE aml.invoice_id=ai.id
+    AND aml.invoice_id is not null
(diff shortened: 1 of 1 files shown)


account_banking_mandate/migrations/9.0.0.2.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    cr = env.cr
+    # Copy mandate_id to account_move_line
+    sql = """
+    UPDATE account_move_line aml SET mandate_id = (
```
pedrobaeza: Please use compact form that saves one query:
```
UPDATE account_move_line aml 
SET mandate_id = ai.mandate_id
FROM account_invoice ai
WHERE aml.invoice_id=ai.id 
AND aml.invoice_id is not null
```

pedrobaeza (changes requested):
Version folder should be 9.0.1.0.0

sergio-teruel:
@pedrobaeza changes done.
ups... version

account_banking_mandate/migrations/9.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    # Copy mandate_id to account_move_line
+    sql = """
+    UPDATE account_move_line aml
+    SET mandate_id = ai.mandate_id
+    FROM account_invoice ai
+    WHERE aml.invoice_id=ai.id
+    AND aml.invoice_id is not null;
```
pedrobaeza: I'm thinking that you should also add `AND aml.mandate_id IS NOT NULL` for avoiding extra operations.

sergio-teruel:
@pedrobaeza changes done

pedrobaeza (approved):

carlosdauden (approved):

PR #325, opened 2016-12-16
[9.0][MIG] account_payment_order: Migration script

sergio-teruel:
WIP
cc @tecnativa

### DIFF
--- a/account_payment_order/migrations/9.0.1.1.1/post-migration.py
+++ b/account_payment_order/migrations/9.0.1.1.1/post-migration.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def map_payment_type(cr):
+    openupgrade.map_values(
+        cr,
+        'payment_order_type', 'payment_type', [
+            ('payment', 'outbound'),
+            ('debit', 'inbound'),
+        ],
+        table='account_payment_order', write='sql')
+
+    openupgrade.map_values(
+        cr,
+        openupgrade.get_legacy_name('state'), 'state',
+        [('done', 'uploaded')],
+        table='account_payment_order', write='sql')
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    cr = env.cr
+    map_payment_type(cr)
+
+    cr.execute("""
+       UPDATE account_payment_order apo
+       SET journal_id=apm.fixed_journal_id
+       FROM account_payment_mode apm
+       WHERE apo.payment_mode_id = apm.id
+    """)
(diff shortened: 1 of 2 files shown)


account_payment_order/migrations/9.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,92 @@
...
+    ],
+    'bank_payment_line': [
+        ('transit_move_line_id', None),
+    ],
+}
+
+
+def install_new_modules(cr):
+    sql = """
+    UPDATE ir_module_module SET state='to install' WHERE name in %s
```
pedrobaeza: I think this is automatically handled by OpenUpgrade. Please check and remove if so. If not, please include the condition that they are in state='uninstalled' (they can be already installed in the source database).

account_payment_order/migrations/9.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,92 @@
...
+# account_payment_order
+to_install = [
+    'account_payment_mode',
+    'account_payment_partner',
+]
+
+table_renames = [
+    ('payment_order', 'account_payment_order'),
+    ('payment_line', 'account_payment_line'),
+    ]
```
pedrobaeza: De-indent this line

account_payment_order/migrations/9.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+# V9 modules that are not exist in v8 and are dependent of
```
pedrobaeza: s/that are not exist/that don't exist

account_payment_order/migrations/9.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,92 @@
...
+        openupgrade.rename_columns(env.cr, column_renames_payment_export)
+
+    if openupgrade.is_module_installed(
+            env.cr, 'account_banking_payment_transfer'):
+        openupgrade.rename_columns(env.cr, column_renames_payment_transfer)
+        openupgrade.update_module_names(
+            env.cr,
+            [('account_banking_payment_transfer', 'account_payment_order')],
+            merge_modules=True)
+    openupgrade.rename_tables(env.cr, table_renames)
```
pedrobaeza, line 99: Please use a function that I have just discovered that is more complete (also handles attachments and so on) and that applies here, as we are not renaming to None, but to a still useful table: _rename_models_ - https://github.com/OCA/openupgradelib/blob/c723fa4ccf288184d163d9daa211c3dc1281c263/openupgradelib/openupgrade.py#L362

account_payment_order/migrations/9.0.1.1.1/post-migration.py
```diff
@@ -0,0 +1,34 @@
...
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def map_payment_type(cr):
+    openupgrade.map_values(
+        cr,
+        'payment_order_type', 'payment_type',
```
pedrobaeza: Both fields exists in new module?

pedrobaeza (changes requested):

pedrobaeza:
Travis is red

sergio-teruel:
@pedrobaeza Changes done

pedrobaeza (approved):

pedrobaeza:
Merging optimistically as OpenUpgrade's policy.

PR #326, opened 2016-12-29
[FIX] we also need to rename our table

hbrunn:
fixup for #320

### DIFF
--- a/account_payment_mode/hooks.py
+++ b/account_payment_mode/hooks.py
@@ -14,8 +14,9 @@ def migrate_from_8(cr):
     if not cr.fetchone():
         return
     try:
-        from openupgradelib.openupgrade import rename_models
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Yeah, that's something I came up with yesterday. Why the method doesn't rename also the columns?

pedrobaeza (approved):
Fast-tracking when CIs finish

hbrunn:
thanks for the fast track. I think the reasoning for this being two functions is that the orm doesn't necessarily exist when you call those functions, and only the orm knows reliably what the table backing a model is, that's not persisted somewhere in the database to my knowledge.

pedrobaeza:
Yeah, you're right. I have added a warning in the method docstring as reminder: https://github.com/OCA/openupgradelib/commit/39416d214750a1dbb7ec7f4eb99e0fc1c7cc4d89

Issue #329, opened 2017-01-11
Partner Bank Account not added when creating a SEPA credit transfer to suppliers from journal items


DukeOfHammond:
When creating a SEPA credit transfer to suppliers with the function "create payment lines from journal items", the partner bank account is not written in the table and it has to be added manually.

pedrobaeza:
You always have to mention the Odoo version

DukeOfHammond:
v9 community

DukeOfHammond:
do you have an ideo when this could be fixed?

pedrobaeza:
Never if nobody cares about it. Take into account that this is a voluntary community. These are the possibilities:

* You fix on your own.
* Even better: you fix on your own and create a PR here for the rest enjoining your patch.
* Somebody feels generous and fix it for you here, and you can enjoy it.
* Somebody fixes it in return for something you have made (help in other thing, review another PR, great contributor... which is not your case).
* You pay someone to fix it, and optionally (but recommended), put here the patch.

DukeOfHammond:
i do have a fix, how can i upload it here?

pedrobaeza:
Edit the file inline in GitHub, and it will automatically a create a PR (pull request).

alexis-via:
If the bank account is set on the move line (which is the case if the supplier invoice has the bank account of the supplier), then the bank account will be set on the payment line when you use the wizard "create payment lines from journal items", thanks to this line of code:

https://github.com/OCA/bank-payment/blob/9.0/account_payment_order/models/account_move_line.py#L51

I made a test today on odoo v9 and I confirm it works well.

PR #330, opened 2017-01-12
Update account_move_line.py

DukeOfHammond:
Fix to issue #329 
When creating a SEPA credit transfer to suppliers with the function "create payment lines from journal items", the partner bank account is not written in the table and it has to be added manually.


pedrobaeza:
@alexis-via, can you check?

alexis-via:
I don't agree with this change, but I'll comment in more details later on (too late now).

alexis-via:
I close this PR because the bug is wrong, as I explained on the bug report.

PR #331, opened 2017-01-12
[10.0] Creation/Modification of bank info only for Account Payment group

alexis-via:
By default in Odoo, members of the group *Contact Creation* can create and modify bank accounts ; this can be a risk, as explained in this mail : https://lists.launchpad.net/openerp-community/msg01035.html

With this module, only the members of the group *Accounting / Payments* can create and modify bank accounts. Now that we took back control of the base account payment features, I think it's time to have that configuration by default.


sbidoul (approved):

pedrobaeza:
If we limit this, we should grant also permissions to salesmen, because at least in our country, they are the one that talk to the customer and ask for this data, so we need to set a rule access for restricting the same way as sale.order for "See only own leads" and the ACL for these groups.

alexis-via:
Another option would be to create a new group "Bank account Manager", that has create/write access on res.partner.bank and res.bank.

pedrobaeza:
I prefer to keep existing groups, there's no possibility of one not having accounting or sales roles but creating bank accounts. There are already too much groups...

alexis-via:
As it seems there is a consensus on this proposal, I plan to merge it tomorrow if nobody adds a negative comment in the meantime.

PR #334, opened 2017-01-23
[8.0][FIX] Account banking sepa credit transfer: documentation and translation updates

thomaspaulb:
I installed this module and encountered a silly spelling error in Dutch, as well as was looking for half an hour for the source of a certain "Compute error". Have fixed the spelling error and added documentation to help next users on their way.

### DIFF
--- a/account_banking_sepa_credit_transfer/README.rst
+++ b/account_banking_sepa_credit_transfer/README.rst
@@ -32,7 +32,9 @@ This module is part of the OCA/bank-payment suite.
 Configuration
 =============
 
-No specific configuration.
+* You need to configure a valid BIC code on the banks belonging to the 
+source and destination bank accounts, or the export will throw a 
+not-so-informative 'Compute error'.
 
 Usage
 =====
--- a/account_banking_pain_base/models/banking_export_pain.py
+++ b/account_banking_pain_base/models/banking_export_pain.py
@@ -264,26 +264,30 @@ class BankingExportPain(models.AbstractModel):
             self, parent_node, party_type, party_type_label,
             order, party_name, iban, bic, eval_ctx, gen_args):
         """Generate the piece of the XML file corresponding to BIC
-        This code is mutualized between TRF and DD"""
+        This code is mutualized between TRF and DD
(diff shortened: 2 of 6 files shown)


pedrobaeza (approved):

account_banking_sepa_credit_transfer/README.rst
```diff
@@ -32,7 +32,9 @@ This module is part of the OCA/bank-payment suite.
 Configuration
 =============
 
-No specific configuration.
+* You need to configure a valid BIC code on the banks belonging to the 
+source and destination bank accounts, or the export will throw a 
+not-so-informative 'Compute error'.
```
alexis-via, line 38: If there is such an error, we should fix it at the source of the problem instead of describing it in the README.
For a SEPA credit transfer, we do need BIC on source bank account but we don't need it any more on destination bank account, even for cross-borders wire transfer... and I think it works fine in Odoo.

thomaspaulb: I think you are right. 

After some investigation I have seen that from version `9.0` on, the whole of this module and the modules it depends on, have been refactored. Probably the error lies in this function ([8.0](https://github.com/OCA/bank-payment/blob/8.0/account_banking_pain_base/models/banking_export_pain.py#L304) vs [9.0](https://github.com/OCA/bank-payment/blob/9.0/account_banking_pain_base/models/account_payment_order.py#L361)) and more specifically [here](https://github.com/OCA/bank-payment/blob/8.0/account_banking_pain_base/models/banking_export_pain.py#L277). And I think it's already fixed from 9.0 onwards.

So I think the best solution would just be to backport the whole of it to 8.0, but that is a lot of effort. Alternatively, we could do with this README update, or a small bugfix in the 8.0 module. Let me see if I can do the latter quickly.

thomaspaulb: Well, it was not that quickly, but at least I added support for the BIC being omitted (inspired by the 9.0 code) and wrote some tests. The real effort would be to backport everything, but at least this feels better than just adding a README comment.

PR #335, opened 2017-01-24
[FIX] fill date instead off maturity date on account move line generated

vrenaville:


fclementic2c:
:+1: 
The initial issue was : When paying with the fixed date option, the payment journal entries date generated by the payment order must be the fixed date defined (currently behaviour : the entry date = creation date and fixed date goes to due date).

mdietrichc2c (approved):

pedrobaeza (changes requested):
For other use cases, the useful date is the maturity date. For example, you are registering a check for 3 invoices with a specific maturity date different from the invoices itself. We are using that fixed date for specifying the maturity date.

So for me, the solution should be:

* If your transfer account is the bank account, put that date as the accounting date.
* If your transfer account is another, put that date as maturity date.

fclementic2c:
@pedrobaeza What is the point of having a due date on a payment account move line ? I mean, there is no aged balance reports based on these lines so how would you see or use this date ? (any wizard, reports or view ? I see filters but...

pedrobaeza:
In Spain, there are complex flows of moving a balance across accounts to see their "travel" around the states: check received, check in the bank, paid... As stated, all of these states can have different due dates, and I want to have in the resulting move line the following due date to track their aging.

fclementic2c:
@pedrobaeza Ok, if you work with the enterprise version you have the reconciliation report giving you this information I think. We need that report in the community version at some point.
Anyway, ok we will change our PR this way :  

* If  transfer account is the bank account, fixed date = accounting date.
* If  transfer account is another -> as current behaviour.

vrenaville:
@pedrobaeza @fclementic2c 
Thanks for the review
I have modified it according to your comments.

guewen (approved):

vrenaville:
@pedrobaeza Could please check my improvement ?
Thanks for your help

yvaucher (approved):

PR #339, opened 2017-02-13
[IMP] account_payment_partner: Fill payment mode in invoices if none is provided

pedrobaeza:
Using same method as in upstream, payment mode is filled on invoice creation if no payment method is provided. This way, we don't need to install _account_payment_sale_ if we don't want to handle several payment modes at sales level. Even more, if we install the module later and we have already existing
sales orders without payment mode filled, those orders will be invoiced with the customer payment mode.

### DIFF
--- a/account_banking_mandate/__openerp__.py
+++ b/account_banking_mandate/__openerp__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Account Banking Mandate',
     'summary': 'Banking mandates',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.1.0',
     'license': 'AGPL-3',
     'author': "Compassion CH, "
               "Tecnativa, "
--- a/account_banking_mandate/models/account_invoice.py
+++ b/account_banking_mandate/models/account_invoice.py
@@ -29,6 +29,24 @@ class AccountInvoice(models.Model):
                 res['mandate_id'] = invoice.mandate_id.id or False
         return res
 
+    @api.model
+    def create(self, vals):
+        """Fill the mandate_id from the partner if none is provided on
+        creation, using same method as upstream."""
+        onchanges = {
+            '_onchange_partner_id': ['mandate_id'],
+        }
+        for onchange_method, changed_fields in onchanges.items():
+            if any(f not in vals for f in changed_fields):
+                invoice = self.new(vals)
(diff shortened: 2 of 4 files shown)


pedrobaeza:
@sbidoul @alexis-via please review

cc @Tecnativa

sbidoul:
I don't understand the goal of this PR. This looks like this could have surprising side effects. What if, for instance, one wants to create an invoice without payment mode?

pedrobaeza:
The goal is explained in the first comment. Please read it again.

What you say is very simple to achieve: just include 'payment_mode_id': False in your vals. As stated, Odoo makes the same with acount_id, payment_term_id, and so on. There's no possibility of side effects, as you only recover the specific field (payment_mode_id). Not others.

pedrobaeza:
I have added the same logic for filling automatically the mandate.

@sbidoul do you have clear my purpose? Any other doubt?

luismontalba (approved):
Tested on runbot

pedrobaeza:
@sbidoul @alexis-via can you please check again?

sbidoul:
I see that it is indeed done this way [upstream](https://github.com/odoo/odoo/blob/963c9ffa4ae554df0ea0220e31a3b67dd4a8d0bb/addons/account/models/account_invoice.py#L329).

What a ugly hack... I hope it will not propagate everywhere there is a create method.

Since this changes the behaviour, can you bump the version number?

pedrobaeza:
Done, thanks.

sbidoul:
@pedrobaeza I meant the minor version (this is an api change, as the behaviour of create() changes).

sbidoul (approved):

pedrobaeza:
Well, it's not an API change (arguments continue the same), nor it requires a module update. That's why I only changed the revision number.

May you proceed with the merge?

sbidoul:
@pedrobaeza if the behaviour of a method changes (other than bugfixes of course), even if the signature remains the same, then it's an API change.

chienandalu (approved):
tested on runbot

pedrobaeza:
@sbidoul version number changed to minor. Please merge.

pedrobaeza:
Thanks!

pedrobaeza:
Cherry-picking on v10

PR #340, opened 2017-02-14
[IMP] use sql in workflow function for performance

hbrunn:
the problem with using the ORM in workflow functions is that we have no caching whatsoever, so as soon as we have a couple of sent payment orders and a lot of move lines to reconcile, triggering the workflow will cost a few seconds, which is too much when reconciling hundreds of lines


pedrobaeza (approved):

pedrobaeza:
@alexis-via @sbidoul can you check? Although I'm not a big fan of direct SQL, in this case the gain is enough to justify it.

alexis-via (approved):

PR #342, opened 2017-02-23
[FIX]date instead off maturity date on account move line generate

vrenaville:


vrenaville:
Please see the PR releated to V10:

https://github.com/OCA/bank-payment/pull/335

yvaucher (approved):

yvaucher:
As this is a backward port of fix #335 thus I merge it

PR #345, opened 2017-03-02
[9.0][IMP] account_payment_partner: Allow hide bank account in report invoice

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/account_payment_partner/views/account_payment_mode.xml
+++ b/account_payment_partner/views/account_payment_mode.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<odoo>
+<data>
+
+    <record id="account_payment_mode_form" model="ir.ui.view">
+        <field name="model">account.payment.mode</field>
+        <field name="inherit_id"
+               ref="account_payment_mode.account_payment_mode_form" />
+        <field name="arch" type="xml">
+            <group name="main" position="inside">
+                <field name="bank_invoice_report"/>
+            </group>
+        </field>
+    </record>
+
+</data>
+</odoo>
--- a/account_payment_partner/__openerp__.py
+++ b/account_payment_partner/__openerp__.py
@@ -18,6 +18,7 @@
         'views/res_partner_view.xml',
         'views/account_invoice_view.xml',
         'views/account_move_line.xml',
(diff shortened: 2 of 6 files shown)


rafaelbn (approved):
Tested in runbot

pedrobaeza:
IMO this should be done another way:

* In payment mode, don't require to put a bank account, via a third option (fixed, multiple or none), or directly making the field not required.
* I think this is already working for multiple bank accounts: in partner onchange. don't put any bank account on this case

account_payment_partner/views/account_payment_mode.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<odoo>
+<data>
```
rigo1985, line 8: In version 10 "< data >" is only necessary when you use "noupdate". 
(For example: < data noupdate = "1" > )
You can omit it in this case.

account_payment_partner/views/account_payment_mode.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
```
rigo1985, line 2: The copyright I think is not necessary in this file.

luismontalba:
How can I chose this option? How can I hide the bank account in invoice report?

pedrobaeza:
@sergio-teruel please tell me about my comments

alexis-via:
I'm not sure to understand the generic need behind PR. Maybe the solution would be to change the onchange so that the field "partner_bank_id" stays empty on the invoice if the field 'bank_account_required' of account.payment.method is not true (or something like this....)

pedrobaeza:
@alexis-via The idea here is to decide at payment mode level if the bank account in the invoice should be printed or not in the invoice report. There are a lot of cases when you don't want to print this bank account, and put other explanations in the payment mode description itself.

pedrobaeza:
@sergio-teruel is this still needed for our projects or solution in v10 is enough?

sergio-teruel:
There is not changes in v10, so it is needed, @rafaelbn, We need this?

rafaelbn:
Please this is ready yo merge!!!  :+1: Could we merge? This is in production for long time...



ping @carlosdauden  @cubells @pedrobaeza

rafaelbn (approved):
:+1:

carlosdauden (approved):

PR #351, opened 2017-03-27
[FIX] we need to decorate the onchange handler with @api.multi

hbrunn:
...otherwise, `website_sale` will choke because it calls the onchange handler with v7 api: https://github.com/OCA/OCB/blob/9.0/addons/website_sale/controllers/main.py#L634


alexis-via (approved):

pedrobaeza (approved):
I want to go ASAP to v10 to avoid this kind of things.

Issue #353, opened 2017-04-03
[10.0] Crash in account_payment_order


alexis-via:
When you create a payment order with a payment mode that has a variable link to a bank account, you get this crash:

```
Traceback (most recent call last):
  File "/home/odoo/erp/odoo/odoo/http.py", line 638, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/odoo/erp/odoo/odoo/http.py", line 675, in dispatch
    result = self._call_function(**self.params)
  File "/home/odoo/erp/odoo/odoo/http.py", line 331, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/odoo/erp/odoo/odoo/service/model.py", line 119, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/odoo/erp/odoo/odoo/http.py", line 324, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo/erp/odoo/odoo/http.py", line 933, in __call__
    return self.method(*args, **kw)
  File "/home/odoo/erp/odoo/odoo/http.py", line 504, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/erp/odoo/addons/web/controllers/main.py", line 878, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/erp/odoo/addons/web/controllers/main.py", line 870, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/odoo/erp/odoo/odoo/api.py", line 679, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/home/odoo/erp/odoo/odoo/api.py", line 664, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/home/odoo/erp/usability/account_usability/account.py", line 112, in name_search
    name=name, args=args, operator=operator, limit=limit)
  File "/home/odoo/erp/odoo/odoo/models.py", line 1591, in name_search
    return self._name_search(name, args, operator, limit=limit)
  File "/home/odoo/erp/odoo/odoo/models.py", line 1604, in _name_search
    ids = self._search(args, limit=limit, access_rights_uid=access_rights_uid)
  File "/home/odoo/erp/odoo/odoo/models.py", line 4205, in _search
    query = self._where_calc(args)
  File "/home/odoo/erp/odoo/odoo/models.py", line 4006, in _where_calc
    where_clause, where_params = e.to_sql()
[...]

PR #354, opened 2017-04-04
[FIX] Bug #353 crash when selecting a payment mode that has a variable link to bank accounts

alexis-via:
Cf bug #353 for a detailed scenario of the crash.

I you have better ideas to fix this bug, please propose. I suggests this solution because it's the only one I found that currently works on v10.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -39,9 +39,6 @@ class AccountPaymentOrder(models.Model):
         'account.journal', string='Bank Journal', ondelete='restrict',
         readonly=True, states={'draft': [('readonly', False)]},
         track_visibility='onchange')
-    allowed_journal_ids = fields.Many2many(
-        'account.journal', compute='_compute_allowed_journals', readonly=True,
-        string='Selectable Bank Journals')
     # The journal_id field is only required at confirm step, to
     # allow auto-creation of payment order from invoice
     company_partner_bank_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


account_payment_order/models/account_payment_order.py
```diff
@@ -162,10 +146,18 @@ def create(self, vals):
     @api.onchange('payment_mode_id')
     def payment_mode_id_change(self):
         journal_id = False
+        res = {'domain': {
+            'journal_id': "[('id', '=', False)]",
```
pedrobaeza, line 150: This should be `{'journal_id': []}`

pedrobaeza (changes requested):

sbidoul:
@alexis-via I think the onchange will not trigger when opening a previously saved payment order, so in that case, the domain will be wrong.

Dynamic domains are a pain to get right...

I think it's part of the reasons that pushed @lmignon to create https://github.com/OCA/web/pull/567. You may want to try it.

alexis-via:
@sbidoul Very interesting, thanks !
Another interesting bug report https://github.com/odoo/odoo/issues/16072

account_payment_order/models/account_payment_order.py
alexis-via: If I make your modification, then I have access to all journals when no payment mode is selected, which is not what I want : I want the user to select a payement mode first. So I prefer my solution, which works well.

alexis-via:
I propose to merge this to avoid the crash. And then, if someone is able to propose a better solution, he will propose another PR.
I don't want to leave an horrible crash like this one too long in an important OCA module.

sbidoul (approved):

account_payment_order/models/account_payment_order.py
pedrobaeza: Then please hide the field when no payment mode is selected in the view via `attrs` attribute.

PR #358, opened 2017-04-13
[9.0] New module account_payment_expense to copy bank details from exployee to account move lines of expense

ghost:


account_payment_expense/README.rst
```diff
@@ -0,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
```
elicoidal: @BT-hstoeckle   Thanks for your contribution to the OCA. 
Let me introduce to you some basic information about contribution in OCA and I will review other points in more comments inside the code..
In general, please follow the principles: 
* Contribution guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
* Please follow the OCA README template:
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
* We recommend to include a screenshot as well to help new comers (nice to have)
* Please use simple headers in all py files following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module
https://github.com/BT-hstoeckle/__openerp__.py#L3

elicoidal: In particular Title line must be at least as long as the text

account_payment_expense/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
elicoidal: Please use standard OCA short headers.

elicoidal (changes requested):
Please review the licensing.
Some UT would be appreciated as well

BT-astauder:
:+1:

alexis-via:
Very good initiative. My ideas to improve it:
- copy the payment_mode_id of the partner to the move line too
- in the inherit of action_move_create(), you could write only on the move lines where account_id == expense.employee_id.address_home_id.property_account_payable_id.id (for better perf)
- add a button "Add to payment order" on the expense, just like I did it for the invoice (cf the module account_payment_order https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/views/account_invoice_view.xml#L15 and https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/models/account_invoice.py#L39

But it is already good as-is :+1:

account_payment_expense/models/hr_expense.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
+
+    @api.multi
+    def action_move_create(self):
+        result = super(HrExpense, self).action_move_create()
+
+        for line in self.account_move_id.line_ids:
```
BT-astauder: Does not work with multiple expenses.

PR #361, opened 2017-05-04
[8.0] account_banking_payment_export dot in bank lines view

drl9319:
Hi
I have a problem with the bank_line view because the decimal and thousands separator are incorrect, can view this image:
bank-lines-error
But with this new code view fine this view like this image:
bank-lines-ok

### DIFF
--- a/account_banking_payment_export/README.rst
+++ b/account_banking_payment_export/README.rst
@@ -18,7 +18,7 @@ for electronic banking. It provides the following technical features:
   payment.mode.type
 * a manual payment mode type is provided as an example, with a default "do
   nothing" wizard
-  
+
 To enable the use of payment order to collect money for customers,
 it adds a payment_order_type (payment|debit) as a basis of direct debit support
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is not the solution. Check why the field is not showing correctly the currency.

rubencabrera:
> This is not the solution. Check why the field is not showing correctly the currency.

Would [this](https://github.com/OCA/bank-payment/blob/8.0/account_banking_payment_export/models/bank_payment_line.py#L20) be fixed by OCA/OCB#461 ?
There's also a pr at odoo/odoo#11505, but pending for almost a year.

pedrobaeza:
No, it's only a problem of visualization of the widget. It should have the `€` symbol.

alexis-via:
As @rubencabrera said, it is very probably fixed by holger's PR https://github.com/OCA/OCB/pull/461

@drl9319 Could you try with a recent OCA/OCB v8 instead of official odoo v8 ?

PR #362, opened 2017-05-11
Rearrange the order of partner block to fit the XML schema of pain.001.001.02

gmarczynski:
So far the code was correct only for pain.001.001.03 (used by default). 
In pain.001.001.03 element Ctry goes before AdrLine in PostalAddress1.
In pain.001.001.02 the order is reversed.

This PR reverses the order of Ctry  and AdrLine for pain.001.001.02 and makes it pass the schema check.

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -358,6 +358,14 @@ class AccountPaymentOrder(models.Model):
         return True
 
     @api.model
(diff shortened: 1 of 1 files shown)


alexis-via:
generate_coutry_block -> generate_country_block

alexis-via:
I confirm the problem and I confirm that the solution is good. Once my remarks are integrated, we can merge.

PR #363, opened 2017-05-16
[10.0] Add 'post_move' option on account.payment.mode

alexis-via:
If you use account_cancel, it's not a problem to auto-post account moves generated upon final validation of the payment order, but for those who don't/can't use account_cancel, it's not a good idea to auto-post account moves.

### DIFF
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Account Payment Order',
-    'version': '10.0.1.1.1',
+    'version': '10.0.1.1.2',
     'license': 'AGPL-3',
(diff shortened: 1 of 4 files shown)


alexis-via:
And posting account moves was not the default option in the banking modules for v8: https://github.com/OCA/bank-payment/blob/8.0/account_banking_payment_transfer/model/account_payment.py#L258
Now, we have the choice in v10 too (the option "entry_posted" doesn't exist any more on the journal).

sbidoul (approved):
+1

pedrobaeza (changes requested):
Several things:

* Change module version number
* Put default=True for not changing behavior on a stable version.

pedrobaeza (approved):

PR #364, opened 2017-05-22
[10.0] Add field default_date_prefered on payment mode

alexis-via:
This new field "default_date_prefered" on payment mode allows to set the default value of the field "date_prefered" on payment orders create with that payment mode.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -141,6 +141,10 @@ class AccountPaymentOrder(models.Model):
             vals['payment_type'] = payment_mode.payment_type
             if payment_mode.bank_account_link == 'fixed':
                 vals['journal_id'] = payment_mode.fixed_journal_id.id
+            if (
+                    'date_prefered' not in vals and
+                    payment_mode.default_date_prefered):
+                vals['date_prefered'] = payment_mode.default_date_prefered
         return super(AccountPaymentOrder, self).create(vals)
 
     @api.onchange('payment_mode_id')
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Bump version number

adrienpeiffer (approved):
:+1:

account_payment_order/models/account_payment_order.py
```diff
@@ -141,6 +141,10 @@ def create(self, vals):
             vals['payment_type'] = payment_mode.payment_type
             if payment_mode.bank_account_link == 'fixed':
                 vals['journal_id'] = payment_mode.fixed_journal_id.id
+            if (
+                    'date_prefered' not in vals and
```
espo-tony: The first term of the condition could be written as:
`not vals.get('date_prefered', None) and`

It's true that on a create, usually invoked saving a new record by user interface, fields set to False / None are not even passed as arguments (and anyway, `date_prefered` is defined as required and with a default, so when invoked by user interface it will never happen to not find it in the arguments). Still, if the create is invoked differently, like from inside another method or through an external XML-RPC request, it could easily happen that `'date_prefered'` is in `vals` but it's set to _None / False_. The suggested change would correctly address this case.

espo-tony (changes requested):

espo-tony (approved):
:+1: Thanks!

alexis-via:
As discussed at the banking/accounting table, we think it's not a good idea to update version numbers in PRs because it will create systematic conflicts, which is a pain to solve and slows down a lot the merge of the PR. The update of version number should probably be done by the guy who merges the PR...

PR #365, opened 2017-05-29
[10.0] Explicit error when bank account is missing on bank journal

alexis-via:
Without this PR, when the bank account is not set on the bank journal, you get an error when generating the SEPA XML file which says "Cannot compute 'Company Name'"... and it's very difficult to understand the root cause of the problem. I had this error this morning during my test and that's why I made this PR.

PR from the Barcelona code sprint !

### DIFF
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Account Payment Order',
-    'version': '10.0.1.1.2',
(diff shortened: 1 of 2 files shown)


sbidoul (approved):
code review

adrienpeiffer (approved):

espo-tony (approved):
:+1:

pedrobaeza (changes requested):
Please change version number

PR #373, opened 2017-07-04
[9.0][FIX] account_banking_mandate: Remove mandate view from partner bank_ids field

sergio-teruel:
Migration scripts

cc @Tecnativa


pedrobaeza:
This is not needed. I have already migrated a lot of DBs and haven't found any problem.

PR #374, opened 2017-07-10
[10.0]Fix payment method onchange with multicompany

grindtildeath:
Current behaviour :
 - When selecting a payment method, all the journals matching payment_method_id.payment_type are selected.

Desired behaviour :
 - When selecting a payment method, as the payment mode has a company_id required, we only want the journals matching payment_method_id.payment_type and the self.company_id.

### DIFF
--- a/account_payment_mode/__manifest__.py
+++ b/account_payment_mode/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Account Payment Mode',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'license': 'AGPL-3',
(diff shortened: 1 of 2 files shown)


adrienpeiffer (approved):
:+1:

alexis-via:
The multi-company record rule on account.journal should naturally filter-out journals from other companies, no ? But it seems like a good modification anyhow.

MiquelRForgeFlow (approved):

pedrobaeza (changes requested):
Please change version number

pedrobaeza (approved):

PR #376, opened 2017-07-26
[9.0] [account_payment_order] Don't allow deletion of an uploaded payment order + allow to cancel an uploaded payment order

mdietrichc2c:


pedrobaeza (changes requested):
Very useful addition! Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

grindtildeath:
@mdietrichc2c Added a few improvements here : https://github.com/mdietrichc2c/bank-payment/pull/1

Most important is also to restrict deletion of bank.payment.line should it be on a paymernt order which is uploaded.

pedrobaeza:
Why restricting on the bank.payment.line model? You don't have direct access via UI to remove that lines, so there's no the possibility to delete them unless you program specific code for that, which has to be then control on that code.

grindtildeath:
@pedrobaeza There is a menu item in this module to access bank.payment.line : https://github.com/OCA/bank-payment/blob/9.0/account_payment_order/views/bank_payment_line.xml#L79

pedrobaeza:
Thanks for clarifying. Then it's correct indeed. I would go further and remove directly the possibility to delete items from that list (`<tree delete="0">`), but I don't know if it has any use that I don't know.

grindtildeath:
@pedrobaeza as we need to restrict the deletion only if payment order state = 'uploaded' I suppose we should still allow to delete bank payment line for payment order which aren't yet in this status.

pedrobaeza:
OK

grindtildeath (approved):

pedrobaeza:
Please fix lint issue:
```
account_payment_order/models/bank_payment_line.py:167:33: E126 continuation line over-indented for hanging indent
```
and it would be good to have tests for this new code.

grindtildeath:
@mdietrichc2c Please merge https://github.com/mdietrichc2c/bank-payment/pull/2

pedrobaeza:
Superseeded by #377 with commit squashing + a couple of errors I have found. If everything goes right, I'll merge and forward port to 10.0.

PR #377, opened 2017-08-02
[9.0] account_payment_order: Don't allow deletion of an uploaded payment order + allow to cancel an uploaded payment order

pedrobaeza:
Replaces #376 as there's no possibility of uploading changes in it (and the commit history in the other PR was a bit dirty).

I have fixed also 2 little things (a return and the import of the tests), so if everything goes right, I will merge directly, assuming the previous approval + my approval.

### DIFF
--- a/account_payment_order/__openerp__.py
+++ b/account_payment_order/__openerp__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Account Payment Order',
-    'version': '9.0.1.1.2',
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Forward ported on d7e9eaaf35ec9b0fa978921f45c3e8dd45bc396e and 7addb7157ab31416d11d51fd680052e80352a092

PR #378, opened 2017-08-02
[9.0][IMP] account_banking_mandate: Improve / fixes

carlosdauden:
Adds a compute field in res.partner to search first valid mandate respecting partner banks order.
Fix payment_mode_id_change to set valid mandate if required
 
@Tecnativa

### DIFF
--- a/account_banking_mandate/models/res_partner.py
+++ b/account_banking_mandate/models/res_partner.py
@@ -15,7 +15,7 @@ class ResPartner(models.Model):
     valid_mandate_id = fields.Many2one(
         comodel_name='account.banking.mandate',
         compute='compute_valid_mandate_id',
-        string='Valid Mandate')
+        string='First Valid Mandate')
 
     @api.multi
     def _compute_mandate_count(self):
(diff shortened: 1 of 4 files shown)


account_banking_mandate/models/res_partner.py
```diff
@@ -11,6 +12,10 @@ class ResPartner(models.Model):
     mandate_count = fields.Integer(
         compute='_compute_mandate_count', string="Number of Mandates",
         readonly=True)
+    valid_mandate_id = fields.Many2one(
+        comodel_name='account.banking.mandate',
+        compute='compute_valid_mandate_id',
+        string='Valid Mandate')
```
pedrobaeza: First Valid Mandate

account_banking_mandate/models/res_partner.py
```diff
@@ -21,3 +26,17 @@ def _compute_mandate_count(self):
             for mandate in mandate_data])
         for partner in self:
             partner.mandate_count = mapped_data.get(partner.id, 0)
+
+    @api.multi
+    def compute_valid_mandate_id(self):
+        # Dict to reduce impact with "bug" that process all partners related
```
pedrobaeza: `Dict for reducing the duplicated searches on parent/child partners`

account_banking_mandate/models/res_partner.py
```diff
@@ -21,3 +26,17 @@ def _compute_mandate_count(self):
...
+        mandates_dic = {}
+        for partner in self:
+            commercial_partner_id = partner.commercial_partner_id
```
pedrobaeza: Name it `commercial_partner`

account_banking_mandate/models/res_partner.py
```diff
@@ -21,3 +26,17 @@ def _compute_mandate_count(self):
...
+            if commercial_partner_id in mandates_dic:
+                partner.valid_mandate_id = mandates_dic[commercial_partner_id]
+            else:
+                mandate_id = partner.commercial_partner_id.bank_ids.mapped(
```
pedrobaeza: This  is incorrect if you have more than one bank account or more than one valid mandate. It should be:
```
mandate_id = partner.commercial_partner_id.mapped('bank_ids.mandate_ids')..filtered(lambda x: x.state == 'valid')[:1].id
```

pedrobaeza (changes requested):
We need to dig deeper in the onchange tough

rafaelbn:
Hi @pedrobaeza , @carlosdauden attend you comments. Please review again. Thanks

pedrobaeza:
@carlosdauden is this still relevant?

PR #381, opened 2017-08-18
[10.0][IMP] account_banking*: Show identifiers at payment mode level

pedrobaeza:
There's not too much sense to have a mini localization module that shows these fields as they are totally optional. Even more, if you have a multi-company DB with several localizations, including one that needs these fields, then it will be shown in the rest of the companies too.

@alexis-via please accept this. They were shown on 8.0, and you made this trick for hiding them on v9. It's not fair to require an extra l10n_es_* module for showing them.

cc @sbidoul

### DIFF
--- a/account_banking_pain_base/README.rst
+++ b/account_banking_pain_base/README.rst
@@ -25,7 +25,18 @@ This module is part of the OCA/bank-payment suite.
 Configuration
 =============
 
-No configuration required.
+#. Go to Accounting > Configuration > Settings.
+#. On the fields "Initiating Party Issuer" and "Initiating Party Identifier",
+   in the section *SEPA/PAIN*, you can fill the corresponding identifiers.
+
+If your country requires several identifiers (like Spain), you must:
+
+#. Go to *Accounting > Configuration > Settings*.
+#. On the section *SEPA/PAIN*, check the mark "Multiple identifiers".
+#. Now go to *Accounting > Configuration > Management > Payment Modes*.
+#. Create a payment mode for your specific bank.
+#. Fill the specific identifiers on the fields "Initiating Party Identifier"
+   and "Initiating Party Issuer".
 
 Usage
 =====
@@ -34,7 +45,7 @@ See 'readme' files of the OCA/bank-payment suite.
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/173/9.0
+   :target: https://runbot.odoo-community.org/runbot/173/10.0
 
 Known issues / Roadmap
 ======================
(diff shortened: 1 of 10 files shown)


alexis-via:
I think it's not fair to display fields such as ICS at payment mode level in the generic module because this non-sense of having several ICS for one company only happens in Spain. If we display it at payment mode level in the generic modules, it will complexify the user interface for all users including non-spanish users for a problem that only happens in Spain.

Developping 2 auto-install modules in the spanish localisation to display those invisible fields is a very easy solution.

You have my opinion now. Let's see what other users think...

pedrobaeza:
@alexis-via, there's other specific localization code already in the repository: https://github.com/OCA/bank-payment/blob/08c9725f2a4c4b899160ce63a02ab34a7d3114db/account_banking_pain_base/models/res_company.py#L26-L52

So it's not too much to ask to include this also. And again, this was in 8.0 and it was removed by you.

oihane (approved):
LGTM

sbidoul:
Perhaps we could add a label text in the view saying to set these fields only if the company value must be changed per payment mode?

pedrobaeza:
The help already says that, but if it helps you to decide, I can do it.

anajuaristi:
:+1:

rafaelbn (approved):
Functionally tested in production environment, credit and debit done successfully.

alexis-via:
@pedrobaeza The problem for me is not to have localisation code in the repo ; in fact, the code to handle identifiers at payment mode level IS present in the repo (and I am the one who put it in the repo for v9 and v10, because it was easier to have the code there). What I don't want is to complexify the interface for all users... for a use-case which only happens in Spain. I really don't want that. Adding a module l10n_es_ that displays the field is the right solution for me.

pedrobaeza:
Again, it's something that it was already, so we are only restoring it. I don't want also the default filter configuration for each payment mode (and they are a lot), but as they are optional, I'm not bothering you about that. Please do the same with these fields.

sbidoul:
@alexis-via would it work if these field are in a kind of "advanced" section of the view, with a label saying "Advanced: use this only to override company-level settings."?

alexis-via:
Another solution would be to add a group "Identifiers on Payment Modes". Only Spanish people will add themselves in this group, to be able to see these fields on payment modes ; only users will not be affected by the additionnal complexity.

rafaelbn:
@alexis-via when you say "Spanish people" I'm curious. If your company works with 3 different banks, you have a single identifier for both credit and debit in the 3 ones? I cannot believe this is a problem only for Spanish banks. We are talking about ***SEPA***! 🤔 
What about Portugal? France? Greece? Germany? Belgium? .... Are you sure that this is a use-case which only happens in Spain?

I mean that ***Deutsche Bank*** in Spain is giving a different identifier that Banco Santander in Spain so...

pedrobaeza:
Any way, the group solution proposed is good for me, so I'll implement it soon and put it here.

rafaelbn:
@dreispt could you please illustrate here a little how it is in Portugal please? https://github.com/OCA/bank-payment/pull/381#issuecomment-324373392

dreispt:
We don't hold any specific attribute for Payment Accounts, other than the IBAN and BIC.
I see an `<Id>` element in the SEPA header, and it is being fileld with the VAT/TIN number (it's the same thing here).

image

alexis-via:
@rafaelbn I do confirm that a French company will only have a single ICS, even if it has 3 bank accounts at 3 different banks to makes SEPA direct debits. An ICS is like a VAT number, you have 1 per company. Simple, no ? :)

alexis-via:
@pedrobaeza I'm very happy that we have found a solution that is simple and OK for both of us !

pedrobaeza:
I have added the group logic according comments and also improve documentation. Please review.

alexis-via (approved):

pedrobaeza:
Thank you very much!

PR #382, opened 2017-08-22
[FIX][10.0] account_banking_mandate: onchange return value missing

astirpe:
This PR fixes the following issue:

**Scenario**
On a fresh database install module "`account`". Do not install "`account_banking_mandate`" yet.
Enable group "`A warning can be set on a partner (Account)`" for the user.
Open the Customer form and display the Internal Notes tab.
Set a blocking message for this customer.
Create a new invoice, set the same Customer as above, verify that the blocking message is shown.

**Bug**
Install module "`account_banking_mandate`".
Create a new invoice, set the same Customer as above, verify that the blocking message is NOT shown anymore.

**Proposal**
Actually the method `def _onchange_partner_id()` calls the super without propagating its returned value (that returned value contains the warning message to be displayed). With this PR, the onchange method propagates the return value got from the super. As a result, the blocking message will be shown correctly.

### DIFF
--- a/account_banking_mandate/__manifest__.py
+++ b/account_banking_mandate/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Account Banking Mandate',
     'summary': 'Banking mandates',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.1.1',
     'license': 'AGPL-3',
     'author': "Compassion CH, "
               "Tecnativa, "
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
Change module version according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

pedrobaeza (approved):

MiquelRForgeFlow (approved):

pedrobaeza:
We can fast-track this obvious patch.

PR #389, opened 2017-09-05
[10.0][account_banking_mandate] if a mandate is linked to the bank account used, put it automatically on the payment line

Cedric-Pigeon:
In 8.0, if a mandate was linked to a banking account, when creating a payment line with this account, the mandate was automatically set: https://github.com/OCA/bank-payment/blob/8.0/account_banking_mandate/models/payment_line.py#L36

Unfortunately this behavior was lost during migration process. 

This PR intends to re-introduce it.

Thanks to @jbeficent for its help.

### DIFF
--- a/account_banking_mandate/README.rst
+++ b/account_banking_mandate/README.rst
@@ -60,6 +60,7 @@ Contributors
 * Alexandre Fayolle
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
 * Sergio Teruel (Incaser) <sergio@incaser.es>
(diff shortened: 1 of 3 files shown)


Cedric-Pigeon:
@alexis-via What do think of that ?

adrienpeiffer (approved):
:+1: (tested locally)

ThomasBinsfeld (approved):
Code review :+1:

rousseldenis (approved):
Code review

PR #392, opened 2017-09-12
[V9.0][BACK_PORT] [FIX] Bug #353 crash when selecting a payment mode

vrenaville:
Hello,

It's a back port of this fix : 
https://github.com/OCA/bank-payment/pull/354


pedrobaeza:
Please take into account my comments from the other PR. They were ignored there, but they are still valid, as also pointed later that the dynamic domain is not triggered on already existing orders.

philhug:
please merge this PR as it causes an exception. I'm affected by this as well since I updated the modules.
the ticket can be kept open to track the remaining issues.

PR #393, opened 2017-09-13
[IMP] account_payment_*: Complete migration scripts

pedrobaeza:
* Put payment modes with bank link = "fixed"
* Populate payment methods from old payment mode types
* Link for avoiding duplicates old types with new methods

cc @Tecnativa

### DIFF
--- a/account_payment_order/migrations/9.0.1.0.0/pre-migration.py
+++ b/account_payment_order/migrations/9.0.1.0.0/pre-migration.py
@@ -59,6 +59,25 @@ column_renames_payment_transfer = {
     ],
 }
 
+# They are being handled after module rename, so source module name is correct
+EXISTING_PAYMENT_MODE_TYPE_XML_IDS = {
+    'account_payment_order.manual_bank_tranfer': (
+        'account.account_payment_method_manual_out',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
cc @hbrunn @sergio-teruel

account_payment_order/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -59,6 +59,25 @@
     ],
 }
 
+# They are being handled after module rename, so source module name is correct
```
yajo, line 62: Shouldn't almost all the patch go into some `9.0.1.0.1/pre-migration.py` file so that people with `9.0.1.0.0` can upgrade and get the changes? 🤔

pedrobaeza: No, this is a no-way. If you don't have it before, now this is not going to work properly. So better to just leave it for new migrations. Also openupgradelib dependency for the rest of the people using this module without real migration prevents this.

yajo (approved):

hbrunn (approved):

PR #396, opened 2017-10-06
[FIX][9.0] Migration script

StefanRijnhart:
Model name is invalid, and xml ids map to tuples of length 1 instead of a string because of trailing commas.

### DIFF
--- a/account_payment_order/migrations/9.0.1.0.0/pre-migration.py
+++ b/account_payment_order/migrations/9.0.1.0.0/pre-migration.py
@@ -61,21 +61,17 @@ column_renames_payment_transfer = {
 
 # They are being handled after module rename, so source module name is correct
 EXISTING_PAYMENT_MODE_TYPE_XML_IDS = {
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Uf, I did it in the shell and then move it to a migration script without putting attention to these details derived from linting, sorry.

Fast tracking for fixing it.

StefanRijnhart:
@pedrobaeza your migration script is great work otherwise :+1:

account_payment_order/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -108,7 +104,7 @@ def migrate_payment_mode_types(env):
         # the same goal
         imr = imr_obj.search([
             ('res_id', '=', row[0]),
-            ('model', '=', 'payment_mode_type'),
+            ('model', '=', 'payment.mode.type'),
         ])
```
pedrobaeza, line 108: I'm wondering how this one works in my installation... I'll double check it.

PR #397, opened 2017-10-09
[ADD] Module which allow to cancel an invoice which is linked to a payment order

benwillig:
When you try to cancel an invoice which is linked to a payment order, an integrity error is raised. This module allows to destroy the link between the invoice and the payment order before cancelling the invoice.


account_payment_order_invoice_cancel/wizards/account_payment_order_invoice_cancel.py
```diff
@@ -0,0 +1,41 @@
...
+    def default_get(self, fields_list):
+        res = super(
+            AccountPaymentOrderInvoiceCancel, self).default_get(fields_list)
+        active_ids = self.env.context.get('active_ids')
+        if active_ids:
+            res['invoice_ids'] = [(6, 0, active_ids)]
+        return res
+
+    @api.multi
+    def _delete_payment_order_link(self):
```
adrienpeiffer: this method doesn't seem useful

account_payment_order_invoice_cancel/tests/test_payment_order.py
```diff
@@ -0,0 +1,63 @@
...
+        return invoice
+
+    def test_cancel_payment_order(self):
+        # Open invoice
+        invoice = self._create_supplier_invoice()
+        invoice.action_invoice_open()
+        # Add to payment order
+        invoice.create_account_payment_line()
+
+        cancel_res = invoice.action_cancel()
```
adrienpeiffer: I think it's better to test this line after ` invoice.move_id.journal_id.update_posted = True`

account_payment_order_invoice_cancel/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def action_cancel(self):
+        invoices_with_payment_orders = self.\
+            _get_invoice_linked_with_payment_orders()
+        self = self - invoices_with_payment_orders
+        res = True
+        if self:
+            res = super(AccountInvoice, self).action_cancel()
```
adrienpeiffer, line 19: wouldn't it be better to call the wizard for all invoices if at least one invoice is linked with a payment line ?
If we imagine a process that allows to cancel several invoices, those that are not associated with a line of payment will be automatically canceled but a wizard will open for the others. If you click on the cancel button of this wizard, the others are already cancel. This may be disturbing.
What do you think about that ?

adrienpeiffer (changes requested):

account_payment_order_invoice_cancel/models/account_invoice.py
benwillig: I understand, but the wizard warns about the fact that invoices are linked to a payment order. I think that if we show invoices which are not linked to a payment order, it would be disturbing too.

adrienpeiffer (approved):

pedrobaeza:
The problem I see with this is the consistency. What about if the payment order has been done and has a transfer account?

benwillig:
Indeed, there is a problem. But, for example, when there was an error during invoice encoding (ex: account on lines), there is no way to correct the invoice and it's impossible to cancel the invoice because it's linked to a payment order. I will add a control to allow this action only for invoices which are linked to a done payment order, because in previous states, user is able to correct everything manually.

pedrobaeza:
Is #433 somehow an alternative of this?

PR #402, opened 2017-10-16
[10.0] account_payment_order: allow blocking direct payment on invoices

Cedric-Pigeon:
This PR adds a flag on cash and banks journals.
This flag prevents the usage of the journal through the Register Payment wizard on an invoice.
The default value is False.

### DIFF
--- a/account_payment_order/README.rst
+++ b/account_payment_order/README.rst
@@ -34,6 +34,9 @@ This module also adds a button *Add to Payment Order* on supplier invoices and a
 
 You can print a Payment Order via the menu Accounting > Payments > Payment Orders and then select the payment oder to print.
 
+You can block the usage of a journal through the Register Payment wizard on an invoice by un-checking the flag "Allow Direct Payment"
+in the advanced settings of the journal.
+
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/173/10.0
@@ -72,6 +75,7 @@ Contributors
 * Jose María Alzaga <jose.alzaga@aselcis.com>
 * Meyomesse Gilles <meyomesse.gilles@gmail.com>
 * Carlos Dauden
+* Cédric Pigeon <cedric.pigeon@acsone.eu>
 
 Maintainer
 ----------
--- a/account_payment_order/models/__init__.py
+++ b/account_payment_order/models/__init__.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from . import account
+from . import account_payment
 from . import account_payment_mode
 from . import account_payment_order
 from . import account_payment_line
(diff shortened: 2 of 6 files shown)


account_payment_order/models/account.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    allow_direct_payment = fields.Boolean(
+        help="But checking this option, you allow to use this journal through"
+             " the Register Payment feature directly on an invoice",
+        default=False
```
sbidoul: default=True to preserve the behaviour of existing installations?

Cedric-Pigeon: Ok, changed.

sbidoul:
Could you add a line in the readme to mention this new feature?

Cedric-Pigeon:
@sbidoul Done.

sbidoul (approved):

account_payment_order/README.rst
```diff
@@ -34,6 +34,9 @@ This module also adds a button *Add to Payment Order* on supplier invoices and a
 
 You can print a Payment Order via the menu Accounting > Payments > Payment Orders and then select the payment oder to print.
 
+You can block the usage of a journal through the Register Payment wizard on an invoice by un-checking the flag "Allow Direct Payment"
```
pedrobaeza, line 37: Put how to configure it in the configuration section.

account_payment_order/models/__init__.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from . import account
```
pedrobaeza, line 3: Should be called account_journal

account_payment_order/models/account.py
```diff
@@ -0,0 +1,14 @@
...
+# © 2013-2017 ACSONE SA (<http://acsone.eu>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    allow_direct_payment = fields.Boolean(
+        help="But checking this option, you allow to use this journal through"
```
pedrobaeza, line 11: Remove initial `But` word

pedrobaeza (commented):
I was reviewing the feature, but I see it very unrelated to the module itself. You are inhibiting an standard Odoo feature inside the module for making payment orders... I don't get the relation

sbidoul:
@pedrobaeza the reasoning to put it in this module is as follow. 

It's because they use payment orders that users want to disable the register payment feature of Odoo on the journals they use for payment orders. If they don't use the payment orders module, then they use the standard feature for payment so there is no need to disable payment on some journals.

Since we did not see use case for that feature without payment orders, I advised Cedric to put this feature in directly in the payment order module, .

pedrobaeza:
OK, you can include this use case in the main section of the README for avoiding further questions in the future

adrienpeiffer (approved):
:+1:

AaronHForgeFlow:
It seems to me that the payment order module is duplicating part of the logic of the account payment. From my point of view once a payment order is confirmed it's also a payment.

PR #401, opened 2017-10-11
[11.0][MIG] account_payment_order

etobella:
Migration to 11.0
Still working on tests and multicompany environments

Depends on

- [x] account_payment_mode #399 
- [x] account_payment_partner #400

PR #400, opened 2017-10-11
[11.0] [MIG] account_payment_partner

etobella:
Migration to 11.0.

Depends on:
- [x] account_payment_mode #399 

Still working on tests and multicompany environments

PR #399, opened 2017-10-11
[11.0] [MIG] account_payment_mode

etobella:
Migration to 11.0

PR #404, opened 2017-10-16
[11.0] [MIG] account_payment_mandate

etobella:
Migration to 11.0.
Some tests were added
Depends on:
 - [x] account_payment_mode #399
 - [x] account_payment_partner #400
 - [x] account_payment_order #401


pedrobaeza:
One thing pending in the roadmap is to add in this generic module a field with type selector that includes here a "Generic mandate", and account_sepa_direct_debit simply adds a value instead of overwriting all fields and menus from this generic one.

account_banking_mandate/README.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Configuration
+=============
+
+TODO
+
+Usage
+=====
+
+To use this module, see menu "Accounting > payment > SEPA direct debit mandates"
```
rgarnau: This does not match with the views. I think the menu would be 'Invoicing > Payments > Banking Mandates' instead.

JordiBForgeFlow (approved):
:+1: Technical and functional review.

HviorForgeFlow (approved):
LGTM :+1:

angelmoya:
Functional tested OK :+1:

PR #411, opened 2017-10-26
Translated spanish texts to german equivalents

j-mastr:
Translated supplied spanish texts to german equivalents


pedrobaeza:
Please use Transifex system instead for translating strings: https://odoo-community.org/page/translate

PR #414, opened 2017-10-30
[IMP] account_payment_order: Display partner account type in payment …

adrienpeiffer:
…order line tree view.

Currently, a payment order is considered as SEPA  only if all partner accounts are IBAN valid and the currency on all payment line is Euro.

with this commit, it's easier to understand why a payment order it's not in SEPA format.


pedrobaeza:
I prefer not to add more and more columns to the already crowded payment line tree. Can this be managed at import level putting a warning or directly not adding the line when payment method is SEPA?

PR #418, opened 2017-11-10
Use bank.payment.line reference in Payment Order Move Lines

ecino:
When closing an account.payment.order the move generated have in the reference field, the name of the payment order.

When payment are not grouped by date it generates a lot of moves that have the same reference and we propose to use instead the bank.payment.line reference in order to differentiate those moves.

We think it makes more sense for reconciling. If several bank.payment.line are groupped in a same move we do not change the previous behaviour.

### DIFF
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Account Payment Order',
-    'version': '10.0.1.5.0',
+    'version': '10.0.1.6.0',
     'license': 'AGPL-3',
     'author': "ACSONE SA/NV, "
               "Therp BV, "
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -375,6 +375,8 @@ class AccountPaymentOrder(models.Model):
             ref = _('Payment order %s') % self.name
         else:
             ref = _('Debit order %s') % self.name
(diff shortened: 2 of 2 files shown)


ecino:
I think this change is a nice improvement for the reconcile of payment lines. Plus it's a very small change, if someone is willing to review?

pedrobaeza (changes requested):
Please change the module version number, and IMO, we should have:
- Current ref
- Plus payment line number (or the name you suggest).

ecino:
@pedrobaeza I updated module version. What do you mean by current ref and payment line number? Do you mean `payment_order.name` and `bank_line.payment_line_ids.name` (but could link to several payment lines)?

pedrobaeza:
The code I suggest is:
```python
        if self.payment_type == 'outbound':
            ref = _('Payment order %s') % self.name
        else:
            ref = _('Debit order %s') % self.name
        if bank_lines and len(bank_lines) == 1:
            ref += " - " + bank_lines.name
```

ecino:
@pedrobaeza Ok, good point, I just updated my code.

pedrobaeza (approved):

pedrobaeza:
@alexis-via @sbidoul what do you think about this?

ecino:
Can we merge?

pedrobaeza:
As this is adding, not replacing, I think it's harmless, so I merge and cherry-pick for v11.

PR #419, opened 2017-11-15
[8.0] Adapts SEPA files to new rules that takes effect in November 19th

omar7r:
Hi,

The November 19th takes effect some new rules in SEPA files, the unique rule that affects technically to 8.0 version, because in 9.0 and 10.0 it is done, is the "PstlAdr" in "Dbtr" node, this node is only required when the debtor is from Switzerland, Monaco, San Marino and the French overseas collectivities Mayotte, Saint Pierre and Miquelon.

In 9.0 and 10.0 versions this node is always present in SEPA files if debtor has country, I based my solution in these versions.

Regards.

### DIFF
--- a/account_banking_pain_base/README.rst
+++ b/account_banking_pain_base/README.rst
@@ -56,7 +56,7 @@ Contributors
 
 * Alexis de Lattre
 * Pedro M. Baeza
-* Stéphane Bidoul              <stephane.bidoul@acsone.eu>
+* Stéphane Bidoul <stephane.bidoul@acsone.eu>
 * Ignacio Ibeas - Acysos S.L.
 * Alexandre Fayolle
 * Raphaël Valyi
--- a/account_banking_pain_base/README.rst
+++ b/account_banking_pain_base/README.rst
@@ -56,13 +56,14 @@ Contributors
 
 * Alexis de Lattre
 * Pedro M. Baeza
-* Stéphane Bidoul		<stephane.bidoul@acsone.eu>
+* Stéphane Bidoul              <stephane.bidoul@acsone.eu>
 * Ignacio Ibeas - Acysos S.L.
 * Alexandre Fayolle
 * Raphaël Valyi
 * Sandy Carter
 * Stefan Rijnhart (Therp)
 * Antonio Espinosa <antonioea@antiun.com>
+* Omar Castiñeira <omar@comunitea.com>
 
 Maintainer
 ----------
--- a/account_banking_pain_base/__openerp__.py
+++ b/account_banking_pain_base/__openerp__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Account Banking PAIN Base Module',
     'summary': 'Base module for PAIN file generation',
-    'version': '8.0.0.4.0',
+    'version': '8.0.0.4.1',
     'license': 'AGPL-3',
     'author': "Akretion, "
               "Noviat, "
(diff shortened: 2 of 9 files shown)


account_banking_pain_base/README.rst
```diff
@@ -56,13 +56,14 @@ Contributors
 
 * Alexis de Lattre
 * Pedro M. Baeza
-* Stéphane Bidoul		<stephane.bidoul@acsone.eu>
+* Stéphane Bidoul              <stephane.bidoul@acsone.eu>
```
pedrobaeza: Curious why this space

pedrobaeza (commented):
I'm not sure about changing the signature of that method. It can be used by specific SEPA adaptations. I think Switzerland has one. Can't you do it through a keyword argument and leave both codes?

account_banking_pain_base/README.rst
omar7r: My editor replaces tabulators for spaces.

pedrobaeza: Well, we can remove that spaces if you want, but no special problem.

omar7r:
The signature change is done in 9.0 and 10.0 versions too. Add a new parameter with the partner will create changes in other specific modules too.

pedrobaeza:
Yeah, but the difference is that 8.0 had the previous signature before this one, 9 and 10 no, so you can find that updating code from bank-payment makes other things (even inside OCA) to break and to require adaptation.

What others think?

pedrobaeza:
@yvaucher I remember you make some specific things for SEPA in Switzerland. Does this affect you?

omar7r:
Date reference: https://www.europeanpaymentscouncil.eu/document-library/implementation-guidelines/sepa-direct-debit-core-scheme-customer-bank

yvaucher:
For the Switzerland version

According to references

https://www.six-interbank-clearing.com/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf

On page 47, address can be structured (better choice) or not structured using AdrLine.

As swiss version has its specifics, just let enough hooks to let us make few changes.
Thus it would be better to extract this in a method to let us overwrite this part.
We will use this format instead:

``` xml
<Cdtr>
 <Nm>MUSTER AG</Nm>
  <PstlAdr>
   <StrtNm>Musterstrasse</StrtNm>
   <BldgNb>24</BldgNb>
   <PstCd>3000</PstCd>
   <TwnNm>Bern</TwnNm>
  <Ctry>CH</Ctry>
 </PstlAdr>
</Cdtr>
```

pedrobaeza:
Is this still needed? @yvaucher how did you deal with the changes?

yvaucher:
@pedrobaeza No change in since I linked is the reference document:

https://www.six-interbank-clearing.com/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf

It has not changed since 2017-12-18. It already includes changes that should take effect from November 2018

However, I realize this affects only version 8.0, I think it's fine to merge it. We had a custom module to generate pain files. https://github.com/OCA/l10n-switzerland/tree/8.0/l10n_ch_sepa

We only joined the OCA effort on this starting from version 9.0 https://github.com/OCA/l10n-switzerland/tree/9.0/l10n_ch_pain_base

yvaucher (approved):

pedrobaeza:
OK, merging then

PR #421, opened 2017-11-23
[10.0][ADD] account_banking_sepa_credit_transfer : pain.001.001.03.ch.02

grindtildeath:


pedrobaeza:
Shouldn't this be in Switzerland localization?

alexis-via:
It's already there: https://github.com/OCA/l10n-switzerland/blob/10.0/l10n_ch_pain_credit_transfer/models/account_payment_method.py

grindtildeath:
@pedrobaeza @alexis-via My bad, hadn't seen this. Thank you.

PR #424, opened 2017-11-28
[9.0][FIX] account_payment_mode: ACL for portal group in account.payment.mode model

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/account_payment_mode/__openerp__.py
+++ b/account_payment_mode/__openerp__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Account Payment Mode',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'license': 'AGPL-3',
     'author': "Akretion,Odoo Community Association (OCA)",
     'website': 'https://github.com/OCA/bank-payment',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Isn't better to put groups=employee in the view that adds the payment mode?

Any way, change module version number

sergio-teruel:
@pedrobaeza No, because a portl user wants view this information in his invoice

pedrobaeza:
My comment is still valid with what I'm saying

pedrobaeza:
But where is this seen? You don't add special QWeb fields for this, so it's a side effect more than the intended behavior.

sergio-teruel:
Account_payyment_partner module adds info to report invoice, I could create this ACL in this module, but is in account_payment_mode where the model is created
https://github.com/OCA/bank-payment/blob/9.0/account_payment_partner/views/report_invoice.xml#L12

pedrobaeza:
Ok, but as said in other place, this makes that any portal user sees all partner banks. Add record rule.

rafaelbn:
So, what should we do here @sergio-teruel @pedrobaeza ? Close it?

pedrobaeza:
Yes, closing it.

PR #425, opened 2017-11-28
[9.0][FIX] account_payment_partner: ACL for portal group for res.partner.bank model

sergio-teruel:


pedrobaeza:
@sergio-teruel please check the Travis status on this.

sergio-teruel:
@pedrobaeza Ready for merge

pedrobaeza:
But with this, you are giving access to any portal user to all your partners' bank accounts. You should restrict it by a record rule, which is not good for other that needs full access, or change it in the place where it fails (like the sudo in invoices).

sergio-teruel:
Ok, security read rule added

chienandalu (changes requested):
Failing on:

```
ParseError: "External ID not found in the system: account_payment_partner.model_account_banking_mandate" while parsing /home/travis/build/OCA/bank-payment/account_payment_partner/security/account_payment_partner_security.xml:5, near
```

pedrobaeza (changes requested):
This PR has several problems:

* Banking mandate model is not in this module, but on account_banking_mandate module.
* You are giving access to portal users to all partner banks.

PR #432, opened 2017-12-18
[8.0][ADD]account_payment_partner

almumu:
When creating a refund, get by default the payment mode from origin invoice


StefanRijnhart:
Not sure about this one. The way I use payment modes, we have distinct payment modes for debit and credit transactions. Therefore, the payment mode of the refund is always different from the original one.

pedrobaeza:
I agree with @StefanRijnhart. This will cause an error of mismatching payment modes. Closing then.

PR #433, opened 2017-12-19
[10.0] [ADD] payment_line cancel module

ecino:
New module that adds an Action on invoices in order to cancel them from payment orders. It also adds a button on payment lines to cancel them, no matter the state of the payment order.

This is useful when a payment is rejected for instance and we want to try again to import the invoice in a new payment order.

image


pedrobaeza:
AFAIK, cancel option is already included in standard module.

ecino:
Sorry, maybe my description is not clear enough. By default, the cancel option is only able to cancel a whole payment order. With this module, it is possible to cancel at the payment.line level, and keep the payment order confirmed.

pedrobaeza:
OK, got it. The module name and description needs to be improved indeed

ecino:
I will change that!

ecino:
I changed the module name to `account_payment_line_cancel` and changed the descriptions. Screenshot included in the PR description.

ecino:
Anyone willing to review this?

account_payment_line_cancel/README.rst
```diff
@@ -0,0 +1,93 @@
...
+   :alt: License: AGPL-3
+
+===================
+Cancel payment line
+===================
+
+This module allows you to cancel and remove a payment line from a payment order when the payment needs to be cancel (for example when a payment is rejected).
+The module allows you as well to free an invoice that was imported in a payment order and automatically cancel the payment lines that were generated.
+
+Installation
```
pedrobaeza: Remove empty sections

account_payment_line_cancel/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+#. Go to invoices
+#. Select a few invoices from the tree view
+#. Use the "Free invoices" entry to remove selected invoices from payment orders
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/173/10.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza: Remove this part

account_payment_line_cancel/README.rst
```diff
@@ -0,0 +1,93 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===================
+Cancel payment line
+===================
+
+This module allows you to cancel and remove a payment line from a payment order when the payment needs to be cancel (for example when a payment is rejected).
```
pedrobaeza: s/needs to be cancel/needs to be cancelled

account_payment_line_cancel/models/account_payment_line.py
```diff
@@ -0,0 +1,101 @@
...
+         from payment order
+
+        A cancel reason can be put in the cancel_reason field.
+
+        :return: True
+        """
+        # Retrieve all account_move_line(s) associated with the previous
+        # account_payment_line(s).
+        all_account_move_lines = self.mapped('move_line_id')
+
```
pedrobaeza: Personally I don't like this unneeded extra empty lines that only forces to scroll you more. PEP8 also discourages its use.

account_payment_line_cancel/models/account_payment_line.py
```diff
@@ -0,0 +1,101 @@
...
+            # Retrieve the counterparts of the previous move_lines.
+            # The counter part should always be unique.
+            account_move_line_counterpart = self.env[
+                'account.move.line'].search([
+                    ('full_reconcile_id', 'in', full_reconcile_ids),
+                    ('id', 'not in', all_account_move_lines.ids),
+                    ('move_id.payment_order_id', 'in', payment_orders.ids)])
+
+            # All the move lines with the same reconcile id are unreconciled.
+            account_move_line_counterpart.remove_move_reconcile()
```
pedrobaeza, line 42: Are you sure this should be done automatically? It's very risky IMO

ovnicraft:
@ecino can you resolve  changes requested ?

account_payment_line_cancel/models/account_payment_line.py
ovnicraft: @pedrobaeza how can be better here? 
first possible risk: line was partial reconciled in another side, its possible ?

ecino: If we don't do it, the cancel method leaves move lines that are no more relevant, and the accounting people need to be aware of that. I don't think it's risky, as we unreconcile and remove the payment lines that were created especially for the payment line.

ecino:
Code corrections done

ovnicraft (approved):
Ok, got it with your comment @ecino its more clear for me.

account_payment_line_cancel/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        'account_cancel',
+    ],
+    'data': [
+        'views/invoice_view.xml',
+        'views/payment_line_view.xml',
+    ],
+    'demo': [],
+    'test': [],
+    'installable': True,
+}
```
MeetKD: Pleass Change website key https://github.com/OCA/bank-payment
Remove empty keys

account_payment_line_cancel/models/account_payment_line.py
```diff
@@ -0,0 +1,98 @@
...
+# @author: Marco Monzione <marco.mon@windowslive.com>, Emanuel Cino
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, _
+
+
+class AccountCancelPayment(models.Model):
+    _inherit = 'account.payment.line'
+
+    cancel_reason = fields.Char()
```
MeetKD, line 12: Char or text?

ecino:
@pedrobaeza Do you see any better alternative? If we don't do it, the cancel method leaves move lines that are no more relevant, and the accounting people need to be aware of that.

pedrobaeza:
OK, if it's safe in your opinion, go ahead.

MeetKD (changes requested):

account_payment_line_cancel/models/account_payment_line.py
ecino: I think this field is no need to be Text, this should be a simple short reason like "bank not valid" or "customer cancelled". We use it with what bank returns and the messages are short.

ecino:
@MeetKD I fixed the manifest file.

ecino:
Can we merge? We need this module for l10n-ch repository, for importing bank return files to parse transactions that failed and remove them from payment orders.

pedrobaeza (approved):

PR #434, opened 2017-12-20
[11.0][MIG] account_payment_sale

ghost:

PR #435, opened 2018-01-01
[11.0] account_payment_sale: Migration to version 11.0

Hamza-Talibi:


pedrobaeza:
There's already a migration in #434. Please check first if the migration is done and review that one instead.

PR #436, opened 2018-01-04
Renamed directory names

melroy89:
Fix folder names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ OCA banking payment addons for Odoo
 On version 11.0, this project focus on payment interface. The indentation below 
 indicates the dependency graph of the main modules.
 
--  `account_banking_pain_base` - Basic export functionality of payment orders (depends on `account_payment_order` & `account_payment_partner`)
+-  `account_payment_order` - Basic export functionality of payment orders
 
     - `account_banking_sepa_credit_transfer` - Export of payment orders in SEPA format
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,13 +8,13 @@ OCA banking payment addons for Odoo
 On version 11.0, this project focus on payment interface. The indentation below 
 indicates the dependency graph of the main modules.
 
-- account_banking_payment_export - Basic export functionality of payment orders
+-  `account_banking_pain_base` - Basic export functionality of payment orders (depends on `account_payment_order` & `account_payment_partner`)
```
pedrobaeza: It should be `account_payment_order`

pedrobaeza (changes requested):

README.md
melroy89: Uh.. The 'base' functionality of the SEPA modules are 'account_banking_pain_base', right? Not account_payment_order, besides the fact that account_banking_pain_base is also depending on account_payment_order, what I wrote between brackets.

pedrobaeza: But this is not for SEPA, but for making payment orders for banks. American banks don't use SEPA for example.

pedrobaeza (approved):

melroy89:
Let's also do fast tracking here 👍

pedrobaeza:
OK, I can assume that "risk" :wink:

PR #437, opened 2018-01-11
[FIX] Ensure that the test finds only one bank journal

aisopuro:
Previously it was possible to find multiple bank journals,
which caused singleton errors when the test tried to set values
on them.


pedrobaeza (approved):

jabibi (approved):
:+1:

PR #438, opened 2018-01-17
[11.0][FIX] account_payment_partner: impossible to define if there was no company in partner

oihane:
If there is a partner defined for all possible companies it does not allow to define a payment mode. I'm not sure if this is the best way, but it works.

cc @carlosdauden @jbeficent @alexis-via @pedrobaeza @sbidoul @etobella

### DIFF
--- a/account_payment_partner/__manifest__.py
+++ b/account_payment_partner/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Account Payment Partner',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.1.0',
     'category': 'Banking addons',
     'license': 'AGPL-3',
     'summary': 'Adds payment mode on partners and invoices',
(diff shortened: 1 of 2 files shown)


alexis-via (approved):

alexis-via:
It seems this mistake was introduced in v11, it is not there in previous versions

alexis-via:
The person that introduced the bug is @etobella https://github.com/OCA/bank-payment/commit/e322814034884e7e1c4549c31cd7e4a2deebfe28

carlosdauden (approved):

etobella (approved):
Sorry, I tried to simplify its usage in multicompany environments.

pedrobaeza:
The idea is if you are admin or any user with multiple visible companies (parent and children), you select only the correct payment mode.

The problem as mentioned is if your partner doesn't define a company because it's "global". Maybe a dynamic domain can help here, but for now, let's remove it for avoiding a non desired bug.

Issue #441, opened 2018-01-22
[11ce] Error when entering a vendor which has 2 banks on a supplier invoice 


ThijsvOers:
Steps to reproduce:
Create a vendor with 2 bank numbers AND fill in the standard  payment mode field. Then create a new invoice and enter the name of the vendor, you will get the following error:

Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 646, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 307, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/odoo-11.0/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 683, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 339, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo-11.0/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 332, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 927, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo-11.0/odoo/http.py", line 512, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/odoo-server/addons/web/controllers/main.py", line 930, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/odoo-server/addons/web/controllers/main.py", line 922, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo-11.0/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo-11.0/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/odoo-11.0/odoo/models.py", line 4995, in onchange
    record._onchange_eval(name, field_onchange[name], result)
[...]

PR #442, opened 2018-01-22
[11.0][FIX] account_payment_partner: onchange if more than one bank

astirpe:
Fixes https://github.com/OCA/bank-payment/issues/441

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -36,7 +36,7 @@ class AccountInvoice(models.Model):
                     self.partner_bank_id = \
                         self.commercial_partner_id.bank_ids.filtered(
                             lambda b: b.company_id == self.company_id or not
-                            b.company_id)
+                            b.company_id)[:1]
                 else:
                     self.partner_bank_id = False
 
--- a/account_payment_partner/__manifest__.py
+++ b/account_payment_partner/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Account Payment Partner',
-    'version': '11.0.1.2.0',
+    'version': '11.0.1.2.1',
     'category': 'Banking addons',
     'license': 'AGPL-3',
     'summary': 'Adds payment mode on partners and invoices',


account_payment_partner/models/account_invoice.py
```diff
@@ -27,18 +27,19 @@ def _onchange_partner_id(self):
...
                     self.commercial_partner_id.bank_ids
                 ):
-                    self.partner_bank_id = \
+                    partner_banks = \
                         self.commercial_partner_id.bank_ids.filtered(
                             lambda b: b.company_id == self.company_id or not
                             b.company_id)
-                else:
-                    self.partner_bank_id = False
+                    if partner_banks:
```
pedrobaeza: No need to make `if`. If you put at the end `[:1]`, it servers for 0 or more elements to filter the first one.

astirpe:
@pedrobaeza this Runbot warning:

```
2018-01-22 12:58:20,470 168 WARNING openerp_test odoo.addons.sms.wizard.send_sms: The `phonenumbers` Python module is not installed. Try: pip install phonenumbers.
```

is the one (I guess) that makes the runbot icon red. I see this warning in every repository for every new PRs on branch 11.0. Can we install the `phonenumbers`? I don't know if you can do it or is someone else.

ThijsvOers (approved):
I have tested the latest commit (more than one bank number per supplier in a multi company environment gave an error when creating a vendor bill) and everything is working fine. 

An extra feature request would be to add an extra priority field to the bank number on a vendor/customer. But maybe there is already such a request.

account_payment_partner/models/account_invoice.py
```diff
@@ -27,6 +27,7 @@ def _onchange_partner_id(self):
                     force_company=self.company_id.id
                 ).partner_id.supplier_payment_mode_id
                 self.payment_mode_id = pay_mode
+                self.partner_bank_id = False
```
pedrobaeza: This is not needed now

account_payment_partner/models/account_invoice.py
```diff
@@ -36,9 +37,7 @@ def _onchange_partner_id(self):
                     self.partner_bank_id = \
                         self.commercial_partner_id.bank_ids.filtered(
                             lambda b: b.company_id == self.company_id or not
-                            b.company_id)
-                else:
-                    self.partner_bank_id = False
```
pedrobaeza: Please keep this for avoiding a double assignation plus possible triggers

pedrobaeza:
The warning is because a module is not correctly informing about a missing library. It should use `logger.info`, not `logger.warn`. Any way, it's easy to fix: add the library to the requirements.txt file in the root of the repository that hosts the module for being automatically installed on runbot.

pedrobaeza (approved):

Issue #440, opened 2018-01-19
[11.0][BUG] account_payment_order: error printing "Payment Order" 


astirpe:
Steps to reproduce in Runbot:

- Invoicing > Payments > Payment Orders
- Select some records
- Print > Payment Order

```
Traceback (most recent call last):
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 1663, in report_download
    response = self.report_routes(reportname, docids=docids, converter='pdf')
  File "/.repo_requirements/odoo/odoo/http.py", line 512, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 1612, in report_routes
    pdf = report.with_context(context).render_qweb_pdf(docids, data=data)[0]
  File "/.repo_requirements/odoo/odoo/addons/base/ir/ir_actions_report.py", line 624, in render_qweb_pdf
    html = self.with_context(context).render_qweb_html(res_ids, data=data)[0]
  File "/.repo_requirements/odoo/odoo/addons/base/ir/ir_actions_report.py", line 654, in render_qweb_html
    data = report_model.get_report_values(docids, data=data)
AttributeError: 'report.account_payment_order.print_account_payment' object has no attribute 'get_report_values'
```

PR #443, opened 2018-01-22
[11.0][FIX] account_payment_order: printing report

astirpe:
Fixes https://github.com/OCA/bank-payment/issues/440

### DIFF
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': 'Account Payment Order',
-    'version': '11.0.1.1.0',
+    'version': '11.0.1.1.1',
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Check that fixes the problem, so I fast-track it for have it correct.

Runbot error is a warning and it's something introduced by Odoo itself with a lib not declared on requirements.txt: https://github.com/odoo/odoo/blob/f68ad51520150cd721814555b9e025da310204eb/addons/sms/wizard/send_sms.py#L18

I don't know if we should add it to OCB requirements. Let me think about that.

pedrobaeza:
I have finally decided to add it: https://github.com/OCA/OCB/pull/696

Issue #439, opened 2018-01-19
[11.0][BUG] account_banking_sepa_direct_debit: error printing "Sepa mandate"


astirpe:
Steps to reproduce in Runbot:

- Invoicing > Payments > Banking mandates
- Select some records
- Print > Sepa mandate

PR #444, opened 2018-01-22
[11.0][FIX] account_banking_sepa_direct_debit: printing report

astirpe:
Fixes https://github.com/OCA/bank-payment/issues/439

### DIFF
--- a/account_banking_sepa_direct_debit/reports/sepa_direct_debit_mandate.xml
+++ b/account_banking_sepa_direct_debit/reports/sepa_direct_debit_mandate.xml
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
+    <template id="assets_common" name="account_banking_sepa_direct_debit pdf assets" inherit_id="web.report_assets_common">
+        <xpath expr="." position="inside">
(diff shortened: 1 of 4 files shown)


pedrobaeza:

astirpe:
@pedrobaeza done

account_banking_sepa_direct_debit/reports/sepa_direct_debit_mandate.xml
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
+    <template id="assets_common" name="account_banking_sepa_direct_debit pdf assets" inherit_id="web.report_assets_common">
```
pedrobaeza, line 4: With this, you are overwriting the style for all the reports. In-line style is not working? If you want to do it this way, please put all the styles under a class for not changing general ones.

pedrobaeza (changes requested):

account_banking_sepa_direct_debit/reports/sepa_direct_debit_mandate.xml
astirpe: @pedrobaeza you're right! I made the changes, could you check?

pedrobaeza (approved):
As this is broken, I fast-track it for fixing it

PR #445, opened 2018-01-31
[8.0][FIX] Add signature_date for _check_valid_state constraint

lembregtse:
We have had problems with users not being able to generate payment files because of "suddenly" missing signature date's for valid mandates. 

I think it is best to also add "signature_date" as a constrains parameter to the "_check_valid_state" constraint on account.banking.mandate, it makes sense that if "signature_date" is part of the constraint check itself to also check the constraint when the field updates.

For our case it would prevent the mandate from removing the "signature_date" when the mandate is still considered "valid"

### DIFF
--- a/account_banking_mandate/__openerp__.py
+++ b/account_banking_mandate/__openerp__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Account Banking Mandate',
     'summary': 'Banking mandates',
-    'version': '8.0.0.2.0',
+    'version': '8.0.0.2.1',
     'license': 'AGPL-3',
     'author': "Compassion CH, "
               "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Thanks for the explanation. It makes sense, but please change module version according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

lembregtse:
Thank you for the comment, I've updated accordingly.

pedrobaeza (approved):
It would be good to check if the problem is present in the rest of versions (9.0, 10.0 and 11.0)

lembregtse:
Yes, I have seen that the same code is used up till 11.0 so the same fix could be applied on those versions.

pedrobaeza:
OK, we can cherry-pick the commit.

StefanRijnhart (approved):

pedrobaeza:
Cherry-picked for:

* v9: 8ac6afeb0e5a215225a7d60d5341caa48f534c0d
* v10: 12f82aa151bbebf964f12e5197f5b3643bac2340
* v11: 5d847f804b69cbafbdfa130e7a1b3942ebc8580d

Issue #446, opened 2018-02-06
[11]Error when creating bank mandate


jalzaga:
When you try to create a bank mandate the system does not allow it and gives the following message: 
Error while validating constraint
'res.partner.bank' object has no attribute 'name'

To reproduce it: Go to a customer, create a bank account for this customer or use an existing one, edit the bank account and add a direct debit mandate, fill in the fields, save, and when you try to save the bank account it gives the error message.
This has been tested in runbot.

JordiBForgeFlow:
On what version?

jalzaga:
version 11 @jbeficent

JordiBForgeFlow:
cc @rgarnau can you try to reproduce?

jalzaga:
@jbeficent  @rgarnau  you can see it on runbot, just try to create for instance a new mandate in the Camptocamp's bank account.

rgarnau:
@jalzaga can you check if this PR fixes the error you mention? https://github.com/OCA/bank-payment/pull/447

jalzaga:
cheked #447  NOT successful  @rgarnau @jbeficent

pedrobaeza:
Final fix is correct, so I close this.

PR #447, opened 2018-02-07
[11.0][FIX] account_banking_mandate: Allow to create mandates from bank accounts

rgarnau:
PR in response to https://github.com/OCA/bank-payment/issues/446

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -18,6 +18,13 @@ class AccountBankingMandate(models.Model):
     _inherit = ['mail.thread']
     _order = 'signature_date desc'
 
+    def _get_default_partner_bank_id_domain(self):
+        if 'default_partner_id' in self.env.context:
+            return [('partner_id', '=', self.env.context.get(
+                'default_partner_id'))]
+        else:
+            return []
+
     format = fields.Selection(
         [('basic', 'Basic Mandate')], default='basic', required=True,
         string='Mandate Format', track_visibility='onchange')
@@ -28,7 +35,8 @@ class AccountBankingMandate(models.Model):
     )
     partner_bank_id = fields.Many2one(
         comodel_name='res.partner.bank', string='Bank Account',
-        track_visibility='onchange')
+        track_visibility='onchange',
+        domain=lambda self: self._get_default_partner_bank_id_domain(),)
     partner_id = fields.Many2one(
         comodel_name='res.partner', related='partner_bank_id.partner_id',
         string='Partner', store=True)
--- a/account_banking_mandate/models/res_partner_bank.py
+++ b/account_banking_mandate/models/res_partner_bank.py
@@ -15,14 +15,13 @@ class ResPartnerBank(models.Model):
         help='Banking mandates represent an authorization that the bank '
              'account owner gives to a company for a specific operation.')
 
-    @api.constrains('company_id', 'mandate_ids')
+    @api.constrains('company_id')
     def _company_constrains(self):
         for rpb in self:
-            if self.env['account.banking.mandate'].search(
+            if self.env['account.banking.mandate'].sudo().search(
                     [('partner_bank_id', '=', rpb.id),
-                     ('company_id', '=', rpb.company_id.id)], limit=1):
+                     ('company_id', '!=', rpb.company_id.id)], limit=1):
                 raise ValidationError(
                     _("You cannot change the company of Partner Bank %s, "
                       "as there exists mandates referencing it that "
-                      "belong to another company.") %
-                    (rpb.name,))
+                      "belong to another company.") % (rpb.display_name,))
--- a/account_banking_mandate/views/account_banking_mandate_view.xml
+++ b/account_banking_mandate/views/account_banking_mandate_view.xml
@@ -12,11 +12,14 @@
     <field name="arch" type="xml">
         <form string="Banking Mandate">
             <header>
-                <button name="validate" type="object" string="Validate" states="draft" class="oe_highlight"/>
-                <button name="cancel" type="object" string="Cancel" states="draft,valid"/>
+                <button name="validate" type="object" string="Validate" states="draft" class="oe_highlight"
+                        invisible="context.get('mandate_bank_partner_view')"/>
+                <button name="cancel" type="object" string="Cancel" states="draft,valid"
+                        invisible="context.get('mandate_bank_partner_view')"/>
                 <button name="back2draft" type="object" string="Back to Draft"
                     states="cancel" groups="account.group_account_manager"
-                    confirm="You should set a mandate back to draft only if you cancelled it by mistake. Do you want to continue?"/>
+                    confirm="You should set a mandate back to draft only if you cancelled it by mistake. Do you want to continue?"
+                    invisible="context.get('mandate_bank_partner_view')"/>
                 <field name="state" widget="statusbar"/>
             </header>
             <sheet>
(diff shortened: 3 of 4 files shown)


jalzaga (commented):
I have created a new mandate for an existing bank account  of Agrolait. When  y tried to save the changes in bank account it gives the following error: 
You cannot change the company of Partner Bank Agrolait, as there exists mandates referencing it that belong to another company.
Then I have created a new bank account for The Jackson Group, I have created a mandate for this bank account but, when I saved the bankaccount it gives the same error: You cannot change the company of Partner Bank The Jackson Group, as there exists mandates referencing it that belong to another company.

rgarnau:
@jalzaga I think now it's fixed. Can you check?

jalzaga (commented):
Checked. Mandates created and validated.
Now it is correct.

account_banking_mandate/views/account_banking_mandate_view.xml
```diff
@@ -32,11 +32,11 @@
                     <field name="format" string="Format"/>
                     <field name="type" string="Type"/>
                     <field name="partner_bank_id"
-                        options="{'create_name_field': 'acc_number'}"
-                        invisible="context.get('mandate_bank_partner_view')" />
+                           options="{'create_name_field': 'acc_number'}"
```
pedrobaeza: Don't make this stylistic changes in a minor fix.

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -145,6 +145,14 @@ def create(self, vals=None):
             vals['unique_mandate_reference'] = \
                 self.env['ir.sequence'].next_by_code(
                     'account.banking.mandate') or 'New'
+        if self._context.get('mandate_bank_partner_view'):
+            partner_id = self._context['partner_id']
```
pedrobaeza: Don't use internal variable. Use the same as the second line (`self.env.context`)

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -145,6 +145,14 @@ def create(self, vals=None):
...
+            acc_num = self.env.context.get('acc_number')
+            partner_bank_id = self.env['res.partner.bank'].search([
+                ('partner_id', '=', partner_id),
+                ('acc_number', '=', acc_num)], limit=1)
```
pedrobaeza: Why not passing directly the res.partner.bank record ID and remove this ugly search?

account_banking_mandate/models/res_partner_bank.py
```diff
@@ -15,14 +15,14 @@ class ResPartnerBank(models.Model):
         help='Banking mandates represent an authorization that the bank '
              'account owner gives to a company for a specific operation.')
 
-    @api.constrains('company_id', 'mandate_ids')
```
pedrobaeza, line 18: Why removing the condition when a new mandate is created?

account_banking_mandate/models/res_partner_bank.py
```diff
@@ -15,14 +15,14 @@ class ResPartnerBank(models.Model):
...
             if self.env['account.banking.mandate'].search(
                     [('partner_bank_id', '=', rpb.id),
-                     ('company_id', '=', rpb.company_id.id)], limit=1):
+                     ('company_id', '!=', rpb.company_id.id)], limit=1):
                 raise ValidationError(
                     _("You cannot change the company of Partner Bank %s, "
                       "as there exists mandates referencing it that "
                       "belong to another company.") %
-                    (rpb.name,))
+                    (rpb.partner_id.name,))
```
pedrobaeza: Don't change this, as the text say Partner Bank, not Partner

account_banking_mandate/models/res_partner_bank.py
```diff
@@ -15,14 +15,14 @@ class ResPartnerBank(models.Model):
         help='Banking mandates represent an authorization that the bank '
              'account owner gives to a company for a specific operation.')
 
-    @api.constrains('company_id', 'mandate_ids')
+    @api.constrains('company_id')
     def _company_constrains(self):
         for rpb in self:
             if self.env['account.banking.mandate'].search(
```
pedrobaeza: You have to make this with `sudo`, or it won't work for regular users.

pedrobaeza (changes requested):
Please take care a bit about PR subject as it's very important for knowing what are you doing. Follow this syntax (I have already changed it for this one):
```
[version][pr_type] module: Real summary of the PR
```

account_banking_mandate/models/account_banking_mandate.py
rgarnau: This was the main problem of this module. The res.partner.bank is not properly passed through the view. Instead of passing the `active_id` of the res.partner.bank, it passes the id of the res.partner belonging to that res.partner.bank.
As mandates are created through a one2many with this approach the job is done correctly.

account_banking_mandate/models/res_partner_bank.py
rgarnau: This constraint is checking that the user doesn't change the company of the res.partner.bank causing incongruencies. If mandates are created doesn't affect this.

pedrobaeza: Yes, it affects as well, as you can introduce a mandate that doesn't follow the rules. Think twice about this.

account_banking_mandate/models/account_banking_mandate.py
pedrobaeza: Is this a one2many? Then the problem is that you are showing the partner_bank_id field in it. If not shown, then all is handled automatically by Odoo.

account_banking_mandate/models/res_partner_bank.py
rgarnau: There's already a line that checks this when creating a mandate: https://github.com/OCA/bank-payment/blob/11.0/account_banking_mandate/models/account_banking_mandate.py#L84

pedrobaeza: Then you can remove that line and let this one. It's redundant. KISS principle.

account_banking_mandate/models/account_banking_mandate.py
rgarnau: I see your point but partner_bank_id it is shown: https://github.com/OCA/bank-payment/blob/11.0/account_banking_mandate/views/account_banking_mandate_view.xml#L34

account_banking_mandate/models/res_partner_bank.py
rgarnau: I don't think so. This applies only when creating a mandate through a res.partner.bank, but if you were to create a mandate from scratch then this code wouldn't be executed.

account_banking_mandate/models/account_banking_mandate.py
pedrobaeza: Then create a specific embedded view for this one and everything solved without this ugly code.

account_banking_mandate/models/res_partner_bank.py
pedrobaeza: Are you sure. Please try, as I think it's also applied. Take into account that the source doesn't matter. What determines if something has changed is the one2many in the model, which is always updated.

oihane: @rgarnau is correct, if you remove the constraint from mandate it allows you to add a mandate that has a different company.

If you maintain the 'mandate_ids' in the constraint, and erase the 'company_id' constraint from mandate (I mean the one that compares mandate and bank account companies) then you can create a mandate directly with not the same company.
If you have the company constraint in the mandate, even if you change the mandate_ids, it will raise first the one in the mandate.

fireshot capture 16 - fr66 1212 1212 1212 1212 1212 121 - od_ - http___localhost_9169_web id 11 vi
You can see in the picture how I was able to have mandates with different companies.

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -145,6 +145,14 @@ def create(self, vals=None):
             vals['unique_mandate_reference'] = \
                 self.env['ir.sequence'].next_by_code(
                     'account.banking.mandate') or 'New'
+        if self._context.get('mandate_bank_partner_view'):
```
oihane: @rgarnau You just can remove 'default_partner_bank_id' from the view, and it will work without needing all this code. Please can you check it out?

rgarnau: @oihane - Thanks for your help! It works but then I cannot validate the mandate right away as the information passed it's not ready until you save (if you validate it gives you an error that bank account is missing). In order to fix this, I'm adding the option of validating in the tree view and hiding it from the mandate when creating from a res.partner.bank. What do you think?

rgarnau:
These are the new changes that have been applied:
- In order to remove the search that @pedrobaeza was mentioning, the `active_id` passed through the view of `res.partner.bank` to `account.banking.mandate` has been removed (which means the mandate cannot be validated at the moment of creation) and instead, that possibility is given with buttons in the view of  `res.partner.bank` and can only be used once you save:

image

- A domain has also been added when creating a mandate from `res.partner`, as the possible domain of `res.partner.bank` in this situation is limited to that particular partner.

@pedrobaeza @oihane do you agree with the changes?

cc @jbeficent

JordiBForgeFlow:
I have reviewed with @rgarnau and seems good to me.

oihane (approved):
LGTM

pedrobaeza (approved):
Let's merge this for unblocking the situation, but still see some weird things. I'll try and tell you back.

PR #449, opened 2018-02-12
10.0 [FIX] account_payment_partner: Invoice report

acsonefho:
**Actual behaviour**
- If you print the report and if the field `bank_id` of the `partner_bank_id` is not set, you'll have a traceback. It's because the `bank_id` field is not mandatory.
So the report try to concatenate `False` (due to not filled value) with a `string` and raise an exception.

**Fix**
- Just use a `t-if` to check if the value is set.

### DIFF
--- a/account_payment_partner/views/report_invoice.xml
+++ b/account_payment_partner/views/report_invoice.xml
@@ -9,7 +9,9 @@
             </p>
             <p t-if="o.partner_bank_id">
                 <strong>Bank Account:</strong>
-                <t t-esc="o.partner_bank_id.bank_id.name + ('' if not o.partner_bank_id.bank_id.bic else ' (' + o.partner_bank_id.bank_id.bic + ')')" />
+                <t t-if="o.partner_bank_id.bank_id">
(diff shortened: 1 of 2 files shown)


account_payment_partner/views/report_invoice.xml
```diff
@@ -9,7 +9,7 @@
             </p>
             <p t-if="o.partner_bank_id">
                 <strong>Bank Account:</strong>
-                <t t-esc="o.partner_bank_id.bank_id.name + ('' if not o.partner_bank_id.bank_id.bic else ' (' + o.partner_bank_id.bank_id.bic + ')')" />
+                <t t-esc="(o.partner_bank_id.bank_id.name or '') + ('' if not o.partner_bank_id.bank_id.bic else ' (' + o.partner_bank_id.bank_id.bic + ')')" />
```
pedrobaeza: I think in this case it's better to put directly `<t t-if="o.partner_bank_id.bank_id">`

pedrobaeza (changes requested):

acsonefho:
@pedrobaeza I update my commit with your comments: Increase the version number and use a `t-if`.

pedrobaeza (approved):

Cedric-Pigeon (approved):

PR #450, opened 2018-02-12
[9.0][MIG] account_payment_mode: Migration script for store related payment_type

sergio-teruel:
cc @Tecnativa
@pedrobaeza Can you take a look?


pedrobaeza (changes requested):
It's strange that this is not saved, but please add at the end `WHERE payment_type IS NULL` for avoiding rewritings.

pedrobaeza:
Check Travis also.

sergio-teruel:
@pedrobaeza Changes done

pedrobaeza (approved):

PR #452, opened 2018-02-20
[10.0] account_banking_sepa_direct_debit: Filter by mandate scheme

omar7r:
Hi,

At least in Spain is necessary in most banks create a sepa file for B2B mandates and another sepa file for CORE mandates, not mixed, for this reason I added the possibility of filter account move lines by mandate scheme, it is useful in the importation wizard of moves in the payment order when the user search manually.

Regards


account_banking_sepa_direct_debit/__manifest__.py
```diff
@@ -28,6 +28,7 @@
         'data/report_paperformat.xml',
         'reports/sepa_direct_debit_mandate.xml',
         'views/report_sepa_direct_debit_mandate.xml',
+        'views/account_move_line_view.xml'
```
jabibi: comma on the last item of the list

account_banking_sepa_direct_debit/models/account_move_line.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2018 Comunitea Servicios Tescnológicos S.L. - Omar Castiñeira
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
```
jabibi: imports order

jabibi (changes requested):
Thanks for the contribution Omar. Just few code review

jabibi (approved):
:+1:

StefanRijnhart (commented):
Don't you want to implement a standard filter that is configured on the payment mode, like there is on journals? Given the current implementation you might want to add an index on the mandate's scheme field. And there is a conflict that needs to be resolved.

omar7r:
Hi @StefanRijnhart 
The payment mode is the same for CORE and B2B mandates, maybe I could create the field in the import invoices wizard in payment order. ok?

StefanRijnhart:
@omar7r Interesting. By creating distinct payment modes (with the same SEPA SDD target) you could link the B2B mode to B2B customers and/or configure the mandate type on the payment mode itself. What would be the downside of that?

omar7r:
It is more complicated for Odoo 's user, having to manage two payment modes for the same, a sepa direct debit.

StefanRijnhart:
Matter of taste, but in that case you just add the index and we're good to go.

omar7r:
@StefanRijnhart index added

StefanRijnhart:
That index on account_move_line's mandate_id is useful as well but I asked for an index on the mandate's scheme field. Can you add that one as well?

pedrobaeza:
But I think this should be handled at other level, as you already have the scheme at mandate level, so duplicating the data is not the best option, and more with the large amount of the move lines. Imagine you change the mandate type by mistake... the recomputing of all associated move lines will be huge.

I think we should add the filtering in the import dialog inside payment orders instead.

omar7r:
The scheme field in account move line proposed is not stored, because it is only needed for searching, then that shouldn't be a problem, but maybe add it in import wizard could be better, although it is only usable for direct debits, I commented it too, 7 comments before.

PR #454, opened 2018-03-01
[10.0][FIX] account_banking_pain_base. Only unidecode unicode objects.

NL66278:
The present code gives warnings when trying to unidecode non unicode strings. This gives yellow branches in runbot. The fix in this branch prevents unneeded warnings (and unneeded tries to decode strings that are utf-8 already).


pedrobaeza (approved):

NL66278:
@pedrobaeza Done. I modified the last digit of the version number.

jabibi (approved):
:+1:

PR #457, opened 2018-03-14
[10.0] Usability of account_payment_order: access journal entries via top-right button

alexis-via:
This is a small usability improvement in the module account_payment_order: access journal entries via top-right button (that display the number of moves) instead of a dedicated tab. Advantage: you can access the journal entries full-screen, instead of inside a pop-up.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -91,7 +91,15 @@ class AccountPaymentOrder(models.Model):
     move_ids = fields.One2many(
         'account.move', 'payment_order_id', string='Journal Entries',
         readonly=True)
+    move_count = fields.Integer(
+        compute='_compute_move_count', readonly=True, store=True,
+        string="# of Journal Entries")
     description = fields.Char()
+
+    @api.depends('move_ids')
+    def _compute_move_count(self):
+        for order in self:
+            order.move_count = len(order.move_ids)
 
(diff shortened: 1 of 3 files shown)


elicoidal (approved):
Could you bump version number?
Other than that LGTM

alexis-via:
Isn't it automatic upon merge now ?

pedrobaeza:
I'm afraid not yet

andhit-r (approved):

account_payment_order/models/account_payment_order.py
```diff
@@ -91,8 +91,16 @@ class AccountPaymentOrder(models.Model):
...
         readonly=True)
+    move_count = fields.Integer(
+        compute='_compute_move_count', readonly=True, store=True,
+        string="# of Journal Entries")
     description = fields.Char()
 
+    @api.depends('move_ids')
+    def _compute_move_count(self):
+        for order in self:
+            order.move_count = len(order.move_ids)
```
pedrobaeza: This can scale bad when lot of moves, as this triggers the read of the records. Better to use `read_group` with domain `[('id', 'in', order.move_ids)]` and get the count from one of the values of the returned dictionary.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #458, opened 2018-03-15
[IMP] account_payment_partner: Add options to show partner bank accou…

carlosdauden:
…nt in invoice report

It is possible that this isn't the most appropriate module, but this is the module that adds the bank account to the invoice report.

@Tecnativa


account_payment_partner/README.rst
```diff
@@ -67,6 +67,8 @@ Contributors
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
 * Danimar Ribeiro
 * Angel Moya <angel.moya@domatix.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+  * Carlos Dauden
```
pedrobaeza, line 123: You need to put a line break between and move my name inside

account_payment_partner/models/account_payment_method.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountPaymentMethod(models.Model):
+    _inherit = 'account.payment.method'
+
+    show_bank_account = fields.Selection(
+        [
```
pedrobaeza: Put keyword argument name (`selection=`)

account_payment_partner/models/account_payment_method.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountPaymentMethod(models.Model):
+    _inherit = 'account.payment.method'
```
pedrobaeza: I would put this at payment mode level, not payment method. This way, using same SEPA direct debit payment method, you can define one payment mode showing full bank number and another not showing it for example.

account_payment_partner/models/account_payment_method.py
```diff
@@ -0,0 +1,23 @@
...
+            ('full', 'Full'),
+            ('first', 'First n chars'),
+            ('last', 'Last n chars'),
+            ('no', 'No'),
+        ],
+        string='Show bank account',
+        default='full',
+        help="Show in invoices partial or full bank account number")
+    show_bank_account_chars = fields.Integer(
+        string="Number of chars",
```
pedrobaeza: For the sake of helping to know which field is each one when using advanced search, I would put here a more descriptive `string`: `Nº of digits for customer bank account`, and a label with only `# of digits`

pedrobaeza (changes requested):

account_payment_partner/models/account_payment_method.py
carlosdauden: I don't agree.
I think that this has more relation with payment method.
In the low likely case that someone wants to show different things using the same payment method you can always duplicate it

pedrobaeza: It's not only to show or not the bank account, but as you have added a very great feature of seeing last characters, first or full, it has to be at payment mode level for better allocation for that possibilities, or you will miss future updates on duplicated payment methods as they are maintained in `<data>` sections of the modules.

account_payment_partner/models/account_invoice.py
```diff
@@ -32,11 +32,7 @@ def _onchange_partner_id(self):
                     self.partner_bank_id =\
                         self.commercial_partner_id.bank_ids[0]
             elif self.type == 'out_invoice':
-                pay_mode = self.partner_id.customer_payment_mode_id
-                self.payment_mode_id = pay_mode
-                if pay_mode and pay_mode.bank_account_link == 'fixed':
-                    self.partner_bank_id = pay_mode.fixed_journal_id.\
-                        bank_account_id
+                self.payment_mode_id = self.partner_id.customer_payment_mode_id
```
carlosdauden, line 39: @alexis-via ¿Can you review?
I remove this because it causes a problem if you have a direct debit payment method with bank_account_link == 'fixed'
I have contemplated the original functionality and I have also created the migration script to avoid causing problems

rafaelbn (approved):
Tested functionally only :+1:

account_payment_partner/models/account_invoice.py
alexis-via: Mmm, it seems are right, my code didn't take into account this case :) Could you add a comment in the code to explain this? Because I'm afraid some people will not understand this change and look into the code... and they would find your comment that explains the problem.

account_payment_partner/models/account_payment_mode.py
```diff
@@ -0,0 +1,26 @@
...
+    show_bank_account = fields.Selection(
+        selection=[
+            ('full', 'Full'),
+            ('first', 'First n chars'),
+            ('last', 'Last n chars'),
+            ('no', 'No'),
+        ],
+        string='Show bank account',
+        default='full',
+        help="Show in invoices partial or full bank account number")
```
alexis-via, line 20: This seems completely over-kill for me. I really don't see why you would need to show only a subset of the bank account number. Maybe this is something specific to your country. account_payment_partner is a "base" module of the OCA/bank-payment project, so it is used by a lot of people, so I would prefer to keep this kind of small feature needed by a very small subset of users out of this module.

alexis-via (commented):
To sum up my position: I am OK to modify the partner_id_change, I'm OK to modify the invoice report in this module, but I think this PR needs to be simplified and add most of the fields that you propose to add to account.payment.mode should NOT be added... it's just too much complexity for something that 99% of users don't need.

account_payment_partner/models/account_payment_mode.py
pedrobaeza: @alexis-via I don't think this is too specific: this allows to print in the invoice to which account number the payment (via SEPA direct debit) is going to be charged so the customer knows that information, but there are some customers that don't want that everyone looking at the invoice sees the full account number (and even GDPR can say a word about that), so that's the reason behind the several options.

Having in another module is something oversized as the module will be here in bank-payment also, and there's not too much code about this. If you don't want to use it, just leave the default option. I promise we will maintain this code in future versions :wink:

alexis-via: Ah, ok, I understand the reason for this feature. I think it should be explained more clearly!

alexis-via: I think we have a problem with the field 'partner_bank_id 'of account.invoice for bank accounts. In the "account" module, the help message of this field says: "Bank Account Number to which the invoice will be paid. A Company bank account if this is a Customer Invoice or Vendor Refund, otherwise a Partner bank account number."
I think we cannot change the meaning of this field i.e. for a customer invoice paid by sepa direct debit, we can't store the Partner bank account number in this field. And it seems to me that it is what you are doing. So I think we need another approach: we should inherit the invoice report both in account_payment_partner and also in account_banking_mandate (in account_banking_mandate, we would display mandate_id.partner_bank_id).

BTW, that would be a good topic for during the next OCA code sprint in Sevilla (I would be ready to work on this)... but we urgently need a date for this code sprint, because my customers are already starting to book my work days in June...

pedrobaeza: Yeah, I'm pushing for having the date ASAP, but we rely on the organizer there in Sevilla for that. I'll try to push even further.

rafaelbn (approved):
Runbot is :red_circle: 

In my side this is functionally :+1:

pedrobaeza:
@carlosdauden please improve help or README and check Runbot status.

pedrobaeza:
@carlosdauden please finish this

sbidoul:
@carlosdauden since you are touching the readme, would you try with the [new readme/*.rst fragments](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)?

cubells:
@alexis-via @pedrobaeza 

can you recheck this?

pedrobaeza (approved):

pedrobaeza:
This has been long time here and all comments have been honored, so I'm merging and forward-porting to newer versions. Thanks to all.

PR #461, opened 2018-03-31
[ADD] 10.0 Account payment order date

ovnicraft:
This module allow to force accounting date in move based on scheduled date from order.
Configurable from payment mode.


pedrobaeza:
I don't understand why you need this module, as the accounting date in the transfer move is set according the "date mode" of the payment order. Please check original code.

ovnicraft:
So is not possible create payments before today, when fixed date is set to before today, system [raise a message](https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/models/account_payment_order.py#L134) so it blocks to create payments with previous date.

I read the code and yes, just in fixed mode can set fixed date for payments [account_payment_order#L266](
https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/models/account_payment_order.py#L266)

then offisetting account take this date https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/models/account_payment_order.py#L400

Prepare move method never takes  date https://github.com/OCA/bank-payment/blob/10.0/account_payment_order/models/account_payment_order.py#L373 so odoo will take today (as default value in [account.move#L83 is defined](https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_move.py#L83)

pedrobaeza:
And isn't better to modify account_payment_order for adding that option? Any way, why do you want to put a date in the past in the accounting date? That go against all the accounting principles...

ovnicraft:
That what this module does, add an option to do this.
Yes and not, sometimes payment move can be regulated on same week. but this differ in one or two days.

ovnicraft:
@pedrobaeza any updates about merge this? 

This module add an option to force date move taking it from order date.

I did a new module to give a new option.

pedrobaeza:
I still don't see this as a valid option. Let's see what others think.

ovnicraft:
@pedrobaeza look it as option as force accounting date on stock inventory. :-)

ovnicraft:
@pedrobaeza what about this ?

pedrobaeza:
@oihane @ dalonsod what do you think about this?

account_payment_date/models/payment_order.py
```diff
@@ -0,0 +1,61 @@
...
+class AccountPaymentMode(models.Model):
+    _inherit = 'account.payment.mode'
+
+    allow_force_date = fields.Boolean('Allow force accounting date')
+
+
+class AccountPaymentOrder(models.Model):
+    _inherit = 'account.payment.order'
+
+    def get_account_date(self):
```
oihane, line 18: Why are you creating this function if after this you are also controlling if allow_force_date is checked or not, you could directly take the schedule date.

ovnicraft: So module by default create entry account date when is done, so i want to give the flexibility to take the date from payment to account entry.

PR #462, opened 2018-04-04
[10.0][IMP] Refactor sepa_credit_transfert

p-tombez:
The goal is to allow overriding of line grouping (e.g. group by currency, partner, company, etc.)

### DIFF
--- a/account_banking_sepa_credit_transfer/models/account_payment_order.py
+++ b/account_banking_sepa_credit_transfer/models/account_payment_order.py
@@ -11,6 +11,19 @@ from lxml import etree
 class AccountPaymentOrder(models.Model):
     _inherit = 'account.payment.order'
 
+    @api.model
+    def _get_line_key(self, line):
+        # The field line.date is the requested payment date
+        # taking into account the 'date_prefered' setting
+        # cf account_banking_payment_export/models/account_payment.py
+        # in the inherit of action_open()
+        return (
+            line.date,
+            line.priority,
+            line.local_instrument,
+            line.category_purpose
+        )
+
     @api.multi
     def generate_payment_file(self):
         """Creates the SEPA Credit Transfer file. That's the important code!"""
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You need to change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers and check Travis error.

leemannd (approved):

ghost (approved):
LGTM

ecino (approved):
LGTM. Can you review my PR #433 in return please?

account_banking_sepa_credit_transfer/models/account_payment_order.py
```diff
@@ -11,6 +11,19 @@
 class AccountPaymentOrder(models.Model):
     _inherit = 'account.payment.order'
 
+    @api.model
+    def _get_line_key(self, line):
+        # The field line.date is the requested payment date
+        # taking into account the 'date_prefered' setting
+        # cf account_banking_payment_export/models/account_payment.py
+        # in the inherit of action_open()
+        return (
```
pedrobaeza, line 20: Better to return a list, that it's dynamic

pedrobaeza (changes requested):

yvaucher (approved):
Appart from the comment from @pedrobaeza LGTM

account_banking_sepa_credit_transfer/models/account_payment_order.py
p-tombez: The returned value is used as a ```dict``` key so it needs to be "hashable"

pedrobaeza: A list is also hashable

p-tombez:
```
>>> {[1,2,3]: 'some value'}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

jabibi: It's okay like that if the main purpose is to use as a key. You can always go from tuple to list and from list to tuple if you want to modify the values

jabibi (approved):
Code review :+1:

yvaucher:
@pedrobaeza can you update your review?

PR #464, opened 2018-04-06
[10.0][FIX] Split account move generated for SEPA direct debits per sequence_type

alexis-via:
In v8, the account move generated for SEPA direct debit was split per sequence type.
With the current code, it was not working any more because the mandates were switched from 'first' to 'recurring' upon generation of the payment file, but the move was generated later when the debit order is marked as uploaded (so, at that step, all mandates were seen as recurring).

With this PR, the mandates are switched from first to recurring AFTER generation of the account move. So the split of the account move per sequence_type is working again. And it's much better to do it at that step because, if you cancel the debit order when it is at the "generated" step, you will keep your mandates unchanged, which is a good thing.

### DIFF
--- a/account_banking_sepa_direct_debit/__manifest__.py
+++ b/account_banking_sepa_direct_debit/__manifest__.py
@@ -8,7 +8,7 @@
 {
     'name': 'Account Banking SEPA Direct Debit',
     'summary': 'Create SEPA files for Direct Debit',
-    'version': '10.0.1.1.1',
+    'version': '10.0.1.1.2',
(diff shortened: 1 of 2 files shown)


sylvainc (approved):

pedrobaeza:

bealdav (approved):

rafaelbn (approved):

pedrobaeza:
Cherry-pick for v11 in 37d58d6319fae47eaae85e4725fef7069f7303da

PR #466, opened 2018-04-13
[FIX] account_payment_order: Take payment mode from move

pedrobaeza:
When adding to a payment order from invoices, the payment mode considered is the one in the invoice, but once you have confirmed it, you can't change it, so if you change the payment mode in the journal
items, it doesn't make the difference.

With this change, this is taken into account for adding to the payment order properly.

@Tecnativa

### DIFF
--- a/account_payment_order/models/account_invoice.py
+++ b/account_payment_order/models/account_invoice.py
@@ -47,6 +47,8 @@ class AccountInvoice(models.Model):
     @api.multi
     def _prepare_new_payment_order(self, payment_mode=None):
         self.ensure_one()
+        if payment_mode is None:
+            payment_mode = self.env['account.payment.mode']
         vals = {
             'payment_mode_id': payment_mode.id or self.payment_mode_id.id,
         }
@@ -57,7 +59,6 @@ class AccountInvoice(models.Model):
     @api.multi
     def create_account_payment_line(self):
         apoo = self.env['account.payment.order']
-        aplo = self.env['account.payment.line']
         result_payorder_ids = []
         action_payment_type = 'debit'
         for inv in self:
--- a/account_payment_order/models/account_invoice.py
+++ b/account_payment_order/models/account_invoice.py
@@ -11,20 +11,13 @@ from odoo.exceptions import UserError
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
-    payment_mode_computed_id = fields.Many2one(
-        comodel_name='account.payment.mode',
-        compute="_compute_payment_mode_computed_id",
-        string="Computed payment mode",
-        help="Technical field for storing the final payment mode to use, "
-             "taking into account data in journal items.",
-    )
     payment_order_ok = fields.Boolean(
-        related="payment_mode_computed_id.payment_order_ok",
+        compute="_compute_payment_order_ok",
     )
 
     @api.depends('payment_mode_id', 'move_id', 'move_id.line_ids',
                  'move_id.line_ids.payment_mode_id')
-    def _compute_payment_mode_computed_id(self):
+    def _compute_payment_order_ok(self):
         for invoice in self:
             payment_mode = (
                 invoice.move_id.line_ids.filtered(
@@ -33,7 +26,7 @@ class AccountInvoice(models.Model):
             )
             if not payment_mode:
                 payment_mode = invoice.payment_mode_id
-            invoice.payment_mode_computed_id = payment_mode
+            invoice.payment_order_ok = payment_mode.payment_order_ok
 
     @api.model
     def _get_reference_type(self):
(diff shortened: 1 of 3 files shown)


luismontalba (approved):
Tested on runbot

account_payment_order/models/account_invoice.py
```diff
@@ -11,8 +11,29 @@
...
+    payment_mode_computed_id = fields.Many2one(
+        comodel_name='account.payment.mode',
+        compute="_compute_payment_mode_computed_id",
+        string="Computed payment mode",
+        help="Technical field for storing the final payment mode to use, "
+             "taking into account data in journal items.",
+    )
     payment_order_ok = fields.Boolean(
-        related='payment_mode_id.payment_order_ok', readonly=True)
+        related="payment_mode_computed_id.payment_order_ok",
```
ecino: Why do you remove the readonly property?

account_payment_order/models/account_invoice.py
```diff
@@ -11,8 +11,29 @@
...
+    )
+
+    @api.depends('payment_mode_id', 'move_id', 'move_id.line_ids',
+                 'move_id.line_ids.payment_mode_id')
+    def _compute_payment_mode_computed_id(self):
+        for invoice in self:
+            payment_mode = (
+                invoice.move_id.line_ids.filtered(
+                    lambda x: not x.reconciled
```
ecino: Why don't you take into account reconciled lines?

account_payment_order/models/account_invoice.py
```diff
@@ -11,8 +11,29 @@
...
+                ).mapped('payment_mode_id')[:1]
```
ecino, line 25: [0] would be more obvious no?

account_payment_order/models/account_invoice.py
```diff
@@ -48,18 +71,24 @@ def create_account_payment_line(self):
...
                     "No Payment Mode on invoice %s") % inv.number)
             if not inv.move_id:
                 raise UserError(_(
                     "No Journal Entry on invoice %s") % inv.number)
-            if not inv.payment_order_ok:
+            payment_mode = inv.move_id.mapped('line_ids.payment_mode_id')
+            if len(payment_mode) > 1:
+                raise UserError(
+                    _("You can't add invoices which journal items have more "
+                      "than one payment mode.")
```
ecino: This adds more restriction than before. I'm not sure it would make sense to check in the move lines from this method, because the user is wanting to create a payment order from the invoice. To me, here it would make sense to take the same payment mode as the one in the invoice.

ecino (commented):
I'm not sure it makes sense to take the payment mode from the move lines when we are creating payment order from the invoice. To me, this sounds confusing. Would you please give me back a review for #418 or #433 ?

account_payment_order/models/account_invoice.py
pedrobaeza: The idea is to get the payment mode for the next unreconciled item, which is the relevant one for being added to a payment order.

account_payment_order/models/account_invoice.py
pedrobaeza: No, because this can be an empty recordset and then have an error.

account_payment_order/models/account_invoice.py
pedrobaeza: This misses the unreconciled flag check also. Adding.

account_payment_order/models/account_invoice.py
pedrobaeza: Mistake by my part, re-adding

pedrobaeza:
@ecino after your comment, I have re-thought the patch, and now it's even better, adding each of the receivable/payable to a payment order if they don't have the same payment mode. Please check.

ecino:
You have a pep8 error

account_payment_order/models/account_invoice.py
```diff
@@ -31,9 +45,11 @@ def line_get_convert(self, line, part):
         return res
 
     @api.multi
-    def _prepare_new_payment_order(self):
+    def _prepare_new_payment_order(self, payment_mode=None):
         self.ensure_one()
-        vals = {'payment_mode_id': self.payment_mode_id.id}
+        vals = {
+            'payment_mode_id': payment_mode.id or self.payment_mode_id.id,
```
ecino: This could raise error if payment_mode is None

pedrobaeza:
@ecino Fixed! Thanks for the review.

ecino (approved):

ecino:
@pedrobaeza I think it's good now. Would you mind checking one of my PRs #418 or #433? Thanks!

pedrobaeza:
Merging and cherry-picking this in v11.

pedrobaeza:
Cherry-picked in f8939814a4e1ec17d7a749af00db2e8fcbd2d91b

PR #467, opened 2018-04-14
[10.0][FIX] Don't store acc_type of res.partner.bank

alexis-via:
We have a problem due to the fact that account_payment_mode adds store=True on the native computed field acc_type of res.partner.bank
This is particularly visible when you add the module on a database with existing accounts or when you migrate from v8 to v10 for example: even if you have the module base_iban, acc_type = 'bank' on IBAN accounts. This seems to be linked to the fact that when Odoo computes+stores the initial value of the field, it doesn't take into account the inherits of the native compute method, so it will not take into account the inherit in base_iban.

There are maybe better ways to fix this (add a search method ?) ; this PR is a first proposal to address this important problem.

### DIFF
--- a/account_payment_mode/__manifest__.py
+++ b/account_payment_mode/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Account Payment Mode',
-    'version': '10.0.1.0.1',
+    'version': '10.0.1.0.2',
     'license': 'AGPL-3',
     'author': "Akretion,Odoo Community Association (OCA)",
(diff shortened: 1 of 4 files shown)


sbidoul (approved):
Thanks!

pedrobaeza (approved):

pedrobaeza:
Cherry-pick for v9: f9d38f280609995c1d305efc2def9bf14d2b6cee
Cherry-pick for v11: a8d031b0fd505037bda1880a3dfaf37efbda7826

PR #469, opened 2018-04-20
[IMP] account_payment_order: New option to allow negative transactions

carlosdauden:
@Tecnativa


pedrobaeza:
I think the problem is the rest of the logic, that is not prepared to handle positive and negative amounts. Have you check it? You should add explicit tests for negative amounts with this for being sure.

pedrobaeza:
@carlosdauden any news?

@alexis-via what do you think about this?

sbidoul:
Could you elaborate the use case?

carlosdauden:
Invoice 100 €
Refund   40 €
Bank payment line 60 €

3 move lines conciled without additional steps

carlosdauden:
If it's interesting I will do the tests

sbidoul:
In this case, the workflow should be a partial reconcile of the invoice and refund, then a payment of the residual. I'm not comfortable at all with allowing negative payment lines.

pedrobaeza:
Why not, Stéphane?

sbidoul:
Is it correct it only makes sense when the grouping of payment lines is in effect and the resulting bank line is positive? Then you must test that too. 

That just sounds quite complex for little benefit IMO.

pedrobaeza:
Thanks for the insights. Indeed this requires more work for avoiding incorrect orders, but there are several advantages to include this feature (following the example of @carlosdauden):

* You perform all the steps at once, avoiding errors or omissions, and having a correct flow. If not, you can have a payment order for 100 € (the full invoice) without noticing the 40 € refund, and then got refused - ERROR -.
* If you want to avoid this, you have to go refund per refund checking for possible reconciliations (or review all unreconciled elements). On automatic flows like sales invoicing, where you don't see the created refunds (if there are returns after the initial invoicing), there's not the possibility of seeing in live when creating the refunds.
* On the communication of the payment order, you only see the invoice number, not the refund one, and the amount is lesser than the invoice total, so this generates questions from the customers about what is this paying/debiting.

With all of this, I think it's worth to work and make it a bit more complex (not too much any way).

pedrobaeza:
@sbidoul @alexis-via any additional comment?

sbidoul:
I still believe this feature is dangerous and requires more checks. Eg disabling it unless grouping is enabled, and raising error in case of negative values in resulting lines.

pedrobaeza:
OK, we can work on that things, but what I want to know is that you agree on the principle of adding the feature. @carlosdauden please work on the constraints, and add A LOT OF TESTS, hehe.

sbidoul:
Yes I agree on the principle. Input from @alexis-via would be nice too.

carlosdauden:
Closed in favour of https://github.com/OCA/bank-payment/pull/526

Issue #398, opened 2017-10-11
Migration script error for 8.0 to 9.0


hmartinet:
OpenUpgrade from 8.0 to 9.0. 

The problem is in my database column 'payment_order_type' does not exists, but the module account_payment_order is add by a dependency with no warranty that the field exists.

2017-10-11 10:27:48,613 31201 INFO oe_lpn_7_0_5_0_migrated openerp.modules.migration: module account_payment_order: Running migration [9.0.1.0.0>] post-migration
2017-10-11 10:27:48,614 31201 INFO oe_lpn_7_0_5_0_migrated OpenUpgrade: account_payment_order: post-migration script called with version 8.0.1.1
2017-10-11 10:27:48,614 31201 INFO oe_lpn_7_0_5_0_migrated openerp.sql_db: Programming error: column "payment_order_type" does not exist
LIGNE 3 :                        WHERE payment_order_type = 'payment'
                                       ^
, in query UPDATE account_payment_order
                       SET payment_type = 'outbound'
                       WHERE payment_order_type = 'payment'
2017-10-11 10:27:48,614 31201 ERROR oe_lpn_7_0_5_0_migrated OpenUpgrade: account_payment_order: error in migration script account_payment_order/migrations/9.0.1.0.0/post-migration.py: column "payment_order_type" does not exist
LIGNE 3 :                        WHERE payment_order_type = 'payment'
                                       ^

2017-10-11 10:27:48,615 31201 ERROR oe_lpn_7_0_5_0_migrated OpenUpgrade: column "payment_order_type" does not exist
LIGNE 3 :                        WHERE payment_order_type = 'payment'
                                       ^
Traceback (most recent call last):
  File "[...]/workdir/openupgradelib/openupgradelib/openupgrade.py", line 1250, in wrapped_function
    if use_env2 else cr, version)
  File "[...]/workdir/9.0/bank-payment/account_payment_order/migrations/9.0.1.0.0/post-migration.py", line 27, in migrate
    map_payment_type(cr)
  File "[...]/workdir/9.0/bank-payment/account_payment_order/migrations/9.0.1.0.0/post-migration.py", line 15, in map_payment_type
    table='account_payment_order', write='sql')
  File "[...]/workdir/openupgradelib/openupgradelib/openupgrade.py", line 1024, in map_values
    logged_query(cr, query, values)
  File "[...]/workdir/openupgradelib/openupgradelib/openupgrade.py", line 746, in logged_query
    cr.execute(query, args)
[...]

PR #471, opened 2018-04-23
[FIX] column "payment_order_type" does not exist during migration to 9.0

metaminux:
#398

### DIFF
--- a/account_payment_order/migrations/9.0.1.0.0/post-migration.py
+++ b/account_payment_order/migrations/9.0.1.0.0/post-migration.py
@@ -24,8 +24,8 @@ def map_payment_type(cr):
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     cr = env.cr
-    if openupgrade.is_module_installed(
-            env.cr, 'account_banking_payment_export'):
+    if openupgrade.table_exists(cr, 'account_payment_order') \
+       and openupgrade.column_exists(cr, 'account_payment_order', 'payment_order_type'):
         map_payment_type(cr)
 
     cr.execute("""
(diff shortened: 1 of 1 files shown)


metaminux:
It seems to me that "payment_order_type" is only present in "account_banking_payment_export" module, so I added a test on its state before calling the "map_payment_type" method.

StefanRijnhart (approved):
Thanks!

pedrobaeza:
I will prefer to test if the column exists. You have the method in openupgradelib called `column_exists`

pedrobaeza:
Any news about this?

metaminux:
Yes, I wanted to try "column_exists" version on a true migration.

I'll change my last commit in a few days I think.

account_payment_order/migrations/9.0.1.0.0/post-migration.py
```diff
@@ -24,7 +24,9 @@ def map_payment_type(cr):
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     cr = env.cr
-    map_payment_type(cr)
+    if openupgrade.table_exists(cr, 'account_payment_order') \
+       and openupgrade.column_exists(cr, 'account_payment_order', 'payment_order_type'):
```
pedrobaeza: Incorrect indentation an line very long (>80 chars)

metaminux: @pedrobaeza Could you please point me to an OCA guideline on how to break line in too long if statement, cause PEP8 is not really strict on that point (see https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator )

Thx

pedrobaeza: Indeed here there's no strict guideline, except complying with that 2 rules (indentation on second line of the if should be double and not > 80 cols). My proposal in this case is:

```python
    table = 'account_payment_order'
    if (openupgrade.table_exists(cr, table) and
            openupgrade.column_exists(cr, table, 'payment_order_type')):
```

metaminux: OK, done

pedrobaeza (approved):

PR #472, opened 2018-04-23
[10.0] Edit due date on SDD operations. Ofter SDD operations are not advance…

jackjack82:
… payments from bank but turn the invoice credit into credit to be received from bank. For this reason, the due date of the original move line shall be kept.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -401,6 +401,11 @@ class AccountPaymentOrder(models.Model):
         else:
             vals.update({'date_maturity': bank_lines[0].date})
 
+        if (len(bank_lines[0].payment_line_ids) == 1 and
+                self.date_prefered == 'now'):
+            m_date = bank_lines[0].payment_line_ids[0].ml_maturity_date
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please explain the use case, as if this is already covered if you select due date in the date selection.

jackjack82:
Sir, given an invoice with a certain deadline, the flow is the following (with accounting entries):

1. SDD created and accepted by the bank, account move registered "immediately":
credit vs customer advanced by bank <> credit vs customer (the invoice is closed immediately)

2. bank pays money in advance for all the invoice lines in the SDD:
cash/bank account <> debit account advanced money from bank

3. client pays the bank
debit account advanced money from bank <> credit vs customer advanced by bank

The account move at step 1 is registered immediately. However, the client still wants to know the original payment deadline. When grouping payment deadlines by account he sees all "advanced" credits and their aging.

The configuration "due date" sets correctly the due date BUT also the account move date, which is not the case.

StefanRijnhart:
I would appreciate a test for this to demonstrate the problem and prevent an accidental revert in the future.

PR #473, opened 2018-04-26
[FIX] multicompany management of account_orders

etobella:
This PR allows to create and execute an order from another company if you are working on a multicompany environment


etobella:
The problem was related to another issue. Sorry for the misunderstanding.

PR #474, opened 2018-05-03
[FIX] only drop column if it exists

hbrunn:
without this, the migration breaks in databases where the field never was stored


pedrobaeza (approved):

pedrobaeza:
We can fast-track this

PR #477, opened 2018-05-08
[FIX] report should show the invoice number, not the move name

etobella:
When I printed the report, I found that it shows the move name, but in some countries, the move name is not the same that the invoice number.
I think it is easier to show the invoice number.

### DIFF
--- a/account_payment_order/report/account_payment_order.xml
+++ b/account_payment_order/report/account_payment_order.xml
@@ -76,7 +76,7 @@
                                 <span t-esc="get_bank_account_name(line.partner_bank_id)"/>
                             </td>
                             <td class="text-center">
-                                <span t-raw="'%s &lt;br&gt;' % line.move_line_id.move_id.name if line.move_line_id.move_id else ''"/>
+                                <span t-raw="'%s &lt;br&gt;' % line.move_line_id.invoice_id.number or line.move_line_id.move_id.name or ''"/>
                             </td>
                             <td class="text-center">
                                 <span t-field="line.date"/>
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

sbidoul (dismissed):
IIRC invoice_id is optional so there should be a fallback on the move name if it is absent

etobella:
@sbidoul Now it will show the invoice number if there is an invoice, otherwise, it will show the move name.

JordiBForgeFlow (approved):

account_payment_order/report/account_payment_order.xml
```diff
@@ -76,7 +76,8 @@
                                 <span t-esc="get_bank_account_name(line.partner_bank_id)"/>
                             </td>
                             <td class="text-center">
-                                <span t-raw="'%s &lt;br&gt;' % line.move_line_id.move_id.name if line.move_line_id.move_id else ''"/>
+                                <span t-raw="'%s &lt;br&gt;' % line.move_line_id.invoice_id.number" t-if="line.move_line_id.invoice_id"/>
+                                <span t-raw="'%s &lt;br&gt;' % line.move_line_id.move_id.name if line.move_line_id.move_id else ''" t-if="not line.move_line_id.invoice_id"/>
```
sbidoul: `line.move_line_id.invoice_id.number or line.move_line_id.move_id.name or ''` should work too and is maybe slightly easier to read.

sbidoul (approved):

pedrobaeza:
Cherry-picked for v10 in cec3b4aafc0e7c9680c76d8090f3ae62ed99e2f8 (v9 doesn't have this report).

PR #479, opened 2018-05-22
[10.0] Several small usability improvements

alexis-via:
- when manually encoding a payment line, get company currency as default currency
- Search on payment order numbers
- Direct search on bank journal name in payment orders
- remove widget="selection" on account.payement.mode : this object is rarely modified, so widget='selection' is not a time saver, but we miss the HTML link, which can be pretty convenient to check the configuration.


yvaucher (approved):

sbidoul (approved):

pedrobaeza:
Cherry-picked for:

* v9: d300de48235d4297fc214fee06acc5de05530545
* v11: 620ff30277df9d58441ddf906802cba240fc89b9

PR #483, opened 2018-06-06
11.0 mig portal payment mode

sergio-teruel:
cc @Tecnativa


sergio-teruel:
This modulo is not necessary. The account.invoice report in website is renderized with sudo

PR #482, opened 2018-06-06
10.0 mig portal payment mode

sergio-teruel:
cc @Tecnativa

PR #484, opened 2018-06-11
[10.0][IMP] account_payment_partner: Add security ACL por portal users.

sergio-teruel:
This PR close this https://github.com/OCA/bank-payment/pull/482
Group portal is defined in base module, so this module must be deleted.


sergio-teruel:
This modulo is not necessary. The account.invoice report in website is renderized with sudo

PR #487, opened 2018-06-13
[IMP] improve file it.po

CiroBoxHub:


pedrobaeza:
The same about Transifex.

CiroBoxHub:
ok

PR #490, opened 2018-07-02
Adding extra SEPA fields: Purpose & Schema

andreparames:
This commit adds two standard SEPA fields to the generator, the Purpose Code and the Creditor Schema:

https://wiki.xmldation.com/Support/Nordea/Purpose_Code

https://wiki.xmldation.com/Support/EPC/CdtrSchmeId


pedrobaeza:
CdtrSchmeId seems to be only for France and England.

PR #491, opened 2018-07-02
Adding extra SEPA fields: Purpose & Schema

andreparames:
This commit adds two standard SEPA fields to the generator, the Purpose Code and the Creditor Schema:

https://wiki.xmldation.com/Support/Nordea/Purpose_Code

https://wiki.xmldation.com/Support/EPC/CdtrSchmeId


pedrobaeza:
It seems CdtrSchmeId is only for France and England, isn't it?

It's better to have each field addition on one commit for better isolation.

You have to check PEP8 errors:
```
account_payment_order/models/bank_payment_line.py:62:13: E126 continuation line over-indented for hanging indent
account_payment_order/models/bank_payment_line.py:71:9: E121 continuation line under-indented for hanging indent
```

andreparames:
> It seems CdtrSchmeId is only for France and England, isn't it?

I  believe those are just the languages in which that page is translated; the field is, as far as I can tell, part of the standard SEPA format, and at least some banks and companies from outside those countries use it.

> It's better to have each field addition on one commit for better isolation. You have to check PEP8 errors

Thanks, I'll restructure and push again.

sbidoul:
@andreparames could you look into the travis errors: it seems there is a schema validation error.

account_banking_pain_base/models/account_payment_mode.py
```diff
@@ -27,3 +27,8 @@ class AccountPaymentMode(models.Model):
         "- Country code (2, optional)\n"
         "- Company idenfier (N, VAT)\n"
         "- Service suffix (N, issued by bank)")
+    initiating_party_schema = fields.Char(
+        string='Initiating Party Schema', size=35,
+        help="This will be used as the 'Initiating Party Schema Name' in "
+        "the PAIN files generated by Odoo. If not defined, no schema will "
+        "be used.\n")
```
sbidoul: It would be nice to have some more help (such as example values, and possibly a pointer to the specification).

account_payment_order/models/bank_payment_line.py
```diff
@@ -58,6 +58,17 @@ class BankPaymentLine(models.Model):
...
+        ("SALA", "Salary payment"),
+        ("PENS", "Pension payment"),
+        ("CBFF", "Capital-forming payment (VWL)"),
+        ("CBFR", "Capital-forming payment for retirement (AVWL)"),
+        ("SSBE", "Social Security Benefit"),
+        ("INTC", "ICC payment"),
+        ("CASH", "Cash Management transfer"),
+        ("SUPP", "Other payment"),
+        ("NETT", "Netting"),
+    ], string="Purpose Code")
```
sbidoul: Don't you need to add this purpose code on `account.payment.line`, and use it as [grouping criteria](https://github.com/OCA/bank-payment/blob/9d0c6e1c3d673b8eb8040162158fef7dd561083b/account_payment_order/models/bank_payment_line.py#L62) to generate the `bank.payment.line`? Indeed, users don't normally generate bank payment lines manually.

andreparames:
I've improved the help description, but I couldn't find a good link to point people to.

sbidoul (approved):
Looks good (code review)

account_payment_order/models/bank_payment_line.py
```diff
@@ -58,6 +58,17 @@ class BankPaymentLine(models.Model):
         'res.currency',
         related='order_id.payment_mode_id.company_id.currency_id',
         readonly=True, store=True)
+    purpose_code = fields.Selection(selection=[
```
pedrobaeza: Instead of repeating the same selection, use the same trick as we do in sale_order_type:

https://github.com/OCA/sale-workflow/blob/e1149e3f39de09cf1ce5094f9ba1845d3e43eadd/sale_order_type/models/sale_order_type.py#L15-L18

PR #493, opened 2018-07-05
[ADD] Allow for multiple SEPA payment methods with different versions

StefanRijnhart:
Currently, I cannot have multiple outbound SEPA payment methods, which is unfortunate if one has multiple banks that support different SEPA versions. The culprit is this constraint: https://github.com/OCA/bank-payment/blob/9.0/account_payment_mode/models/account_payment_method.py#L33. I am proposing to extend that constraint in this module to pain_version in the unicity check.


pedrobaeza (approved):

alexis-via (approved):

StefanRijnhart:
Thanks for the reviews. Ported to later branches https://github.com/OCA/bank-payment/pull/496 and https://github.com/OCA/bank-payment/pull/497

PR #498, opened 2018-07-31
[ADD] Account Holder

BT-dherreros:
Let's say you have a contract with a couple but the payment goes on the account of the wife, we must have a separated field with the account holder on the bank account.


pedrobaeza:
I don't think the account holder concept should be introduced here. Any way, if I don't remember bad, there's already a char field for the bank account owner.

BT-astauder:
@pedrobaeza I have not found a char field for the bank account owner. At least not on res.partner.bank . Where do you suggest to set the account owner? We do have problems that transactions are refused when the bank account holder is not exactly the same as we set in the payment file.

pedrobaeza:
OK, I see, but I don't get why you should have this field. Just create another partner with that bank account. This overhead is too much for that corner case IMO. What do you think, @sbidoul and @alexis-via ?

BT-astauder:
Creating another partner was also our first approach. But it must be the same commercial partner otherwise the partner ledger would be wrong. 

Second approach was to set the couple as a company with name e.g. "Mr. and Mrs. Smith". The wife could be then a contact in this company and you would be able to create a invoice in the name of the wife. For the payment this unfortunately does make a differences, as bank accounts are only on the company possible and never on the contacts (what is absolutely correct from a juristic point of view).

From a juristic point of view, the problem we are solving here is, that we have a contract with someone. We sent the invoice to him. But the person can decide to pay someone else. The couple that decides to pay on the bank account of the wife is just a common example.
Even we payed on the account of the wife, we still do not contract just with the wife. This can easily be seen in the case we overpaid and want our money back. Then we can not only claim the money from the wife, but also from her husband. The reason for that is, that the contract we have is with the couple. The couple decided to pay on the wife's account, so the couple is liable for the payback of the overpayment.

Maybe this was not that important before payment with SEPA. But now banks seem to compare the account with the account holder at least if the transfer exceed a certain amount.

account_banking_pain_base/models/res_partner_bank.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import api, fields, models, _
+
+
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    @api.model
+    def _default_account_holder(self):
+        return self.partner_id.name
```
sbidoul: Would it make sense to leave it empty by default?

sbidoul:
What I miss in this PR is how this new field would be actually used in SEPA file generation code.

Would it make sense to allow payment to a bank account of another partner? At first glance this [restriction we have now](https://github.com/OCA/bank-payment/blob/ff5efa0f20d7acaff653fba751c4c5116b614718/account_payment_order/views/account_payment_line.xml#L22) is there for user comfort and avoiding errors in the most common case, but it could be lifted in some circumstances. Perhaps with a check box to explicitly allow payment to/from an account of another partner?

pedrobaeza:
I prefer this option instead of adding the holder concept.

account_banking_pain_base/models/res_partner_bank.py
BT-astauder: This here should be removed as it is not working anyway. The default is set in views/res_partner.xml by default_account_holder in context.

alexis-via:
I also think this is a very corner-case that doesn't deserve a modification of the datamodel of the generic OCA modules.

BT-astauder:
@sbidoul The account_holder is set in account_banking_pain_base/models/account_payment_order.py

We have now two independent project where clients complain about refused payments by the bank because of mismatched account holders. This may be connected with higher regulations in terms of sanctions and money laundry. So this is no longer a corner case for me but daily business.

The domain currently set on the invoice @sbidoul mentioned was also an approach we thought about. But there are two arguments against that in my eyes: 

First of all this domain is very handy. Normally you only have one bank account and with this it selects automatically the correct one. So this domain is very handy. To remove it, only for the reason to not have an additional field makes a easy problem complex.
For the second argument think about this situation: Company A you have a contract with is bought by company B (fully integrated, not a subsidiary). Company B will not keep the name A. But that does not mean that they go now and change all contract from A to name B. So your contract is still on name A, so you have to keep the partner name A. But the bank account will be changed to B. 
The suggested solution to lift the domain means now, that you create partner B and move the bank account from A to B (I don't know if you even can do that in the current UI). With every invoice you sent to partner A you have to remember (there is no reference in Odoo) to set the bank account of B and you have to change it manually.

I absolutely share your hesitations to have a name field on the bank account as it does not look like a clean Odoo way. But in the end all solutions we thought about are much more complex in terms of user experience.

I also have to mention a drawback of our solution: Because the account is not connected to the "correct" partner (in my case e.g. the couple instead the wife) it is possible that when I contract only with the wife I have to have the same bank account twice in Odoo. This in fact is not very nice.

gody1001:
Add fleet business inc.bankfleet

BT-astauder:
Related: https://github.com/odoo/odoo/issues/26000

BT-astauder:
@gody1001 What do you mean? Can you provide a link for this addon?

BT-astauder:
Related: https://github.com/odoo/odoo/issues/19754

BT-dherreros:
Hi @pedrobaeza , runbot fails with some errors which I guess are the ones in:
https://runbot.odoo-community.org/runbot/static/build/3338652-498-9cd67f/logs/job_10_test_base.txt
related to package.json. Could you confirm this, and if so let me know how to fix them?
Also, the CLA is unreachable, I get a 404 when I try to access the link set by the bot.

pedrobaeza:
@gurneyalex @sbidoul can you please correct the link to CLA as on v11 the proper one is http://odoo-community.org/page/cla (the other one is not valid as it contains `website.` part).

@BT-dherreros check my updated link, and the runbot fail is this warning:
```
2018-09-18 11:17:55,457 166 WARNING openerp_test odoo.fields: Field account.banking.mandate.display_name depends on itself; please fix its decorator @api.depends().
```

BT-astauder:
@pedrobaeza, @sbidoul, @alexis-via, @gody1001 Would you be so kind to reconsider this PR after my (longly) argumentation? I also put this on the wishlist for odoo 13.0, but don't want to wait till then... ;)

account_banking_pain_base/post_install.py
```diff
@@ -3,9 +3,18 @@
...
+logger = logging.getLogger(__name__)
 
 
 def set_default_initiating_party(cr, registry):
+    logger.info('Launching the post init hook')
+    cr.execute("""
+        UPDATE res_partner_bank SET account_holder =
+        (SELECT name FROM res_partner WHERE res_partner.id =
+        res_partner_bank.partner_id);
+        """)
```
sbidoul: IMO account_holder should be kept optional

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -373,7 +373,7 @@ def generate_party_block(
             party_type_label = 'Creditor'
         elif party_type == 'Dbtr':
             party_type_label = 'Debtor'
-        name = 'partner_bank.partner_id.name'
+        name = 'partner_bank.account_holder'
```
sbidoul: So this should end up being something like `partner_bank.account_holder or partner_bank.partner_id.name`

sbidoul (commented):
I'm still not convinced but I made a couple of remarks.

account_banking_pain_base/models/res_partner_bank.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import api, fields, models
+
+
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    account_holder = fields.Char(default=' ')
+
+    @api.onchange('partner_id')
+    def _onchange_account_holder(self):
```
pedrobaeza: This should be named as `_onchange_partner_id_ account_banking_pain_base`

account_banking_pain_base/migrations/11.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 Brain-Tec Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def set_account_holder(cr):
```
pedrobaeza: Don't duplicate code. Import post_init_hook method. Any way, I also agree this shouldn't be filled by default.

account_banking_pain_base/models/account_payment_order.py
BT-astauder: Thank you very much for the review.

The idea here was to fill account_holder with the hook (for migration and new installations) and with default (for new bank accounts created when addon is installed). So by default in all situations it account_holder has the same values as partner_id.name.
Like that, when you do not care at all about account_holder, the behaviour should be exactly the same as before.
But we can add your suggestion to have a fallback in case someone puts an empty string in the field account_holder.

pedrobaeza: Why duplicating data in all cases when 99% of them will be the same? Please keep it empty while not equal.

account_banking_pain_base/models/res_partner_bank.py
BT-dherreros: Why should it be this name? isn't it too long?

account_banking_pain_base/migrations/11.0.2.0.0/post-migration.py
BT-dherreros: I think the best way is first to rename the post_install file to hooks which is the name it should have following the guidelines. Then, as you said, import and use the method in the post-migration file.

pedrobaeza: But as said, you shouldn't fill the same data, so remove all. My comment is only "didactic"

account_banking_pain_base/models/res_partner_bank.py
pedrobaeza: Guidelines:

* Onchange name should be contain the field name of the field you are hooking it.
* It shouldn't be the general one (`_onchange_parner_id`), as other module can declare the same and choke both (only one will be called, which one? you don't know). The only way to assure your onchange name is unique is adding the module name to the usual method name.
* This is only needed when you are not the module that declares the field (which is the case). If you are declaring the field, no brother module can declare an onchange on your field without depending on your module.

BT-dherreros: Thanks @pedrobaeza, I will change the name right away.

BT-astauder:
@pedrobaeza I think we do have different approaches. We thought to use the account holder field as the single source for the SEPA file. That's why we copied all data there.

When I understand your comments correct, you see it more as an "Deviating account holder" that is set only when different to the situation today. In that case I would suggest we add a boolean the user has to check to make the "Deviating Account Holder Name" field visible. This is more user friendly than just keeping the field empty in case you want the name from res.partner.

Your solution has the advantage that changing the name in res.partner also changes the name in the bank account (as it does now) and only in the exceptional case the checkbox is marked, the account holder will be not changed.

Do you agree when we implement it like that?

pedrobaeza:
Well, for me adding a new boolean field is not needed as it's an unneeded overhead. Just put a proper placeholder on the new field saying something like "Fill holder if different from partner". @sbidoul what do you think?

sbidoul:
@BT-astauder I think the bookean is unncessary if the account holder field has a help text saying to populate it only when different from the linked partner.

BT-dherreros:
@sbidoul I've added help text that indicates that the account_holder field should be fill if the name of the holder and partner differ. I think is the best approach if this information is optional.

account_banking_pain_base/models/account_payment_order.py
sbidoul: This one remains to be attended.

pedrobaeza:
And the removal of the init hook.

account_banking_pain_base/hooks.py
```diff
@@ -1,22 +0,0 @@
...
-
-
-def post_init_hook(cr, registry):
-    logger.info('Launching the post init hook')
-    cr.execute("""
-        UPDATE res_partner_bank SET account_holder =
-        (SELECT name FROM res_partner WHERE res_partner.id =
-        res_partner_bank.partner_id);
-        """)
-    with api.Environment.manage():
```
pedrobaeza: You only have to remove your part, not the previous one, which is still needed.

pedrobaeza:
Ouch, now we need to remove the default context key and the onchange due to the change on the concept.

Sorry for not saying everything together.

BT-dherreros:
I'm having an issue with runbot/travis. https://github.com/BT-dherreros/bank-payment/blob/cf8795d36dd1864360d42c3ac8844892db59176d/account_banking_sepa_direct_debit/models/account_banking_mandate.py#L39 line fails the lint check on travis if it doesn't have the preceding underscore. Yet if the underscore is there runbot says that display_name shouldn't depend on itself and to check the api.depends(). It passed the checks this morning, why is it failing now?

pedrobaeza:
I think it can be an incorrect warning. Any way, it has appeared when you have renamed the method from `compute_display_name` to `_compute_display_name`, if you want to leave it as it was previously, the warning appeared in Travis is not blocking.

BT-dherreros:
> I think it can be an incorrect warning. Any way, it has appeared when you have renamed the method from `compute_display_name` to `_compute_display_name`, if you want to leave it as it was previously, the warning appeared in Travis is not blocking.

It must stay as _compute_display_name to follow the guideline.

>Compute Field: the compute method pattern is _compute_<field_name>

How should I proceed to avoid the runbot failure?

pedrobaeza:
As said, it seems an incorrect warning. If you want to reset that warning, try to restore the compute without the `_`, although it goes against guidelines.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -373,7 +373,7 @@ def generate_party_block(
             party_type_label = 'Creditor'
         elif party_type == 'Dbtr':
             party_type_label = 'Debtor'
-        name = 'partner_bank.partner_id.name'
+        name = 'partner_bank.account_holder' or 'partner_bank.partner_id.name'
```
sbidoul: This is not correct. Should rather be:

```python
if partner_bank.account_holder:
    name = 'partner_bank.account_holder'
else:
    name = 'partner_bank.partner_id.name'
```

gody1001:
Money

BT-dherreros:
@pedrobaeza , @sbidoul can you point to a documentation about codecov? I wan't to fix those errors but this is the first time I've dealt with it

pedrobaeza:
This is because now there's an untested line (you can see it clicking on "Details"). Don't worry too much about it, but if you want to avoid that without developing more tests, there's a trick using a one line ternary operator (although this is cheating, hehe):
```python
name = ('partner_bank.account_holder'
        if partner_bank.account_holder
        else 'partner_bank.partner_id.name')
```

BT-dherreros:
Finally all checks have passed, should I take care of other changes before the possible merge?

account_banking_pain_base/models/res_partner_bank.py
```diff
@@ -0,0 +1,17 @@
...
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    account_holder = fields.Char(default=' ',
+                                 help='Fill this field if the account'
+                                      ' holder\'s name differ from the'
+                                      ' partner name')
+
+    @api.onchange('partner_id')
+    def _onchange_partner_id_account_banking_pain_base(self):
```
pedrobaeza: You should remove this onchange as said that no duplication of data is needed

account_banking_pain_base/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
...
+<odoo>
+
+    <record id="partner_view_buttons" model="ir.ui.view">
+        <field name="name">partner.view.buttons.form</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="account.partner_view_buttons"/>
+        <field name="arch" type="xml">
+            <xpath expr="//button[field[@name='bank_account_count']]" position="attributes">
+                <attribute name="context">
+                    {'search_default_partner_id': active_id, 'default_partner_id': active_id, 'default_account_holder': name}
```
pedrobaeza: Don't fill by default `account_holder`. In fact, remove all this file as it doesn't add anything meaningful.

pedrobaeza (changes requested):
Thanks any way for the cleaning of warnings.

BT-dherreros:
I've made the requested changes but the lint check failed again at the infamous _compute directive. I hadn't touch it for the last 2 commits, why is it failing again?

pedrobaeza:
That's not the problem. The problem is:
```
account_banking_pain_base/models/res_partner_bank.py:4:1: F401 'odoo.api' imported but unused
```

BT-dherreros:
@pedrobaeza the requested changes were added on my last commit to this PR

pedrobaeza (approved):

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
lmignon, line 3: post_install and at_install must be removed

account_banking_sepa_direct_debit/tests/test_sdd.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
```
lmignon, line 1: post_install and at_install must be removed

lmignon (changes requested):

pedrobaeza:
@lmignon this is out of the scope of this PR.

account_banking_pain_base/models/res_partner_bank.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2018 Brain-Tec Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    account_holder = fields.Char(default=' ',
```
sbidoul: why `default=' '`?

sbidoul (commented):
One last remark

BT-dherreros:
@sbidoul I've checked if a default was still necessary and it wasn't. I removed it.

sbidoul:
Ok, now good for me. Thanks for the work.

Now ideally, to facilitate forward and back-ports, the commits should be reworked to have 3
- one for the new feature
- one for the lint changes (ie removed utf-8 headers)
- one for the methods and hook renamings

PR #499, opened 2018-08-21
[FIX] [account_payment_order] set payment_type for payment modes with newly assigned payment method

hbrunn:
without this, the related field stays empty and the payment modes in question aren't available on payment orders


pedrobaeza (approved):
I was wondering why this happens to me in some databases, but not other. I think it's due to the old related fields issue you reported some time ago. Any way, with this we make sure they are filled.

Fast-tracking following OpenUpgrade policy.

Issue #495, opened 2018-07-24
[9] Payment order creation fails with 'structured'


Martronic-SA:
I'me getting this error:

> bank-payment-9.0/account_payment_order/models/account_move_line.py", line 34, in _prepare_payment_line_vals
>     ref2comm_type[self.invoice_id.reference_type]
> KeyError: u'structured'

It seems that the 'structured' reference type is added by account_payment_order module,
but there is not correspondance between reference_type 'structured' on invoice and communication type on payment line given by the metchod account_payment_line.py:

>     def invoice_reference_type2communication_type(self):
>         """This method is designed to be inherited by
>         localization modules"""
>         # key = value of 'reference_type' field on account_invoice
>         # value = value of 'communication_type' field on account_payment_line
>         res = {'none': 'normal'}
>         return res

Do I miss the module that is supposed to add the structured communication type?

Thanks in advance

pedrobaeza:
Can you make the PR fixing the problem?

PR #514, opened 2018-10-12
Fix structured communication

Martronic-SA:
Adds 'structured' to the invoice_reference_type2communication_type ans the 'structured' type is added by the module itself in account_invoice but breaks the account_move_line.py as explained here: https://github.com/OCA/bank-payment/issues/495

### DIFF
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -127,7 +127,7 @@ class AccountPaymentLine(models.Model):
         localization modules"""
         # key = value of 'reference_type' field on account_invoice
         # value = value of 'communication_type' field on account_payment_line
-        res = {'none': 'normal'}
+        res = {'none': 'normal', 'structured': 'structured'}
         return res
 
     @api.multi


Martronic-SA:
If it's ok, I can make the same for Odoo v 10.0 ?

pedrobaeza:
Isn't better to get the same values as you have in the origin querying the field selection? (something similar to https://github.com/OCA/sale-workflow/blob/32dd8662015262184ee62cbe83fbafabe1481906/sale_order_type/models/sale_order_type.py#L17)

sbidoul:
@pedrobaeza here it's mapping one selection to another, so your example does not seem to apply.

sbidoul (approved):
(code review)

pedrobaeza (approved):
You're right, @sbidoul merging and cherry-picking

pedrobaeza:
v10: 6d99bff7dfd7649fc49c24b3cde2be1f2d671666
v11: af07e8e2

Martronic-SA:
Perfect! thank you!

PR #516, opened 2018-10-24
[11.0][MIG] account_payment_line_cancel

MaxyMoos:
Small modification added to the 10.0 branch code in order to add a message with all the info about a cancelled payment line in the payment order when cancelling a payment line that is not linked to an invoice.

### DIFF
--- a/account_payment_line_cancel/models/__init__.py
+++ b/account_payment_line_cancel/models/__init__.py
@@ -1,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import account_payment_line
+from . import  invoice
--- a/account_payment_line_cancel/models/__init__.py
+++ b/account_payment_line_cancel/models/__init__.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from . import account_payment_line
-from . import  invoice
+from . import bank_payment_line
+from . import invoice
--- a/account_payment_line_cancel/models/bank_payment_line.py
+++ b/account_payment_line_cancel/models/bank_payment_line.py
@@ -1,0 +1,21 @@
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+from odoo.addons.account_payment_order.models.bank_payment_line \
+    import BankPaymentLine
+
+
+class APLCBankPaymentLine(models.Model):
+    _inherit = "bank.payment.line"
+
+    @api.multi
+    def unlink(self):
+        if not self.env.context.get('force_unlink'):
+            for line in self:
+                order_state = line.order_id.state
+                if order_state == 'uploaded':
+                    raise UserError(_(
+                        "Cannot delete a payment order line whose payment "
+                        "order is in state '{}'. You need to cancel it "
+                        "first.").format(order_state))
+        return super(BankPaymentLine, self).unlink()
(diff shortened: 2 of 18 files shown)


account_payment_line_cancel/views/payment_line_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="account_payment_line_cancel_view" model="ir.ui.view">
```
tbaden, line 3: cosmetical change:
use original id

> prefer naming the inherited view after the original over using a name which follows the guidelines

account_payment_line_cancel/views/invoice_view.xml
```diff
@@ -0,0 +1,32 @@
...
+          <footer>
+            <button string="Free Invoices" name="invoice_free" type="object" default_focus="1" class="oe_highlight"/>
+            or
+            <button string="Cancel" class="oe_link" special="cancel"/>
+          </footer>
+        </form>
+      </field>
+    </record>
+
+    <act_window id="action_account_invoice_free"
```
tbaden, line 27: cosmetical change: id pattern

> For an action: the main action respects <model_name>_action. Others are suffixed with _<detail>, where detail is an underscore lowercase string explaining the action (should not be long). This is used only if multiple actions are declared for the model.

account_payment_line_cancel/views/invoice_view.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Copyright (C) 2014 Compassion (http://www.compassion.ch)
+  @author: Cyril Sester <cyril.sester@outlook.com>
+  The licence is in the file __openerp__.py
+-->
+<odoo>
+    <record id="account_invoice_free_view" model="ir.ui.view">
```
tbaden, line 8: cosmetical change: id pattern

> For a view: <model_name>_view_<view_type>, where view_type is kanban, form, tree, search, ...

tbaden (approved):
code review: lgtm
the cosmetical changes mentioned would be nice, but not mandatory

ecino (approved):
:+1:

pedrobaeza:
#599 sets other flow for the same problem, and not sure which one is better, as removing the line from the payment order seems very hard...

ecino:
My feeling is that #599 only unbind the payment line to the move line (leaving the payment line orphan) and then cancels the invoice. Here at least there are some clear messages posted on the thread of the invoice and the payment order so that one can understand what happened. It also better handles the case where the payment is already confirmed by unreconciling the move created (the other module only says it is dangerous to use the function).

ecino:
@rousseldenis is not responding. I gave my opinion, could we make a decision?

CasVissers-360ERP:
I agree with @ecino, this module is a missing piece in payment orders.

account_payment_line_cancel/models/bank_payment_line.py
```diff
@@ -0,0 +1,21 @@
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+from odoo.addons.account_payment_order.models.bank_payment_line \
+    import BankPaymentLine
+
+
+class APLCBankPaymentLine(models.Model):
```
CasVissers-360ERP, line 8: Should this not be AccountPaymentLine?

account_payment_line_cancel/models/__init__.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from . import account_payment_line
-from . import  invoice
+from . import bank_payment_line
+from . import invoice
```
CasVissers-360ERP, line 4: Maybe rename to account_invoice?

CasVissers-360ERP (changes requested):
Code review

PR #518, opened 2018-10-25
[10.0][IMP] Add Italian SEPA SDD CBI Support

robyf70:
Hi,

with this PR I've added the Italian CBI SEPA SDD support. It's currently working since several months now without any issue on several Italian banks.

In particular there is one change which need to be discussed, the  because I haven't found a different way to address it. Actually I've added the scheme selection under account.payment.order because this is necessary as per payment order, so I would like to know if you have a different approach for that.

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -163,10 +163,16 @@ class AccountPaymentOrder(models.Model):
     def generate_pain_nsmap(self):
         self.ensure_one()
         pain_flavor = self.payment_mode_id.payment_method_id.pain_version
-        nsmap = {
-            'xsi': 'http://www.w3.org/2001/XMLSchema-instance',
-            None: 'urn:iso:std:iso:20022:tech:xsd:%s' % pain_flavor,
-        }
+        if pain_flavor.startswith('CBIBdySDDReq'):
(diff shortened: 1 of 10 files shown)


pedrobaeza:
This should go in OCA/l10n-italy

robyf70:
I saw that this repo has German related SEPA things no?

pedrobaeza:
Well, that's an historic error IMO, but it also needs less different things.

robyf70:
the main problem is that it isn't easy to have it in OCA/l10n_italy, it's a bit "intrusive" patch because has to touch certain parts to work at best

pedrobaeza:
Then propose here hooks for easing that task.

robyf70:
Ok! The PR I've proposed aim that. I've divided it in 3 patches to monitor and rollback the changes and I did already all checks to verify that all previous things are still working as before.

pedrobaeza:
I see a lot of specific Italian code here. Please open a new one only including hook methods for the things you need.

robyf70:
The specific Italian code are surrounded by if to check if we are in the CBI specs, the else follow the standard code.

robyf70:
>  Please open a new one only including hook methods for the things you need.

Let me check what is possible to do here

PR #520, opened 2018-11-06
[12.0][MIG] account_payment_order: Migration to v12.0

sergio-teruel:
cc @Tecnativa


account_payment_order/report/account_payment_order.py
```diff
@@ -0,0 +1,45 @@
...
+    @api.model
+    def get_bank_account_name(self, partner_bank):
+        """
+
+        :param partner_bank:
+        :return:
+        """
+        if partner_bank:
+            name = ''
+            if partner_bank.bank_name:
```
ValentinVinagre, line 36: I think this is more correct, what you think?:
```suggestion
            """
    :param partner_bank: partner bank record
    :return: string with the name or False 
"""
name = False
if partner_bank:
    if partner_bank.bank_bic:
        name = '%s BIC %s' % (name, partner_bank.bank_bic)
    elif partner_bank.acc_number:
        if partner_bank.bank_bic:
            name = '%s - ' % (name)
        else:
            name = '%s %s' % (name, partner_bank.acc_number)
    elif partner_bank.bank_name:
        name = '%s: ' % partner_bank.bank_id.name
return name
```

account_payment_order/models/account_move.py
```diff
@@ -0,0 +1,12 @@
+# © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
```
ValentinVinagre, line 4: does not lack the import of '_' for translations?

account_payment_order/models/account_payment_mode.py
```diff
@@ -0,0 +1,148 @@
...
+                ]).ids
+            elif self.payment_method_id.payment_type == 'inbound':
+                aj_ids = ajo.search([
+                    ('type', 'in', ('sale_refund', 'sale')),
+                    ('company_id', '=', self.company_id.id)
+                ]).ids
+            self.default_journal_ids = [(6, 0, aj_ids)]
+
+    @api.onchange('generate_move')
+    def generate_move_change(self):
```
ValentinVinagre, line 133: It would not be better with @api.multi and write?

ValentinVinagre (commented):
hello,
I have indicated some comments, if you think they are necessary you could incorporate them.
A greeting.

account_payment_order/models/account_payment_mode.py
pedrobaeza: An onchange mustn't write ever, as it's a on-memory operation. `@api.multi` is by default, but by convention this should be always applied to only one record.

ValentinVinagre: It's good know it. Thanks @pedrobaeza

ValentinVinagre (changes requested):
Hi @pedrobaeza and @sergio-teruel,
I tested with @HaraldPanten the module 'account_banking_sepa_credit_transfer' and the depency 'account_payment_order' we found one issue... this module needs the field "reference_type" from the invoice.
https://github.com/OCA/bank-payment/blob/7c7c41c77ae345aeba3f3f5a46e193c2e28df273/account_payment_order/models/account_move_line.py#L35
In v.12 this field doesn't exist and odoo base uses the field 'invoice_reference_type' alocated in the 'res.company'.

https://github.com/odoo/odoo/blob/f5c3dafb045f4eb5e832c45596dd59396667ea09/addons/account/models/account_invoice.py#L397

On the other hand, this field doesn't have the option 'none' that is the default option to odoo v.11. 

https://github.com/odoo/odoo/blob/f5c3dafb045f4eb5e832c45596dd59396667ea09/addons/account/models/company.py#L66

How can we manage these changes?

Thanks

account_payment_order/models/account_move.py
jeroenev: _ is only used when translating strings from python code/methods
strings defined on Odoo fields in python or defined in XML are automatically translatable
seems like this file does not contain python methods, only field definitions
so _ for translations is not needed

pedrobaeza:
@ValentinVinagre first thing to do is to manage all dependencies resolved and merge, and then come to this one and let tests to do their job, although manual testing or the already given feedback is valuable. About the dependencies, if the person that made the initial PR is unresponsive, we can superseed the PR and make a new one, preserving current commits, but adding on top the needed changes for completing the work.

pedrobaeza:
@ValentinVinagre I have restored the field `reference_type` as I don't find a proper replacement. This may change in the future if the users of that field (essentially, other countries with fixed communication structures) see that they can rely on a reference type at other level (payment mode level or company level as the core).

Issue #515, opened 2018-10-19
Error validating payment XML file


Byloth:
After the commit 7bf8d6aa8dfe7d738b55d6e77564d952afec2874, if you try to generate a payment XML file from `account.payment.order` Odoo will throw this error message:

```
The generated XML file is not valid against the official XML Schema Definition. The generated XML file and
the full error have been written in the server logs. Here is the error, which may give you an idea on the
cause of the problem : Element '{http://www.six-interbank-
clearing.com/de/pain.001.001.03.ch.02.xsd}PstlAdr': This element is not expected. Expected is ( 
{http://www.six-interbank-clearing.com/de/pain.001.001.03.ch.02.xsd}ClrSysMmbId )., line 40
```

---

This error is caused by this [few lines of codes](https://github.com/OCA/bank-payment/blob/7bf8d6aa8dfe7d738b55d6e77564d952afec2874/account_banking_pain_base/models/account_payment_order.py#L318,L326) of the function named `generate_fininst_postal_address`.

sbidoul:
@Byloth thanks for the report.

The problem comes from https://github.com/OCA/l10n-switzerland/blob/379343c4ad40dcc31455513383b1d60077c5d6f3/l10n_ch_pain_credit_transfer/data/pain.001.001.03.ch.02.xsd#L577-L580 where the Swiss flavour forbids the PstlAdr element.

That is a strange practice to forbid an element that is allowed in the parent standard. But who am I to judge how the Swiss flavor is designed.

I think this should be addressed in l10n-switzerland with an override of the method that generates the postal address, in case the flavor is pain.001.001.03.ch.02.

Unfortunately I'll be offline for a short week, so I cannot work on this now.

sbidoul:
https://github.com/OCA/bank-payment/commit/7bf8d6aa8dfe7d738b55d6e77564d952afec2874 is apparently causing issues with other pain flavors. I'll investigate deeper ASAP.

sbidoul:
Closing as this is being handled in #517 and https://github.com/OCA/l10n-switzerland/pull/453

Byloth:
Great @sbidoul,  
so is it functionally correct to handle this problem as I did in [`l10n-switzerland`'s PR #453](https://github.com/OCA/l10n-switzerland/pull/453)?

Has it been resolved in #517?  
Should I close the [PR](https://github.com/OCA/l10n-switzerland/pull/453) or should I keep it open?  
Could it be a problem with those changes made in #517?

[...]

PR #517, opened 2018-10-24
[11.0] expand pain support

sbidoul:
Forward port of #502

PR #522, opened 2018-11-07
[11.0] only generate FinInstId PstlAdr for some pain flavors

sbidoul:


sbidoul:
This is a forward port and fix of https://github.com/OCA/bank-payment/commit/7bf8d6aa8dfe7d738b55d6e77564d952afec2874 (10.0), which was erroneously reverted in #517 (sorry for the noise). 

Also providing a hook to fix issue https://github.com/OCA/bank-payment/issues/515 raised by l10n-switzerland.

pedrobaeza (approved):

j-zaballa (approved):

ecino:
This is becoming an old issue and we are having troubles with pain files in Switzerland since end of last year. I propose that we merge this with high priority.

ecino (approved):

Byloth (approved):

ecino:
Wait.. isn't it conflicting with what has been changed in v10 ? See 6bca6c3010d714573f67635d8a22dbd4fe94db77
In v10 a new field has been added to be able to manually activate or not the address, should we use the same mechanism in v11 instead of adding this method that must be then updated for each plain flavor?

PR #527, opened 2018-11-28
[10.0] Make bank address optional in XML generation

alexis-via:
Some banks, at least in France, refuse files with bank addresses => new field 'pain_bank_address' on account.payment.method.

Improve bank address block
Other small code improvements

By the way, after some more reading of the EPC guidelines, I'm not fully convinced about the way the generation of the address block was modified by @sbidoul , see my comment on this commit https://github.com/OCA/bank-payment/commit/89f4735d03fb5b98f2c9637694331d6efd209a29


sbidoul (approved):
That looks much better indeed.

pedrobaeza:
Should this be forward-port it to v11?

alexis-via:
Yes, we should forward-port this.

FYI, I had the confirmation of one of my customers that, with this new option on payment methods unchecked, the SEPA credit transfer XML files generated by Odoo are working again with HSBC France.

PR #528, opened 2018-12-04
[IMP] account_payment_partner: Add options to show partner bank accou…

cubells:
…nt in invoice report

Based on https://github.com/OCA/bank-payment/pull/458

cc @Tecnativa


pedrobaeza (approved):
I have preserved commit history, plus adapting tests for this version. Fast-tracking this when CIs finish as this has been approved on previous version and tests cover the new features (plus other old ones).

account_payment_partner/models/account_invoice.py
```diff
@@ -42,13 +42,10 @@ def _onchange_partner_id(self):
...
-                    force_company=self.company_id.id
-                ).partner_id.customer_payment_mode_id
-                self.payment_mode_id = pay_mode
-                if pay_mode and pay_mode.bank_account_link == 'fixed':
-                    self.partner_bank_id = pay_mode.fixed_journal_id. \
-                        bank_account_id
+                # No bank account assignation is done here as this is only
+                # needed for printing purposes and it can conflict with
+                # SEPA direct debit payments. Current report prints it.
+                self.payment_mode_id = self.partner_id.customer_payment_mode_id
```
i-vyshnevska: @pedrobaeza
partner_bank_id used not only for printing purposes it makes impact on l10n_ch_base_bank, can I return it initialization here?

pedrobaeza: I don't get your point, but setting in invoice the bank account is a problem in a lot of levels. If you need it for specific modules, set it in that module, not in the general one.

i-vyshnevska: ok, thanks

PR #509, opened 2018-10-02
12.0 mig account payment mode

mourad-ehm:

PR #529, opened 2018-12-06
[MIG 12.0] account payment mode (with account payment method)

levkar:
This PR migrates account_payment_mode to 12.0 and strips account_payment_method out of it for better modularity.

This module is stripped because any module that extends (and eventually depending on) account.payment.method needs to depend on account_payment_mode. For example for just having payment means on payment method one needs to install account_payment_mode. See account_payment_unece here: https://github.com/OCA/community-data-files/pull/47

@astirpe @alexis-via


alexis-via:
The object account.payment.method is defined in the "account" module, not in the module account_payment_mode. The module account_payment_mode only adds a few small details on that object. This module account_payment_mode was designed to be very small so that many upper module could depend on it without problems. That's why we have a separate module account_payment_partner for example. It really don't feel the need to split account_payment_mode in 2 different modules...

levkar:
Then there is no way of using account_payment_unece without installing account_payment_mode.

If that is the case, better we revert back to original and make account_payment_unece depend back on account_payment_mode.

alexis-via:
@levkar What is the problem exactly of depending on account_payment_mode ? This module is so small (and it was on purpose) that I don't really see a problem to depend on it. OK, maybe it will add the object account.payment.mode that you may not use, but it is really a problem ? Modularity is good, but having a high number of super-small modules would bring more problems than solutions I think.

levkar:
I would prefer small modules over unnecessary dependencies. I can fork and change it in my own repo. 

There is yet another PR #509 without the split. He can use this PR as reference since I fixed all the comments in PR #509.

Closing this PR. Thanks for the feedback.

PR #531, opened 2018-12-17
[12.0] [REF] Account Payment Order: admin access rights

ThomasBinsfeld:
- [x] Add Admin in the security group

### DIFF
--- a/account_payment_order/security/payment_security.xml
+++ b/account_payment_order/security/payment_security.xml
@@ -5,7 +5,7 @@
 
 <record id="group_account_payment" model="res.groups">
     <field name="name">Accounting / Payments</field>
-    <field name="users" eval="[(4, ref('base.user_root'))]"/>
+    <field name="users" eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"/>
     <field name="category_id" ref="base.module_category_extra"/>
 </record>
 


pedrobaeza (approved):

benwillig (approved):

ThomasBinsfeld:
@pedrobaeza Do not merge. It doesn't work as expected. I'll try to fix it tomorrow.

pedrobaeza:
OK, I approved it over code without trying as enterprise is not available here.

setup/account_payment_order_enterprise/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
+
+This directory caches those eggs to prevent repeated downloads.
+
+However, it is safe to delete this directory.
+
```
sbidoul: @ThomasBinsfeld remove this file

account_payment_order/security/payment_security.xml
```diff
@@ -5,7 +5,7 @@
 
 <record id="group_account_payment" model="res.groups">
     <field name="name">Accounting / Payments</field>
-    <field name="users" eval="[(4, ref('base.user_root'))]"/>
+    <field name="users" eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"/>
```
sbidoul, line 8: Why do we need both?

account_payment_order_enterprise/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+        This addon ensures the compatibility of the account_payment_order
+        addon with the Odoo 12 Enterprise Accounting addon.""",
+    'version': '12.0.1.0.0',
+    'license': 'LGPL-3',
+    'author': 'Odoo Community Association (OCA), ACSONE SA/NV',
+    'website': 'https://github.com/OCA/bank-payment',
+    'category': 'Banking addons',
+    'depends': [
+        # Odoo Enterprise
+        'account_accountant',
```
sbidoul: I don't think it's ok to have a dependency on an enterprise module in OCA, if only because tests will fail.

pedrobaeza: This one wasn't there when I approved it! Nasty boy :stuck_out_tongue_closed_eyes:

account_payment_order/security/payment_security.xml
ThomasBinsfeld: @sbidoul `base.user_root` access does not grant access to Admin since they are two different users in Odoo 12. I observed that it has been not as it pretty much everywhere in Odoo 12. https://github.com/odoo/odoo/commit/2f7c03d9ca5809b65dc6eaa778064668d34fcd2e

ThomasBinsfeld:
@pedrobaeza @sbidoul I removed the Enterprise compatibility part.

PR #532, opened 2018-12-19
[11.0][account_banking_mandate] fixes to support multicompany

JordiBForgeFlow:
The following fixes are required when you operate in a multicompany environment.

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -157,11 +157,27 @@ class AccountBankingMandate(models.Model):
                     'account.banking.mandate') or 'New'
         return super(AccountBankingMandate, self).create(vals)
 
-    @api.multi
+    @api.onchange('company_id')
+    def _onchange_company_id(self):
+        res = {}
+        partner_domain = self._get_default_partner_bank_id_domain()
+        if self.company_id:
+            if self.partner_bank_id and \
+                    self.company_id and \
+                    self.partner_bank_id.company_id and \
+                    self.partner_bank_id.company_id != self.company_id:
+                self.partner_bank_id = False
+            partner_domain.append(('company_id', '=', self.company_id.id))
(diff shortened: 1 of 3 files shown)


account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -157,11 +157,25 @@ def create(self, vals=None):
...
         return super(AccountBankingMandate, self).create(vals)
 
-    @api.multi
+    @api.onchange('company_id')
+    def _onchange_company_id(self):
+        res = {}
+        partner_domain = self._get_default_partner_bank_id_domain()
+        if self.company_id:
+            if self.partner_bank_id and \
+                    self.partner_bank_id.company_id != self.company_id:
```
etobella, line 168: and self.company_id and self.partner_bank_id.company_id

etobella (changes requested):

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -157,11 +157,27 @@ def create(self, vals=None):
...
+            domain = {
+                'partner_bank_id': partner_domain}
+        res['domain'] = domain
+        return res
+
     @api.onchange('partner_bank_id')
     def mandate_partner_bank_change(self):
-        for mandate in self:
-            mandate.partner_id = mandate.partner_bank_id.partner_id
+        self.partner_id = self.partner_bank_id.partner_id
```
MiquelRForgeFlow, line 180: not sure, but shouldn't you add `if self.partner_bank_id.partner_id:` ?

JordiBForgeFlow: Well, if there's no partner then self.partner_id = False. No need for the if

AaronHForgeFlow (approved):
:+1: Functional test on runbot

MiquelRForgeFlow (approved):
LGTM

PR #533, opened 2018-12-21
[12.0] [MIG]  Account Banking Mandate Sale

ThomasBinsfeld:


account_banking_mandate_sale/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    # This field commercial_partner_id should be moved
+    # in an OCA base module named for example sale_commercial_partner
+    commercial_partner_id = fields.Many2one(
+        related='partner_id.commercial_partner_id', string='Commercial Entity',
+        store=True, readonly=True)
```
tbaden, line 14: improve readability 

```
    commercial_partner_id = fields.Many2one(
        related='partner_id.commercial_partner_id',
        string='Commercial Entity',
        store=True,
        readonly=True)
```

account_banking_mandate_sale/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+    mandate_id = fields.Many2one(
+        'account.banking.mandate', string='Direct Debit Mandate',
+        ondelete='restrict', readonly=True,
+        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]})
```
tbaden, line 18: improve readability
```
mandate_id = fields.Many2one(
        'account.banking.mandate', 
        string='Direct Debit Mandate',
        ondelete='restrict', 
        readonly=True,
        states={'draft': [('readonly', False)], 'sent': [('readonly', False)]})
```

tbaden (changes requested):
thanks for the work 👍 
I did a code review. just some small changes.

max3903 (approved):

tbaden (approved):

pedrobaeza:
@max3903 if the rebased branch has failed, it's because there's something wrong. You mustn't merge manually as you are breaking the branch. cc @sbidoul

sbidoul:
@pedrobaeza @max3903 does the bot message need to be clarified?

pedrobaeza:
For me it's clear enough, but you can add that warning: "Merging this PR in the main branch will break it" or similar.

max3903:
@sbidoul Yes it needs clarifications. In this case, the failing tests are codecov ones. Code coverage is not a requirement for merging a beta (or not-specified development status) module. ocabot should take maturity level into consideration. See https://github.com/OCA/maintainer-quality-tools/pull/602

pedrobaeza:
@max3903 that's not the problem: the problem is that this branch, when rebased over current main branch + current Odoo, it's failing in Travis, so the problem is not only codecov.

sbidoul:
@max3903 clarification here: https://github.com/OCA/oca-github-bot/pull/56

max3903:
@sbidoul Thank you.

@pedrobaeza The branch is failing due to another module `account_payment_partner`:
https://travis-ci.org/OCA/bank-payment/jobs/554740660#L1841
or am I missing something?

pedrobaeza:
We don't really know. Maybe this module is interacting with the other and make it failed. But now we don't know as this is merged. Previously to this, we would be able to relaunch the main branch and see if it's failing too, and thus the problem would be that something upstream has broken the branch, and thus we can tell the module maintainer to check.

In this current situation, please investigate and solve the problem, or at least, point to the problem to the proper person.

max3903:
@pedrobaeza Will do.

PR #534, opened 2018-12-21
[IMP] set a filename according to the name of the mandate

vrenaville:
Actually, the file store in the mandate is not named correctly, so the user cannot open it if he wants to download it from the odoo interface.


pedrobaeza:
Well, I think you are talking about Windows need of having the extension in the name...

Any way, why do you need an extra field for this? Check also the CI.

PR #535, opened 2018-12-29
[FIX] only drop column if it exists

hhgabelgaard:
In my v10 base acc_type is a computed non-stored field
Basically same change as here: 11681f5114bc9cf6cf82100bfb9262bf5fddc4a6


pedrobaeza (approved):
Fast-tracking this.

PR #538, opened 2019-01-21
[12.0] [REF] Payment Order: better performance of draft2open method

ThomasBinsfeld:
Add a norecompute in draft2open to avoid the recomputation of total_company_currency n times (where n is the number of payment lines)

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -280,7 +280,12 @@ class AccountPaymentOrder(models.Model):
                             payline.ml_maturity_date,
                             requested_date))
                 # Write requested_date on 'date' field of payment line
-                payline.date = requested_date
+                # norecompute is for avoiding a chained recomputation
+                # payment_line_ids.date
+                # > payment_line_ids.amount_company_currency
+                # > total_company_currency
+                with self.env.norecompute():
+                    payline.date = requested_date
                 # Group options
                 if order.payment_mode_id.group_lines:
                     hashcode = payline.payment_line_hashcode()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you explain why `total_company_currency` is recomputed changing the date?

ThomasBinsfeld:
@pedrobaeza The `_compute_total` method depends on `payment_line_ids.amount_company_currency` (https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/models/account_payment_order.py#L145). 

And the `_compute_amount_company_currency` which computes the `payment_line_ids.amount_company_currency` field depends on `date` (https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/models/account_payment_line.py#L77).

The field `amount_company_currency` on `account.payment.line` is not stored but Odoo 12.0 is able to find the transitive dependencies over non-stored fields : https://github.com/odoo/odoo/commit/3fbd86bcbe6dffe923e10d4690ee302c6a779f90

So the `total_company_currency` of the payment order is recomputed every time the `date` changes on some of its payment lines.

pedrobaeza:
Thanks for the detailed explanation, @ThomasBinsfeld! Now all makes sense. Putting a summary as comment before this will help in the future for knowing why is there. Something like:
```python
# this is for avoiding a chained recomputation
# payment_line_ids.date > payment_line_ids.amount_company_currency > total_company_currency
```

ThomasBinsfeld:
@pedrobaeza Done. Thanks for your review.

pedrobaeza (approved):

sbidoul (approved):

PR #539, opened 2019-01-21
[12.0] [IMP] Payment Order: due on field in payment line wizard

ThomasBinsfeld:
New field `due_on` on the payment line wizard allowing to select the operator (`<=` or `=`) to apply on the due date or move date. Default operator remains `<=`.

### DIFF
--- a/account_payment_order/wizard/account_payment_line_create.py
+++ b/account_payment_order/wizard/account_payment_line_create.py
@@ -30,6 +30,14 @@ class AccountPaymentLineCreate(models.TransientModel):
         ('due', 'Due Date'),
         ('move', 'Move Date'),
         ], string="Type of Date Filter", required=True)
+    due_on = fields.Selection(
+        string="Due on",
+        selection=[
+            ('<=', 'Before or equal'),
+            ('=', 'Equal')],
+        default='<=',
+        required=True,
+    )
     due_date = fields.Date(string="Due Date")
     move_date = fields.Date(
         string='Move Date', default=fields.Date.context_today)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
For me this promotes bad habits, as you have a lot or risk of forgetting some dues filtering with this. The reason for not having it is because of this. @alexis-via do you agree with me?

account_payment_order/wizard/account_payment_line_create_view.xml
```diff
@@ -15,6 +15,7 @@
             <group name="main">
                 <field name="order_id" invisible="1"/>
                 <field name="date_type"/>
+                <field name="due_on"/>
                 <field name="move_date" attrs="{'required': [('date_type', '=', 'move')], 'invisible': [('date_type', '!=', 'move')]}"/>
                 <field name="due_date" attrs="{'required': [('date_type', '=', 'due')], 'invisible': [('date_type', '!=', 'due')]}"/>
```
sbidoul, line 20: date_type, due_on and move_date/due_date should be on the same line

So it reads like
* Due Date is before or equal to <some date>
* Move Date is equal to <some date>

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -30,6 +30,14 @@ class AccountPaymentLineCreate(models.TransientModel):
         ('due', 'Due Date'),
         ('move', 'Move Date'),
         ], string="Type of Date Filter", required=True)
+    due_on = fields.Selection(
+        string="Due on",
+        selection=[
+            ('<=', 'Before or equal'),
+            ('=', 'Equal')],
```
sbidoul, line 37: s/Equal/is equal to/
s/Before or equal/is before or equal to/

sbidoul:
@pedrobaeza the default will not change. We have real use cases where the customer wants to do a different payment file for each due date. It sounds legit to me.

@ThomasBinsfeld I made a couple of UX improvement suggestion.

Actually the resulting UX will be better: currently I always ask myself which operator will be applied when selecting the date. With the PR it will be explicit (is before or equal to <the date I select>).

sbidoul (changes requested):
UX improvement

pedrobaeza:
For me it's still a :-1:

PR #541, opened 2019-01-22
[10.0][ADD] new module account_payment_clearing_system

BT-tbaechle:
This module allows to set a 'Clearing System Member Identification' type on a bank (res.bank) and
provide a Clearing System Member Identification Number which is needed for banks that don't have a
Bank Identifier Code (BIC) to generate a PAIN (payment instruction) file that can be used for international
payments.

The module was built using the specifications listed here: https://www.iso20022.org/external_code_list.page


pedrobaeza:
Please don't use plurals in module name.

BT-tbaechle:
@pedrobaeza are there any other changes required or does this look good now? Thanks!

pedrobaeza:
Well, the formal requirements are good (Travis, Runbot and even coverage), but now a review of the feature and the code should be done by reviewers interested in this one. I'm not one of them, as I only give a quick kick-in in most of the PRs.

PR #542, opened 2019-01-22
[11.0][FIX] account_payment_mode res.partner view banks group attribute

j-zaballa:
Description of the issue/feature this PR addresses:

The modules adds `{'invisible': [('parent_id', '!=', False)]}` attribute
to `banks` group on res.partner form view. The rationale for this,
according to a comment on the view itself is:
> ... there is a domain on the 'partner_id' field of res.partner.bank
(base module) which prevents the selection of a contact

However, the domain the comment refers to is
`['|', ('is_company', '=', True), ('parent_id', '=', False)]`
So, I think the domain for the group to be invisible should be
`[('parent_id', '!=', False), ('is_company', '=', False)]`. In addition,
a parent that is a company is always its own commercial partner. So we
should have access to its payment information.

Current behavior before PR:

Bank and mandates information is hidden when a partner is a company and
has a parent.

Desired behavior after PR is merged:

Banking information should be visible if a partner is a company.

### DIFF
--- a/account_payment_mode/views/res_partner.xml
+++ b/account_payment_mode/views/res_partner.xml
@@ -12,7 +12,7 @@ which prevents the selection of a contact -->
     <field name="inherit_id" ref="account.partner_view_buttons"/>
     <field name="arch" type="xml">
         <group name="banks" position="attributes">
-            <attribute name="attrs">{'invisible': [('parent_id', '!=', False)]}</attribute>
+            <attribute name="attrs">{'invisible': [('parent_id', '!=', False), ('is_company', '=', False)]}</attribute>
         </group>
     </field>
 </record>


pedrobaeza:
Why removing the possibility on partners without that check? They can be "autonomos" (I don't know the right word in English.

j-zaballa:
> Why removing the possibility on partners without that check? They can be "autonomos" (I don't know the right word in English.

We can see self-employed peoples's bank information, as long as the contact does not have a parent. It is the same condition that base module has for `res.partner.bank` `partner_id` field.

https://github.com/odoo/odoo/blob/d79007cb0943a0c53361e6f39547764fc78e3d5a/odoo/addons/base/res/res_bank.py#L65

pedrobaeza:
OK, I see.

pedrobaeza (approved):

jonerikceberio (approved):

sbidoul (approved):

pedrobaeza:
Cherry-picked in v12 in 3ec955ea74454bad7bfd475550f091ea1a6cb71d

PR #547, opened 2019-01-27
[11.0] Set account number partner_id (supplier)

altoch:
Set a context to the bank account created from the purchase order, if not set then is assigned to the main company as the account owner.


pedrobaeza:
We have here the same problem mentioned in the other place if we don't set a commercial partner in the purchase order, but the field is not available as this level. I suggest to put `create="0"` instead for avoiding the creation from here.

altoch:
Is it posible to merge?

altoch (commented):
Fix closing tag

PR #550, opened 2019-02-06
[12.0] [FIX] Account Banking SEPA Direct Debit: creditor identifier in config

ThomasBinsfeld:
The sepa_creditor_identifier in res.config.settings is readonly by default (related) but shouldn't be.


pedrobaeza (approved):
Right, this can be fast-tracked.

PR #551, opened 2019-02-12
Add the button to mark as done a payment order.

BT-cserra:


pedrobaeza (approved):

sbidoul:
Hi,

I'm not sure about this one. In v8 the transition to done was made automatically in upstream Odoo upon reconciliation (via transfer account).

In v9+ this feature is gone. But in practice it was not really useful and the only good way to track progress of a payment order is through reconciliation or bank statement matching (or unreconciliation through payment return).

So my question is: why/when would a user click on "done" on payment orders. In my mind, confirming the order has been correctly upload/sent to the bank marks the end of the process for the payment order itself, and the rest takes place through regular account reconciliation.

Actually I would rather remove the 'done' state.

BT-fschubert:
Hi,
we added this button due to customers demand and the relation was lost to get this state automatically - from my point of view it is an intermediate step until a useful relation is either found on odoo core or added to a useful trigger in a bank-payment module. (We didn't had the time to search for a final approach).
I disagree to remove the "done" state due to usability reasons, cause related to our manual intermediate solution the user has no chance to see a difference between "Uploaded" and "Done" and he is conditioned to have an end of a process with state "Done".

tbaden (approved):
code review: lgtm

pro done state for a clean end in process

pedrobaeza:
@sbidoul your final call

KingLeo40:
Hello

KingLeo40:
docs/content/github/setting-up-and-managing-your-github-user-account/permission-levels-for-a-user-account-repository.md

KingLeo40:
I got the minner

KingLeo40:
account_banking_mandate/__manifest__.py

PR #526, opened 2018-11-28
[10.0] Improve handling of refunds

alexis-via:
Ability to sort payment lines by maturity date

PR #552, opened 2019-02-19
[12.0] [REF] Payment Order: allow add to order from invoice if order exists

ThomasBinsfeld:
This commit allows to add an invoice to an order even if it's in orders that are not in progress.
In progress orders are orders in draft, open or generated states.

### DIFF
--- a/account_payment_order/models/account_invoice.py
+++ b/account_payment_order/models/account_invoice.py
@@ -79,7 +79,8 @@ class AccountInvoice(models.Model):
                 lambda x: (
                     not x.reconciled and x.payment_mode_id.payment_order_ok and
                     x.account_id.internal_type in ('receivable', 'payable') and
-                    not x.payment_line_ids
+                    not any(p_state in ('draft', 'open', 'generated')
+                            for p_state in x.payment_line_ids.mapped('state'))
                 )
             )
             if not applicable_lines:


pedrobaeza:
This would add the same move line in several payments. For me that's incorrect.

sbidoul:
@pedrobaeza this is necessary so one can resubmit a payment that has been rejected. Very useful in combination with account_payment_return.

sbidoul:
@pedrobaeza this is necessary so one can resubmit a direct debit order that has been rejected.

sbidoul (approved):
This is consistent with the payment order wizard at https://github.com/OCA/bank-payment/blob/f54de999456049f5b53a2e44bb622fb42fca5c28/account_payment_order/wizard/account_payment_line_create.py#L117

pedrobaeza:
And which is the allowed state? Isn't very specific of a flow with transfer account? What happens if not having transfer account? I think we should add code in the module `account_payment_order_return`

sbidoul:
@pedrobaeza it's not specific to transfer account. It works like this:
1. you add an invoice to a payment order
2. it' can't be added to another payment order because the payment is not sent
3. when the payment order is sent, the invoice is reconciled therefore it can't be added to a payment order
4. if for some reason the payment is rejected (with or without using payment return), the invoice is unreconciled
5. you can add it again to a payment order, back to step one.

This PR simply makes sure the "add to payment order" button on the invoice behaves the same as the "add moves to payment order" wizard on payment orders. Nothing exotic, really :)

sbidoul:
> which is the allowed state

`sent`

pedrobaeza:
My colleague @carlosdauden is analyzing this for replying you, as we have been using `account_payment_return` modules since a lot.

sbidoul:
Really this is not related to the `account_payment_return` module. This is just to have the same rule on the two ways to add an invoice to a payment order.

pedrobaeza (approved):
OK, take my tentative approval for now. We want to go further to re-evaluate both filters any way.

sbidoul:
> We want to go further to re-evaluate both filters any way

Ok, ping me when you want to discuss it. That filter was important for partial payments too.

carlosdauden (approved):

pedrobaeza:
@sbidoul indeed in previous versions we use the wizard, so this problem didn't arise. We don't see more problems with it. Other thing is the refunds, that we have still pending, but that's another topic.

pedrobaeza:
Cherry-picked for v11 in 3dbe3686cd06781c98517394b4dc42764cd720ee for consistency.

sbidoul:
About the refunds, there is a big rework by @alexis-via in https://github.com/OCA/bank-payment/pull/526: we are planning to test it in the coming weeks.

pedrobaeza:
Oh, that's great. I don't know why I miss that one. The initial comment is not very helpful. but the title seems clear...

PR #554, opened 2019-02-22
[FIX] account_payment_order: select payable and receivable lines

sebalix:
When selecting move lines to create transactions, take into account
payable and receivable lines (to include refunds) as it is done in v11.

### DIFF
--- a/account_payment_order/wizard/account_payment_line_create.py
+++ b/account_payment_order/wizard/account_payment_line_create.py
@@ -102,7 +102,7 @@ class AccountPaymentLineCreate(models.TransientModel):
             domain += [
                 ('credit', '>', 0),
                 #  '|',
-                ('account_id.internal_type', '=', 'payable'),
+                ('account_id.internal_type', 'in', ['payable', 'receivable']),
                 #  '&',
                 #  ('account_id.internal_type', '=', 'receivable'),
                 #  ('reconcile_partial_id', '=', False),  # TODO uncomment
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Refunds are not supported by default, and with a good reason. #526 is supposed to add full support.

sebalix:
@pedrobaeza I'm not an expert, you may be right. But it is done like this in v11: https://github.com/OCA/bank-payment/blob/11.0/account_payment_order/wizard/account_payment_line_create.py#L108

pedrobaeza:
Well, I can't say about that then. @alexis-via you have more context now that you worked on #526. Can you judge?

sbidoul:
If it's how it's done in v11, then I don't know how it went through, but it looks too simplistic to me. And the comment in that v11 code is totally out of sync with the implementation (that's the problem with comments...).

The logic is explained in the comment at [line 92](https://github.com/OCA/bank-payment/pull/554/files#diff-47375d82ff1578161a39362d5d156ee0R92). That logic is not in place anymore because the commented out code at lines 106-108 (that part was disabled during migration to 9/10 due to the change in reconciliation model.

pedrobaeza:
@sbidoul then I don't get your point. Are you saying that current v11 code is not correct? And this PR neither?

sbidoul:
@pedrobaeza hm, I tested again. This looks correct, When a partial refund is reconciled with the invoice, the refund as reconciled=True, but the invoice stays with reconciled=False, so this works correctly.

So +1 for this PR for aligning with 11/12. Sorry for the noise.

sbidoul (approved):

sebalix:
@sbidoul thank you for your checks!

i-vyshnevska (approved):

PR #555, opened 2019-02-25
[12.0] [REF] direct debit: refactor some checks

ThomasBinsfeld:
Refactoring of the checks on the mandate validity allowing easier inheritance.
The check is done before (in `draft2open`) to avoid having orders confirmed but unable to generate a consistent file anyway.

### DIFF
--- a/account_banking_sepa_direct_debit/models/account_payment_line.py
+++ b/account_banking_sepa_direct_debit/models/account_payment_line.py
@@ -1,0 +1,48 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+
+class AccountPaymentLine(models.Model):
+    _inherit = 'account.payment.line'
+
+    @api.multi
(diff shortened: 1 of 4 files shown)


sbidoul (approved):
This enables earlier detection of errors.

account_banking_sepa_direct_debit/models/account_payment_line.py
```diff
@@ -0,0 +1,48 @@
...
+        This method checks whether the payment line(s) are ready to be used
+        in the SEPA Direct Debit file generation.
+        :raise: UserError if a line does not fulfils all requirements
+        """
+        for rec in self:
+            if not rec.mandate_id:
+                raise UserError(_(
+                    "Missing SEPA Direct Debit mandate on the line with "
+                    "partner {partner_name} (reference {reference}).").format(
+                    partner_name=rec.partner_id.name, reference=rec.name))
```
sbidoul, line 31: BTW, interestingly, this check on mandate_id is already present in account_banking_mandate.

pedrobaeza: I think it's for having a custom message.

pedrobaeza (approved):

PR #562, opened 2019-03-08
11.0 [FIX] tests for account_banking_sepa_direct_debit tests

etobella:
This PR fixes the crash of the tests.


etobella:
@tecnativa @eficent

pedrobaeza:
If odoo/odoo#31699 is merged, we won't need this.

PR #566, opened 2019-03-15
[12.0] [REF] Banking Mandate: searches with sudo in constraint

ThomasBinsfeld:
The searches in the constraint should be done as sudo since it should not depend on the access of the user. Creation/modification of a mandate should not depend on the access level of the user on invoices, move lines and payment lines.


sbidoul (approved):
Absolutely.

pedrobaeza (approved):

sbidoul:
I think I mentioned one day to someone at Odoo that constraints should always be evaluated as sudo by the framework.

PR #567, opened 2019-03-15
[12.0] [ADD] Payment Order: only payment order option on payment methods

ThomasBinsfeld:
New option on payment methods: payment order only (unchecked by default) allowing to enforce the use of payment orders for some payment methods

### DIFF
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -22,6 +22,7 @@
         'base_iban',  # for manual_bank_tranfer
     ],
     'data': [
+        'views/account_payment_method.xml',
         'security/payment_security.xml',
         'security/ir.model.access.csv',
         'wizard/account_payment_line_create_view.xml',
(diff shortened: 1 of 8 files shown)


sbidoul:
@ThomasBinsfeld code looks good.

Do you think it's possible to hide payment methods that have this flag set from the "register payment" wizard?

image

ThomasBinsfeld:
@sbidoul It should be possible. I'll look into it.

chrisdec:
Functional test 👍

chrisdec (approved):
functional test 👍

pedrobaeza (approved):

pedrobaeza:
Some tests for not dropping the coverage would be great

ThomasBinsfeld:
@pedrobaeza @sbidoul Done

pedrobaeza:
One nitpicking for next times: this is not `[ADD]`, but `[IMP]`, as you are not adding a module, but improving an existing one

sbidoul:
@alexis-via this is the PR that added `{in,out}bound_payment_order_only`. It looks like the goal was to avoid the manual creation of account.payment objects for some payment methods and force the use of payment order for those.

PR #569, opened 2019-03-19
[FIX] Refactor address nodes generation to allow override

ecino:
There are still issues with pain files generation in Switzerland. One problem is because of the new added nodes TwnNm and Pstcd that are not supported together with AdrLine nodes. We propose a small refactor that allows removing those nodes in `l10n_switzerland`.

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -424,6 +424,16 @@ class AccountPaymentOrder(models.Model):
         return True
 
     @api.model
+    def _has_detailed_address(self, gen_args):
+        """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
How is this related with #570?

ecino:
@pedrobaeza No one cared about this issue in Switzerland for months, and suddenly we have two people doing a fix at the same time. I didn't have time to review #570 but it seems that the changes of @p-tombez allow more information to be sent to the bank. I'll close mine when I'm sure the other works as well.

ecino:
Closing in favor of #570 which is a better fix.

pedrobaeza:
Closing as #570 is merged.

PR #570, opened 2019-03-20
[10.0] account_banking_pain_base: Allow override of address block

p-tombez:
Move the address block generation in separate method to allow override.
This will be used in OCA/l10n-switzerland


yvaucher (approved):
This makes it more hookable with different country rules with Structured / Unstructured format where some more rules applies like in switzerland where it is forbidden to mix both.

pedrobaeza:
Please switch the order of the methods for smaller diff for easing review.

pedrobaeza:
How is this related with #569?

ecino (approved):

ecino:
Tested, good to me!

PR #573, opened 2019-03-27
[IMP] 10.0 bank_payment_order move invoice_ref calculation to python

mpanarin:
Moves the logic of invoice_ref calculation to python method, as well as fixes a bug when it sometimes returns False.
Lesser logic in templates - better

### DIFF
--- a/account_payment_order/report/account_payment_order.py
+++ b/account_payment_order/report/account_payment_order.py
@@ -21,6 +21,7 @@ class AccountPaymentOrderReport(models.AbstractModel):
             'data': data,
             'env': self.env,
             'get_bank_account_name': self.get_bank_account_name,
+            'get_invoice_ref': self.get_invoice_ref,
             'formatLang': formatLang,
         }
         return self.env['report'].render(
@@ -46,3 +47,9 @@ class AccountPaymentOrderReport(models.AbstractModel):
             return name
         else:
             return False
+
+    @api.model
+    def get_invoice_ref(self, line):
+        value = (line.move_line_id.invoice_id.number or
+                 line.move_line_id.move_id.name)
+        return value or ''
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I prefer not, as this avoids the customization in the template itself.

mpanarin:
Well, that is exactly the premise.

Adding logic to the template is a generally bad practice. Impossible to test as well as not really obvious.
If you want to customize this behaviour -> extend python method all you want :\
or replace the object in the template

pedrobaeza:
OK, but I prefer to have a computed (non stored) field then.

simahawk:
> OK, but I prefer to have a computed (non stored) field then.

IMO the only benefit would be to cache it, which is also the downside: you cache something useless as it is only for the report. 

If you want to override it, you can still easily override by overriding the report model.

pedrobaeza:
That's exactly why I want it as a field, for caching purposes. It's not useless, as it keeps for the same move line data if several uses. This time maybe it's not the best example for this, as you barely repeat this print, but the other reason is to keep code in same place and for reuse: if you want that same reference in other place, you don't have to rewrite the same code.

sbidoul:
Since the field is for the report only, this approach looks good to me. 

A field would be hard to name: invoice_ref_for_payment_order_report?

sbidoul (approved):

simahawk:
Exactly. The real point IMO is that if every one starts to add fields for reporting only then you'll have tons or useless (or not so used) fields on every model. If you want to do it properly you should put it in a base module which is reused by every other module needing the field, but once you get here you are just adding more complexity for a little gain.

account_payment_order/report/account_payment_order.py
```diff
@@ -21,6 +21,7 @@ def render_html(self, docids, data=None):
             'data': data,
             'env': self.env,
             'get_bank_account_name': self.get_bank_account_name,
```
simahawk, line 23: the only thing I'd change for the future (in backward compat mode): pass `'report': self'` to not have to override this method every time you need another one. Then in the template you can use `report.get_me_something`

simahawk (approved):

Tonow-c2c (approved):
Just Code review

PR #576, opened 2019-04-02
[12.0] [IMP] Invoice: allow modification of mandate on open invoices

ThomasBinsfeld:
In case of a wrong mandate, it should be possible to re-generate the debit order with another mandate.


sbejaoui (approved):
LGTM

PR #577, opened 2019-04-08
[ADD] Added Support for payments in Austria

wpichler:


pedrobaeza:
For me this should be in Austrian localization, not here. In the past, a specific German format was introduced here and it's a pain for non German contributors/migrators to know if the compatibility is kept when migrating or doing any change.

wpichler:
Sounds good for me - i will close this an open a new in austria localization

wpichler:
There is no l10n-austria - can you create one ?

pedrobaeza:
Doing soon!

wpichler:
ok - i will close this for now

pedrobaeza:
Everything is set in https://github.com/OCA/l10n-austria!

wpichler:
Thanks

PR #585, opened 2019-04-22
[IMP] account_banking_mandate: sql constraints

nseinlet:
SQL constraints are faster than python equivalent, use them when 
possible. Added index on the O2M fields used in python constraints.


nseinlet:
@blabllavita to improve perf

pedrobaeza (approved):
I think these changes can be valuable, and more coming from the Odoo's performance expert!

PR #587, opened 2019-04-23
[11.0][IMP] account_banking_sepa_credit_transfer: sepa and other payments in same pain file

BT-tpiric:
- Moved SEPA indicator from payment order to bank-payment-line
- On generating pain.001 file the lines will be grouped by sepa too and in pain.001-file there will be Payment Information Groups (PmtInf, B-Level) for sepa payments too.

This change should allow to have sepa payments and other payments in same payment order.


davidwul:
Hi, why this PR was closed, is it a bad idea?
This would enable in CH to have the pain001 with SEPA at C-level, which would be much more usable.
Any hint why this has been closed?

PR #591, opened 2019-05-07
[12.0][IMP] account_banking_mandate: bank account deletion + indexes

ThomasBinsfeld:
- Deleting a bank account linked to a valid mandate can lead to inconsistencies.
- Indexes on bank account and partner.

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -36,7 +36,9 @@ class AccountBankingMandate(models.Model):
     partner_bank_id = fields.Many2one(
         comodel_name='res.partner.bank', string='Bank Account',
         track_visibility='onchange',
-        domain=lambda self: self._get_default_partner_bank_id_domain(),)
+        domain=lambda self: self._get_default_partner_bank_id_domain(),
+        ondelete='restrict',
+    )
     partner_id = fields.Many2one(
         comodel_name='res.partner', related='partner_bank_id.partner_id',
         string='Partner', store=True)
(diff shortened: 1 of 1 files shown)


sbidoul:
@ThomasBinsfeld looks good, but why not ondelete=restrict? Deleting a bank account is a rather rare operation, so the additional code for a nice error message maybe overkill.

ThomasBinsfeld:
@sbidoul Because deleting a mandate is not possible if it has been used. A `ondelete='restrict'` would make bank accounts deletion impossible too.

sbidoul (approved):

pedrobaeza:
Why the extra commit?

ThomasBinsfeld:
@pedrobaeza I noticed that these two indexes would be relevant while navigating through the code. It's not really linked to bank account deletion so i committed it separately.

pedrobaeza:
Well, but there's no text talking about that in the PR. I'll change the PR description and comment.

About the restriction, if you add such a code, please add corresponding test. If not, it's tolerable to use `ondelete='restrict'` (I even prefer it, as the check is at DB level, not at ORM).

sbidoul:
I've re-read and I now see it allows deleting bank accounts on cancelled mandates. OTOH if the mandate has been used, it would lose trace of the bank account. 

Did we not have some bank_account_active module somewhere?

ThomasBinsfeld:
> Did we not have some bank_account_active module somewhere?

Found [this](https://github.com/OCA/partner-contact/tree/10.0/partner_bank_active)

pedrobaeza:
Odoo added 2 weeks ago the effective possibility of reordering the bank accounts in core, so disabling them are not strictly needed.

ThomasBinsfeld:
@sbidoul @pedrobaeza Replaced by a `ondelete='restrict'`.

pedrobaeza (approved):

sbidoul (approved):

PR #593, opened 2019-05-08
[12.0] [IMP] Banking Mandate: set reference at creation if falsy value in dict

### LINKED ISSUE #592: [12.0] [RFC] Banking Mandate: unique reference at validation
It's currently possible to import a mandate with no reference and validate it.

What about making the `unique_mandate_reference` readonly (in Python code, not only in views) and generating the reference at validation instead of creation (closer than what is done on invoices) + a constraint making the reference required in other states than draft.

ThomasBinsfeld:
fixes https://github.com/OCA/bank-payment/issues/592

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -186,7 +186,8 @@ class AccountBankingMandate(models.Model):
 
     @api.model
     def create(self, vals=None):
-        if vals.get('unique_mandate_reference', 'New') == 'New':
+        unique_mandate_reference = vals.get('unique_mandate_reference')
+        if not unique_mandate_reference or unique_mandate_reference == 'New':
             vals['unique_mandate_reference'] = \
                 self.env['ir.sequence'].next_by_code(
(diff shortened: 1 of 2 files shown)


account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -183,7 +183,7 @@ def _check_valid_state(self):
 
     @api.model
     def create(self, vals=None):
-        if vals.get('unique_mandate_reference', 'New') == 'New':
+        if not vals.get('unique_mandate_reference'):
```
sbidoul: @ThomasBinsfeld if the user provides `New` I think you must preserve the current behaviour.

ThomasBinsfeld: Done

guewen (approved):

pedrobaeza:
Can you add a test for avoiding regressions?

ThomasBinsfeld:
@pedrobaeza Done

pedrobaeza (approved):

PR #599, opened 2019-05-15
[10.0][ADD] account_invoice_cancel_with_payment_order

rousseldenis:


rousseldenis:
@chrisdec

account_invoice_cancel_with_payment_order/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        return payment_lines
+
+    @api.multi
+    def check_payment_line_linked(self):
+        self.ensure_one()
+        payment_lines = self.get_payment_line_linked()
+        if payment_lines:
+            return True
+        else:
+            return False
```
sbidoul, line 29: this method is not really useful, IMO

account_invoice_cancel_with_payment_order/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+
+from odoo import api, models, _
+
+
+class AccountInvoice(models.Model):
+
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def get_payment_line_linked(self):
```
sbidoul, line 13:
```suggestion
    def _get_payment_line_linked(self):
```

account_invoice_cancel_with_payment_order/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        payment_line_obj = self.env['account.payment.line']
+        payment_lines = []
+        if self.move_id.line_ids:
+            inv_mv_lines = [x.id for x in self.move_id.line_ids]
```
sbidoul, line 17: use `.ids`, use `mapped` for recordsets of multiple records

account_invoice_cancel_with_payment_order/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        payment_lines = self.get_payment_line_linked()
+        if payment_lines:
+            return True
+        else:
+            return False
+
+    @api.multi
+    def action_invoice_cancel(self):
+        for invoice in self:
+            if invoice.check_payment_line_linked():
```
sbidoul, line 34:
```suggestion
            if invoice._get_payment_line_linked():
```

account_invoice_cancel_with_payment_order/wizards/validate_invoice_cancel.xml
```diff
@@ -0,0 +1,29 @@
...
+    <record model="ir.ui.view" id="validate_invoice_cancel_form_view">
+        <field name="name">validate.invoice.cancel.form (in account_invoice_cancel_with_payment_order)</field>
+        <field name="model">validate.invoice.cancel</field>
+        <field name="arch" type="xml">
+            <form string="Validate Invoice Cancel">
+                <group>
+                   <p>WARNING ! You try to cancel an invoice related to a payment order.</p>
+                </group>
+                 <group groups="account.group_account_manager">
+                   <p>This process can be dangerous. You assume the possible consequences of this action.</p>
```
sbidoul, line 16: It would be nice to give a hint about the consequences, otherwise it's a bit stressful to be a user :)

Also, I think the behaviour must be different depending on the state of the payment order.
Indeed, if the linked payment line is in a payment order that is not sent, I tend to think the consequence is minimal.

If the payment line is in a payment order that has not been sent, the payment line is not reconciled with the move line and when the invoice is cancelled, and later the payment sent, the invoice will not be reconciled, so in that case I'd say it should be forbidden to cancel the invoice, and the user must first remove it from the payment order.

sbidoul (changes requested):
A few changes and a big question.

pedrobaeza:
@rousseldenis maybe you can check #516 for a different approach of the same problem to check which one seems better

ecino:
Why adding this module when this one already exists? https://github.com/OCA/bank-payment/tree/10.0/account_payment_line_cancel

I have feeling it does the same but not handling every possible situation. If there are some missing parts in the `account_payment_line_cancel` why not proposing changes to this module instead of creating a new one? Not sure to understand the benefits of this one.

PR #603, opened 2019-05-21
[12.0][REF] account_payment_purchase: make independent from stock

astirpe:
In Odoo V12 the module `purchase` does not depend from module `stock` anymore.

This PR is a proposal to split module `account_payment_purchase` in order to make it not dependent from `stock` as well. This will avoid to  force the installation of `stock` if not needed, saving extra costs of purchasing an unused Odoo app.

For this purpose, a new auto-installable module is added: `account_payment_purchase_stock`.


MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

JordiBForgeFlow (approved):

pedrobaeza (changes requested):
I don't agree with this. This change can't be done in a stable version. Do it for 13, but not here. It will break installations.

astirpe:
Thanks!

MiquelRForgeFlow:
Included in v13 migration https://github.com/OCA/bank-payment/pull/674 😉

astirpe:
Shall I close this PR?

PR #605, opened 2019-05-21
[12.0][MIG] account_voucher_killer

astirpe:


pedrobaeza:
Why this module here in this repo? It has nothing to do with bank. I think this should go to OCA/account-payment

astirpe:
Moved to https://github.com/OCA/account-payment/pull/254

PR #606, opened 2019-05-21
[IMP] - Add default value for batch_booking

sbejaoui:
This PR add a default value for account payment order based on account payment mode setting.


account_banking_pain_base/models/account_payment_mode.py
```diff
@@ -33,3 +33,8 @@ class AccountPaymentMode(models.Model):
         "the PAIN files generated by Odoo. This value is determined by the "
         "financial institution that will process the file. If not defined, "
         "no scheme will be used.\n")
+
+    default_batch_booking = fields.Boolean(
+        string='Default Batch Booking',
+        help="Default value for payment order batch booking related "
+             "to this payment mode")
```
sbidoul, line 40: Don't you need to add the field to the view too?

sbejaoui: yes

sbidoul:
@sbejaoui batch_booking is defined in account_banking_pain_base, so I think the default value must be defined in the same module no?

sbejaoui:
@sbidoul , I did it in the first place, but I realized after that there was no dependency between`account_payment_order` and `account_banking_pain_base`.

https://github.com/OCA/bank-payment/blob/3c8913335067a2d3a136b50c921c8579f0010fef/account_payment_order/__manifest__.py#L20-L23

I got an attribute error in travis as you can check.

sbidoul:
@sbejaoui account_banking_pain_base depends on account_payment_order, so that should work.

williamphilippin-droid:
LGTM!

PR #608, opened 2019-06-03
[IMP] Create a payment order in the past based on payment mode configuration

OpenCode:
Add the possibility to create payment orders with passed date based on payment mode configuration


daramousk:
I am not sure what the use case is for this one, why would you want to create a payment order in the past? I think there might be legal issues with this as well.

PR #615, opened 2019-07-15
[FIX] test_invoice_refund

rven:
Trying to trigger the OCA runbot

### DIFF
--- a/account_payment_partner/README.rst
+++ b/account_payment_partner/README.rst
@@ -111,6 +111,9 @@ Contributors
 
   * Pedro M. Baeza <pedro.baeza@tecnativa.com>
   * Carlos Dauden <carlos.dauden@tecnativa.com>
+* `DynApps <https://www.dynapps.be>`_:
+
+  * Raf Ven <raf.ven@dynapps.be>
 
 Maintainers
 ~~~~~~~~~~~
--- a/account_payment_partner/readme/CONTRIBUTORS.rst
+++ b/account_payment_partner/readme/CONTRIBUTORS.rst
@@ -9,3 +9,6 @@
 
   * Pedro M. Baeza <pedro.baeza@tecnativa.com>
   * Carlos Dauden <carlos.dauden@tecnativa.com>
+* `DynApps <https://www.dynapps.be>`_:
+
+  * Raf Ven <raf.ven@dynapps.be>
--- a/account_payment_partner/static/description/index.html
+++ b/account_payment_partner/static/description/index.html
@@ -456,6 +456,10 @@ If you spotted it first, help us smashing it by providing a detailed and welcome
 <li>Carlos Dauden &lt;<a class="reference external" href="mailto:carlos.dauden&#64;tecnativa.com">carlos.dauden&#64;tecnativa.com</a>&gt;</li>
 </ul>
 </li>
+<li><a class="reference external" href="https://www.dynapps.be">DynApps</a>:<ul>
+<li>Raf Ven &lt;<a class="reference external" href="mailto:raf.ven&#64;dynapps.be">raf.ven&#64;dynapps.be</a>&gt;</li>
+</ul>
+</li>
 </ul>
 </div>
 <div class="section" id="maintainers">
--- a/account_payment_partner/tests/test_account_payment_partner.py
+++ b/account_payment_partner/tests/test_account_payment_partner.py
@@ -307,7 +307,8 @@ class TestAccountPaymentPartner(common.SavepointCase):
         self.assertEquals(invoice_refund.payment_mode_id,
                           invoice.payment_mode_id)
         self.assertEquals(invoice_refund.partner_bank_id,
-                          invoice.partner_bank_id)
+                          self.env.ref(
+                              'account_payment_mode.main_company_iban'))
 
     def test_partner(self):
         self.customer.write({


pedrobaeza:
Why this?

rven:
Latest version of Odoo gives following error:

2019-07-15 12:24:20,457 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: FAIL: test_invoice_refund (odoo.addons.account_payment_partner.tests.test_account_payment_partner.TestAccountPaymentPartner) 
2019-07-15 12:24:20,457 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: Traceback (most recent call last): 
2019-07-15 12:24:20,457 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: `   File "/opt/odoo/buildout/parts/oca/bank-payment/account_payment_partner/tests/test_account_payment_partner.py", line 310, in test_invoice_refund 
2019-07-15 12:24:20,457 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: `     invoice.partner_bank_id) 
2019-07-15 12:24:20,458 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: ` AssertionError: res.partner.bank(38,) != res.partner.bank() 
2019-07-15 12:24:20,458 18065 INFO odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: Ran 12 tests in 6.216s 
2019-07-15 12:24:20,458 18065 ERROR odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner: FAILED 
2019-07-15 12:24:20,458 18065 INFO odoo_test odoo.addons.account_payment_partner.tests.test_account_payment_partner:  (failures=1) 
2019-07-15 12:24:20,458 18065 INFO odoo_test odoo.modules.module: odoo.addons.account_payment_partner.tests.test_account_payment_partner tested in 6.22s, 4085 queries 
2019-07-15 12:24:20,458 18065 ERROR odoo_test odoo.modules.module: Module account_payment_partner: 1 failures, 0 errors

rven:
image

pedrobaeza:
OK, then you should fix the problem instead of only adding you as contributor...

account_payment_partner/readme/CONTRIBUTORS.rst
```diff
@@ -9,3 +9,5 @@
 
   * Pedro M. Baeza <pedro.baeza@tecnativa.com>
   * Carlos Dauden <carlos.dauden@tecnativa.com>
+* `DynApps <https://www.dynapps.be>`_:
```
pedrobaeza, line 12: You must leave an empty line before and after for proper RST syntax

pedrobaeza:
Aren't you going to fix it?

rven:
Yes I am

pedrobaeza:
OK, thanks :smiley:

account_payment_partner/README.rst
```diff
@@ -111,6 +111,9 @@ Contributors
 
   * Pedro M. Baeza <pedro.baeza@tecnativa.com>
   * Carlos Dauden <carlos.dauden@tecnativa.com>
+* `DynApps <https://www.dynapps.be>`_:
```
pedrobaeza, line 114: Leave an empty line before

rven: image
Before is not needed?

pedrobaeza: OK, I see... I though it was, sorry.

pedrobaeza (approved):

account_payment_partner/README.rst
rven: It has no influence on the result...

pedrobaeza (changes requested):
Hey, wait, I'm seeing that there's specific code that tries to avoid this behavior:

https://github.com/OCA/bank-payment/blob/7c5cb9638b4139fe25e1df605f1b4f90345aab2a/account_payment_partner/models/account_invoice.py#L95-L109

so we shouldn't reset the test, but fixing the behavior

pedrobaeza:
Or should we leave the same bank account? @sbidoul @alexis-via can you check this new upstream behavior?

rven:
Isn't it ok.

```python
    def _get_default_bank_id(self, type, company_id):
        """When setting the default bank account, we have two cases:
         in the case of 'out_invoice', 'in_refund', we want the bank account of the company.
         otherwise, we want the bank account of the partner.
        """
        if type not in ('out_invoice', 'in_refund') or not company_id:
            return False
        return self._get_partner_bank_id(company_id)
```

pedrobaeza:
Any news, @sbidoul @alexis-via ?

sbidoul:
So if I understand this thread correctly, Odoo changed behaviour when creating refunds from invoices?

I would say the test is correct (i.e. preserve the bank account from the invoice in the refund), for the same reason it makes sense to preserve the invoice payment mode in the refund.

So at first sight, I'd say:
- we need to fix the behaviour and preserve the test
- raise the issue at Odoo?
- see what it does for customer invoices (in case of direct debit): IMO the bank account must be preserved in refunds too

pedrobaeza:
Yeah, in a nutshell, Odoo considers that on refund, it should be changed, and thus, the error in test. Personally, I don't care too much about bank account indicated there, as it's almost ignored in all places by bank-payment/l10n-spain suite. But it seems others like @alexis-via relies a lot on that data.

We must discard option 2, as Odoo has made this change because they think the contrary, so the question is to merge this PR, assuming the new behavior is correct, changing the test, or overwrite behavior for keeping old one, and don't change test.

sbidoul:
I'm not sure Odoo made this change intentionally. As a default value, their change can make sense, but when creating a refund from an invoice the default may not apply the same way.

In any case I vote for keeping the test as is and fix the behaviour, to be consistent with the copy of the payment mode.

pedrobaeza:
If you see the source commit, it was an OPW fixed that way.

pedrobaeza:
Let's fix the branch for now merging this and if we change the behavior later, we can revert back to the original test.

PR #618, opened 2019-07-22
[10.0][ADD] account_payment_mode_auto_reconcile

grindtildeath:
This module adds a checkbox 'auto_reconcile_outstanding_credits' on account
payment modes to allow automatic reconciliation on account invoices if it is
checked.

Automatic reconciliation of outstanding credits will only happen on customer
invoices at validation if the payment mode is set or when the payment mode is
changed on an open invoice.

### DIFF
--- a/account_payment_mode_auto_reconcile/models/account_invoice.py
+++ b/account_payment_mode_auto_reconcile/models/account_invoice.py
@@ -23,8 +23,8 @@ class AccountInvoice(models.Model):
     )
 
     @api.multi
-    def action_invoice_open(self):
-        res = super(AccountInvoice, self).action_invoice_open()
+    def invoice_validate(self):
+        res = super(AccountInvoice, self).invoice_validate()
         for invoice in self:
             if invoice.type != 'out_invoice':
                 continue
--- a/account_payment_mode_auto_reconcile/models/account_invoice.py
+++ b/account_payment_mode_auto_reconcile/models/account_invoice.py
@@ -30,7 +30,7 @@ class AccountInvoice(models.Model):
                 continue
             if not invoice.payment_mode_id.auto_reconcile_outstanding_credits:
                 continue
-            partial_allowed = self.payment_mode_id.auto_reconcile_allow_partial
+            partial_allowed = invoice.payment_mode_id.auto_reconcile_allow_partial
             invoice.with_context(
                 _payment_mode_auto_reconcile=True
             ).auto_reconcile_credits(partial_allowed=partial_allowed)
--- a/account_payment_mode_auto_reconcile/models/account_invoice.py
+++ b/account_payment_mode_auto_reconcile/models/account_invoice.py
@@ -30,10 +30,10 @@ class AccountInvoice(models.Model):
                 continue
             if not invoice.payment_mode_id.auto_reconcile_outstanding_credits:
                 continue
-            partial_allowed = invoice.payment_mode_id.auto_reconcile_allow_partial
+            partial = invoice.payment_mode_id.auto_reconcile_allow_partial
             invoice.with_context(
                 _payment_mode_auto_reconcile=True
-            ).auto_reconcile_credits(partial_allowed=partial_allowed)
+            ).auto_reconcile_credits(partial_allowed=partial)
         return res
 
     @api.multi
@@ -50,11 +50,11 @@ class AccountInvoice(models.Model):
                     payment_mode
                     and payment_mode.auto_reconcile_outstanding_credits
                 ):
-                    partial_allowed = payment_mode.auto_reconcile_allow_partial
+                    partial = payment_mode.auto_reconcile_allow_partial
                     invoice.with_context(
                         _payment_mode_auto_reconcile=True
                     ).auto_reconcile_credits(
-                        partial_allowed=partial_allowed
+                        partial_allowed=partial
                     )
                 # If the payment mode is not using auto reconcile we remove
                 #  the existing reconciliations
--- a/account_payment_mode_auto_reconcile/models/account_invoice.py
+++ b/account_payment_mode_auto_reconcile/models/account_invoice.py
@@ -31,7 +31,9 @@ class AccountInvoice(models.Model):
             if not invoice.payment_mode_id.auto_reconcile_outstanding_credits:
                 continue
             partial_allowed = self.payment_mode_id.auto_reconcile_allow_partial
-            invoice.auto_reconcile_credits(partial_allowed=partial_allowed)
+            invoice.with_context(
+                _payment_mode_auto_reconcile=True
+            ).auto_reconcile_credits(partial_allowed=partial_allowed)
         return res
 
     @api.multi
@@ -49,7 +51,9 @@ class AccountInvoice(models.Model):
                     and payment_mode.auto_reconcile_outstanding_credits
                 ):
                     partial_allowed = payment_mode.auto_reconcile_allow_partial
-                    invoice.auto_reconcile_credits(
+                    invoice.with_context(
+                        _payment_mode_auto_reconcile=True
+                    ).auto_reconcile_credits(
                         partial_allowed=partial_allowed
                     )
                 # If the payment mode is not using auto reconcile we remove
@@ -86,7 +90,16 @@ class AccountInvoice(models.Model):
                 aml = self.env['account.move.line'].browse(
                     payment.get('payment_id')
                 )
-                aml.with_context(invoice_id=invoice.id).remove_move_reconcile()
+                for apr in aml.matched_debit_ids:
+                    if apr.amount != payment.get('amount'):
+                        continue
+                    if (
+                        apr.payment_mode_auto_reconcile
+                        and apr.debit_move_id.invoice_id == invoice
+                    ):
+                        aml.with_context(
+                            invoice_id=invoice.id
+                        ).remove_move_reconcile()
 
     @api.depends(
         'type', 'payment_mode_id', 'payment_move_line_ids', 'state',
@@ -115,7 +128,8 @@ class AccountInvoice(models.Model):
                 )
             ):
                 invoice.payment_mode_warning = _(
-                    'Changing payment mode will unreconcile existing payments.'
+                    'Changing payment mode will unreconcile existing auto '
+                    'reconciled payments.'
                 )
             elif (
                 invoice.state == 'open' and not invoice.payment_move_line_ids
--- a/account_payment_mode_auto_reconcile/models/account_invoice.py
+++ b/account_payment_mode_auto_reconcile/models/account_invoice.py
@@ -1,0 +1,84 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import json
+
+from odoo import api, fields, models
+
+
+class AccountInvoice(models.Model):
+
+    _inherit = "account.invoice"
+
+    # Allow changing payment mode in open state
+    # TODO: Check if must be done in account_payment_partner instead
+    payment_mode_id = fields.Many2one(
+        states={'draft': [('readonly', False)], 'open': [('readonly', False)]}
+    )
+    auto_assigned_oustanding_credits = fields.Boolean(
+        compute='_compute_auto_assigned_oustanding_credits',
+    )
+
+    @api.multi
+    def action_invoice_open(self):
+        res = super(AccountInvoice, self).action_invoice_open()
+        for invoice in self:
+            if not invoice.payment_mode_id.auto_reconcile_outstanding_credits:
+                continue
+            partial_allowed = self.payment_mode_id.auto_reconcile_allow_partial
+            invoice.auto_reconcile_credits(partial_allowed=partial_allowed)
+        return res
+
+    @api.onchange('payment_mode_id')
+    def payment_mode_id_change(self):
+        super(AccountInvoice, self).payment_mode_id_change()
+        if self.state != 'open':
+            return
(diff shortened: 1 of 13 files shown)


JordiBForgeFlow:
Could you tell me what is the use case? Why would one auto-reconcile at the time of validating the invoice, with no control on what to reconcile against? Thanks!

grindtildeath:
@jbeficent The company is using Direct Debit orders to get payments from selected customers and the idea is to reconcile any existing credit notes for such customers before the invoice is sent and the debit order is created. We also backported v12.0 report to print the invoices with payments (https://github.com/OCA/account-invoicing/pull/567) so that their customers receive the amount that will be debited.

JordiBForgeFlow:
Thanls for the clarification. Will review the PR

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+    def auto_reconcile_credits(self, partial_allowed=True):
+        for invoice in self:
+            if not invoice.has_outstanding:
+                continue
+            credits_info = json.loads(
+                invoice.outstanding_credits_debits_widget
+            )
+            # Get outstanding credits in chronological order
+            # (using reverse because aml is sorted by date desc as default)
+            credits = credits_info.get('content')
```
i-vyshnevska: seems credits are reserved word

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+                continue
+            if not invoice.payment_mode_id.auto_reconcile_outstanding_credits:
+                continue
+            partial_allowed = self.payment_mode_id.auto_reconcile_allow_partial
+            invoice.auto_reconcile_credits(partial_allowed=partial_allowed)
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super(AccountInvoice, self).write(vals)
```
i-vyshnevska, line 41: shouldn't  it would be performed before write?

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+                payment_mode = invoice.payment_mode_id
+                if (
+                    payment_mode
+                    and payment_mode.auto_reconcile_outstanding_credits
+                ):
+                    partial_allowed = payment_mode.auto_reconcile_allow_partial
+                    invoice.auto_reconcile_credits(
+                        partial_allowed=partial_allowed
+                    )
+                elif invoice.payment_move_line_ids:
```
i-vyshnevska, line 61: some comments pls, seems we call auto_unreconcile_credits for  default payment_mode is it correct?

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
```diff
@@ -0,0 +1,151 @@
...
+                'product_id': self.product.id,
+                'name': self.product.name,
+                'price_unit': 500.0,
+                'quantity': 1,
+                'account_id': self.acc_rev.id,
+            })]
+        })
+        self.assertTrue(self.payment_mode.auto_reconcile_outstanding_credits)
+        self.assertEqual(self.invoice_copy.residual, 1500)
+        auto_rec_invoice.action_invoice_open()
```
i-vyshnevska, line 97: should this behavior run when we go from open -> done?

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+            invoice.display_payment_mode_warning = True
+            if (
+                invoice.state != 'open' and invoice.payment_mode_id and
+                invoice.payment_mode_id.auto_reconcile_outstanding_credits
+            ):
+                invoice.payment_mode_warning = _(
+                    'Validating invoices with this payment mode will reconcile'
+                    ' any outstanding credits.'
+                )
+                invoice.display_payment_mode_warning = True
```
jcoux: Already done line 98.

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+            elif (
+                invoice.state == 'open' and invoice.payment_move_line_ids and (
+                    not invoice.payment_mode_id or not
+                    invoice.payment_mode_id.auto_reconcile_outstanding_credits
+                )
+            ):
+                invoice.payment_mode_warning = _(
+                    'Changing payment mode will unreconcile existing payments.'
+                )
+                invoice.display_payment_mode_warning = True
```
jcoux: Already done line 98.

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+            elif (
+                invoice.state == 'open' and not invoice.payment_move_line_ids
+                and invoice.payment_mode_id
+                and invoice.payment_mode_id.auto_reconcile_outstanding_credits
+                and invoice.has_outstanding
+            ):
+                invoice.payment_mode_warning = _(
+                    'Changing payment mode will reconcile outstanding credits.'
+                )
+                invoice.display_payment_mode_warning = True
```
jcoux: Already done line 98.

jcoux (commented):
Just technical review.
I don't know if functionally it's good or not.

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
grindtildeath: I don't really understand your question. The auto reconciliation will be done when the invoice is open, and if there are enough outstanding credit to reconcile, the invoice will be set to done...

account_payment_mode_auto_reconcile/models/account_payment_mode.py
```diff
@@ -0,0 +1,19 @@
...
+class AccountPaymentMode(models.Model):
+
+    _inherit = "account.payment.mode"
+
+    auto_reconcile_outstanding_credits = fields.Boolean(
+        help="Reconcile automatically outstanding credits when an invoice "
+             "using this payment mode is validated, or when this payment mode "
+             "is defined on an open invoice."
+    )
+    auto_reconcile_allow_partial = fields.Boolean(
```
TDu, line 17: There is no mention of this in the description. It could help

account_payment_mode_auto_reconcile/models/account_invoice.py
grindtildeath: We can't modify `vals` to do the reconciliation, so better do it afterwards

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
i-vyshnevska: if invoice wasn't fully reconciled and after new lines appeared, they will not be added here
if this action should performed only once than ok,
 maybe it can be done in action_invoice_validate and unreconcilation in action_invoice_cancel instead of write method

account_payment_mode_auto_reconcile/models/account_invoice.py
i-vyshnevska: "is not using auto reconcile" we don't check it here, we drop lines always

jcoux (approved):
Technically good for me :+1:

account_payment_mode_auto_reconcile/models/account_invoice.py
grindtildeath: If we get there, it means the payment mode is not set to auto-reconcile (otherwise it would pass the `if` above)

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
grindtildeath: As written in the description "Automatic reconciliation of outstanding credits will only happen on customer
invoices at validation if the payment mode is set or when the payment mode is
changed on an open invoice."

account_payment_mode_auto_reconcile/models/account_payment_mode.py
grindtildeath: Attended in 1e2ebca

account_payment_mode_auto_reconcile/models/account_invoice.py
i-vyshnevska: exactly, mode is not in 'auto-reconcile', so how do we now was this lines added by auto reconcile or manually?
this will applly one policy for all invoices

tschanzt (approved):

TDu (approved):

account_payment_mode_auto_reconcile/models/account_invoice.py
grindtildeath: I thought I'd leave it so, but now fixed in f519930

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
```diff
@@ -0,0 +1,193 @@
...
+
+from odoo.tests import SavepointCase
+from odoo.tools import DEFAULT_SERVER_DATE_FORMAT as DATE_FORMAT
+
+
+class TestPartnerAutoReconcile(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestPartnerAutoReconcile, cls).setUpClass()
```
i-vyshnevska, line 15: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))

i-vyshnevska (approved):

jbaudoux:
The communication in the direct debit contains the move line reference if not empty, otherwise it contains the move name. For invoices with direct debit, you would configure the communication as free. So the move line does not get a communication and the payment line will contain the invoice number (i.e. move name) which is the most meaningful for the customer as there are no debit advice sent to the bank.
When you deduct a credit note at invoice confirmation, you better notify the customer. Easiest would be to write in the invoice and move line reference the move name (invoice number) and list of credit notes that have been registered as payment. This will allow the customer to understand why the total amount of the invoice has not been debited.

account_payment_mode_auto_reconcile/models/account_invoice.py
```diff
@@ -82,6 +84,16 @@ def auto_reconcile_credits(self, partial_allowed=True):
...
                 invoice.assign_outstanding_credit(credit.get('id'))
 
+    @api.multi
+    def _filter_payment_same_journal(self, credits_dict):
+        """Keep only credits on the same journal than the invoice."""
+        self.ensure_one()
+        line_ids = [credit['id'] for credit in credits_dict]
+        lines = self.env['account.move.line'].browse(line_ids).filtered(
+            lambda line: line.journal_id == self.journal_id
+        )
```
sebalix: @TDu To get better performance (don't know how many records we could process here), could you rewrite this using `search` instead of `browse`+`filtered`? That way no prefetch will be done on the database as you are also only using `line.ids` right after.

TDu: Thanks for your review, no compromise on performance, fixed.

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
```diff
@@ -0,0 +1,235 @@
...
+            })]
+        })
+        self.assertTrue(self.payment_mode.auto_reconcile_outstanding_credits)
+        self.assertEqual(self.invoice_copy.residual, 1500)
+        auto_rec_invoice.action_invoice_open()
+        self.assertEqual(auto_rec_invoice.residual, 500)
+
+    def test_invoice_auto_reconcile_different_journal(self):
+        """Check not reconciling credits on different journal."""
+        self.payment_mode.auto_reconcile_same_journal = True
```
i-vyshnevska, line 218:
```suggestion
        self.payment_mode.auto_reconcile_same_journal = False
```
?

account_payment_mode_auto_reconcile/models/account_payment_mode.py
```diff
@@ -0,0 +1,26 @@
...
+        default=True,
+        string="Allow partial",
+        help="Allows automatic partial reconciliation of outstanding credits",
+    )
+    auto_reconcile_same_journal = fields.Boolean(
```
i-vyshnevska, line 22: instead of forcing reconciliation only under one journal it should fall to best match based on amount and date 
how for instance it will run reconciliation if under one journal there will be 2 payments for 200 and 300 with invoice on 300?
and forcing to one journal will exclude reconciliation of payment with different currencies, which can be useful, it's better to match partial payment in different currency than didn't match anything

pedrobaeza:
Shouldn't this go to account-reconcile repo?

account_payment_mode_auto_reconcile/tests/test_partner_auto_reconcile.py
TDu: No this is correct the test is checking that when the option is set credits on a different journal is not reconciled.

account_payment_mode_auto_reconcile/models/account_payment_mode.py
TDu: The truth is : this is a request from the consultants that have been thinking hard to please their customer. I implemented the solution they thought was the best fit.
Also this is only an option that can be activated or not !

grindtildeath:
@pedrobaeza I agree, I'll open a new PR there asap :wink:

PR #620, opened 2019-07-23
[FIX] account_payment_partner: remove wrong code line

ernestotejeda:
I don't know why this line of code was added (https://github.com/OCA/bank-payment/blob/7c5cb9638b4139fe25e1df605f1b4f90345aab2a/account_payment_partner/tests/test_account_payment_partner.py#L292), that breaks the tests

Cc @Tecnativa


pedrobaeza:
This is being handled in #615

PR #621, opened 2019-08-02
[FIX] account_payment_order: TypeError: a bytes-like object is required, not str

smangukiya:
Account Payment Order form view, click on Generate file button and it will give you below error:
```
/oca-addons/bank-payment/account_payment_order/models/account_payment_order.py", line 339, in open2generated
    'datas': base64.b64encode(payment_file_str),
  File "/usr/lib/python3.6/base64.py", line 58, in b64encode
    encoded = binascii.b2a_base64(s, newline=False)
TypeError: a bytes-like object is required, not 'str'
```

@max3903

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -332,6 +332,7 @@ class AccountPaymentOrder(models.Model):
         payment_file_str, filename = self.generate_payment_file()
         action = {}
         if payment_file_str and filename:
+            payment_file_str = payment_file_str.encode()
             attachment = self.env['ir.attachment'].create({
                 'res_model': 'account.payment.order',
                 'res_id': self.id,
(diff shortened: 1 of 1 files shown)


max3903:
@smangukiya Tests are failing:
https://travis-ci.org/OCA/bank-payment/jobs/567107608#L2033

smangukiya:
> @smangukiya Tests are failing:
> https://travis-ci.org/OCA/bank-payment/jobs/567107608#L2033

@max3903 : Added check to encode string type variable only

pedrobaeza:
I'm not sure this patch should be land here or in the specific exporter. With which file format are you trying before this patch?

smangukiya:
> patch

Hi @pedrobaeza : 
I am getting an error (shown below) when we click on "Generate Payment File" button from Invoicing>Payments>Payment Orders form view to generate ACH file:

image

pedrobaeza:
What is ACH? The error will be probably there, not in this module.

max3903:
@pedrobaeza the US equivalent to SEPA.

pedrobaeza:
Then I insist: the problem should be on that module, not here, as SEPA is generated without problem, and if you change this, you will probably break SEPA.

max3903:
@smangukiya Any progress on this? Issue may be with the ach modules in https://github.com/OCA/l10n-usa

PR #622, opened 2019-08-05
[10.0][IMP] account_banking_pain_base: Provide a more detailed message when _prepare_field fails

SimoRubi:
This PR aims to provide more details to user that gets a `ValidationError` while generating the Payment file.
Here I also allow inheriting modules to customize the exception management by overriding the new method `except_prepare_field`.

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -100,15 +100,23 @@ class AccountPaymentOrder(models.Model):
                 for unallowed_ascii_char in unallowed_ascii_chars:
                     value = value.replace(unallowed_ascii_char, '-')
         except:
-            line = eval_ctx.get('line')
-            if line:
-                raise UserError(
-                    _("Cannot compute the '%s' of the Payment Line with "
-                        "reference '%s'.")
-                    % (field_name, line.name))
-            else:
-                raise UserError(
-                    _("Cannot compute the '%s'.") % field_name)
+            error_msg_prefix = _("Cannot compute the field '{field_name}'.") \
+                .format(
+                field_name=field_name)
+
+            error_msg_details_list = self.except_messages_prepare_field(
+                eval_ctx, field_name)
+            error_msg_data = _("Data for evaluation:\n"
+                               "\tcontext: {eval_ctx}\n"
+                               "\tfield path: {field_value}") \
+                .format(
+                eval_ctx=eval_ctx,
+                field_value=field_value)
+            raise UserError('\n'.join(
+                [error_msg_prefix]
+                + error_msg_details_list
+                + [error_msg_data]))
+
         if not isinstance(value, (str, unicode)):
             raise UserError(
                 _("The type of the field '%s' is %s. It should be a string "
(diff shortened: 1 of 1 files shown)


SimoRubi:
@OCA/banking-maintainers can someone please take a look? Thanks

pedrobaeza (commented):
Why adding that weird method that is called from only one place? Do all in the same except

And I wouldn't give more priority to the line variable or partner_bank, just add that 2 things to the general exception message if found.

SimoRubi:
Hi, thanks for your review!

* > Why adding that weird method that is called from only one place? Do all in the same except

  The method is added in order to allow other modules that override `_prepare_field` to provide more 
  detailed messages, since the method is 
  > designed to be inherited !

* > And I wouldn't give more priority to the line variable or partner_bank, just add that 2 things to the 
  general exception message if found.

  ok, I'm going to change that then

pedrobaeza:
> The method is added in order to allow other modules that override _prepare_field to provide more
detailed messages, since the method is designed to be inherited !

Then please put that in the docstring in order to know why it's done that way.

For me, the design should be to return a message in case of error, and trigger the exception downwards in case of any.

SimoRubi:
docstring and design modified

SimoRubi:
I'll fix everything up into one commit when it's ok

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -122,6 +125,28 @@ def _prepare_field(self, field_name, field_value, eval_ctx,
...
+        """
+        Inherit this method to provide more detailed error messages for
+        exceptions to be raised while evaluating `field_name` using `eval_ctx`.
+        :return: List containing the error messages.
+        """
+        error_messages = list()
+        line = eval_ctx.get('line')
+        if line:
+            error_messages.append(
+                _("Cannot compute the '%s' of the "
```
pedrobaeza: This text is redundant across messages. I was thinking of joining all the messages (including the one you have put right now outside), but only adding the significant part (second line).

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -100,15 +100,18 @@ def _prepare_field(self, field_name, field_value, eval_ctx,
...
-                    _("Cannot compute the '%s' of the Payment Line with "
-                        "reference '%s'.")
-                    % (field_name, line.name))
-            else:
-                raise UserError(
-                    _("Cannot compute the '%s'.") % field_name)
+            error_msg_list = self.except_messages_prepare_field(
+                eval_ctx, field_name)
+            if error_msg_list:
+                raise UserError('\n'.join(error_msg_list))
```
pedrobaeza: I'm thinking that this is not needed. Only append this to the general one, removing redundant part.

SimoRubi:
@pedrobaeza I tried to understand what you were suggesting, here is what I implemented.
Does it look better now?

pedrobaeza:
Great, you have got the idea! Thanks for the efforts.

I have fixed Travis file, so you can rebase for having green CI.

SimoRubi:
Rebased and fixed up

pedrobaeza (approved):
Great work! This way, we will have a better error message understanding. @sbidoul @alexis-via do you agree with this change?

tafaRU (approved):

SimoRubi:
> Great work! This way, we will have a better error message understanding. @sbidoul @alexis-via do you agree with this change?

Yep, and hopefully the user will manage to solve the issue without contacting us!

pedrobaeza:
Yes, I have several tickets about these cases. @sbidoul @alexis-via I recall you for checking this useful improvement.

SimoRubi:
@OCA/banking-maintainers can we please merge this? Thanks!

pedrobaeza:
As no more feedback is provided, let's continue with the merge: /ocabot merge patch

@SimoRubi can you please forward-port to later versions?

SimoRubi:
> As no more feedback is provided, let's continue with the merge: /ocabot merge patch
> 
> @SimoRubi can you please forward-port to later versions?

:+1:

PR #630, opened 2019-09-09
[12.0][FIX] account_payment_sale: Propagate payment_mode to invoices create…

### LINKED ISSUE #536: [12.0][BUG] account_payment_sale
Hi
I have found a bug in module account_payment_sale, but I am not sure how to fix it nicely.
The payment_mode_id does not propagate from the sale order to the invoice. 
I guess the tests are a bit to naive, that is why they pass anyway.
Here we try to propagate the payment mode : https://github.com/OCA/bank-payment/blob/12.0/account_payment_sale/models/sale_order.py#L35
Here, the invoice is created with the right value (coming from the SO) : https://github.com/OCA/OCB/blob/12.0/addons/sale/models/sale.py#L521
And it is overriden here https://github.com/OCA/OCB/blob/12.0/addons/sale/models/sale.py#L570

I really don't get why they have refactored it this way, they create the invoice and then they override a lot of values...
And I do not really see a clean solution to solve this.
Any idea?

lmignon:
…d from a call to sale.action_invoice_create

Into the normal process of the invoice creation from a sale order by odoo, the onchange_partner is called. In some case, this call will reset the specific payment_mode and/or partner_bank_id set on the sale_order.
To fix this issue, we must override the _finalize_invoices method, take a snapshot of these values before the call to super and restore these values if these are changed.
fixes #536

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -33,3 +33,31 @@ class SaleOrder(models.Model):
         """Copy bank partner from sale order to invoice"""
         vals = super()._prepare_invoice()
         return self._get_payment_mode_vals(vals)
+
+    def _finalize_invoices(self, invoices, references):
+        """
+        Invoked after creating invoices at the end of action_invoice_create.
+
+        We must override this method since the onchange on partner is called by
+        the base method and therefore will change the specific payment_mode set
+        on the SO if one is defined on the partner..
+
+        :param invoices: {group_key: invoice}
+        :param references: {invoice: order}
+        """
+        payment_vals_by_invoice = {}
+        for invoice in invoices.values():
+            payment_vals_by_invoice[invoice] = {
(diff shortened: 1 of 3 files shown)


lmignon:
@florian-dacosta @simahawk @rousseldenis

lmignon:
@florian-dacosta @simahawk @rousseldenis

florian-dacosta (approved):
LGTM

mlaitinen (commented):
LGTM, but maybe bump the version in the manifest?

lmignon:
@mlaitinen The version will be bumped by ocabot at merge time

mlaitinen (approved):
Right, makes sense. LGTM then!

rousseldenis (approved):
Code review

simahawk (approved):

sebastienbeau (approved):

PR #637, opened 2019-10-09
[REM] Mandate: Remove check on on partner_bank company

qgroulard:
Hi,
I was questioning the point of this check.
Globally, we may argue that the field company_id on a res.partner.bank serves no purpose (except for visibility) since a res.partner.bank is tightly coupled to a res.partner and must be unique among all companies anyway.
Thus, if this field makes no real sense, what's the point of checking it ?

I would appreciate reading your thoughts on that. 
Do you see any reason to have the company_id field on res.partner.bank that I'd miss ? Or to still perform this check ?

Btw, the error message is wrong here. It should say that the company of the mandate differs from the company of the bank account.

qgr

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -129,13 +129,6 @@ class AccountBankingMandate(models.Model):
     @api.constrains('company_id', 'payment_line_ids', 'partner_bank_id')
     def _company_constrains(self):
         for mandate in self:
-            if mandate.partner_bank_id.company_id and \
-                    mandate.partner_bank_id.company_id != mandate.company_id:
-                raise ValidationError(
-                    _("The company of the mandate %s differs from the "
-                      "company of partner %s.") %
-                    (mandate.display_name, mandate.partner_id.name))
-
             if self.env['account.payment.line'].sudo().search(
                     [('mandate_id', '=', mandate.id),
                      ('company_id', '!=', mandate.company_id.id)], limit=1):
--- a/account_banking_mandate/tests/test_mandate.py
+++ b/account_banking_mandate/tests/test_mandate.py
@@ -86,45 +86,6 @@ class TestMandate(TransactionCase):
                 fields.Date.from_string(
                     fields.Date.context_today(mandate)) + timedelta(days=1))
 
-    def test_constrains_02(self):
-        bank_account = self.env.ref('account_payment_mode.res_partner_12_iban')
-        mandate = self.env['account.banking.mandate'].create({
-            'partner_bank_id': bank_account.id,
-            'signature_date': '2015-01-01',
-            'company_id': self.company.id,
-        })
-
-        with self.assertRaises(ValidationError):
-            mandate.company_id = self.company_2
-
-    def test_constrains_03(self):
-        bank_account = self.env.ref('account_payment_mode.res_partner_12_iban')
-        mandate = self.env['account.banking.mandate'].create({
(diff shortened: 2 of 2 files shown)


etobella:
This check is necessary in order to ensure the consistency of the mandate. We know that odoo rules will not allow to show this information, but it is necessary in order to work on multicompany environments.

It is possible to use all the companies at the same time, but it is necessary to add this kind of constrains in order to ensure consistency.

etobella:
However, the description is wrong, as you noticed.

etobella:
Also, this kind of check are necessary on 13.0, as multicompanies can be managed altogether.

So I think there is no sense on removing it from here.

sbidoul:
@etobella our use case is as follow:

- "virtual" company X
- company A and B children of X
- other companies in the database
- partners (and their bank accounts) linked to company X because they are shared by and need to be visible from A and B (but not other companies in the system)
- mandates are linked to company A or B because they are agreements with those real companies

So mandate with company A is related to a bank account linked to company X and the check fails.
But in reality company_id on mandate and company_id on res.partner.bank do not have the same semantics so it would seem it does not make sense to compare them.

What do you think?

pedrobaeza:
@sbidoul maybe you are twisting a bit the company hierarchy for fitting your needs? If I'm not wrong, in v13 that hierarchy is removed. Maybe you need to switch to the approach given by https://github.com/OCA/multi-company/tree/11.0/partner_multi_company, allowing you to limit a partner to discrete companies, instead of using that parent company as wildcard instead.

sbidoul:
@pedrobaeza that's a different approach, but I think the problem with this check we want to remove would remain.

pedrobaeza:
Yeah, it can be... I think you have to remove the test for that part as well then.

etobella:
Removing the option does not seem the best approach, you could make a hook in order to disallow it in your case.

If you are removing it, now it could give problems to other companies.

sbidoul:
@etobella do you remember which are the use case that could cause problem.

To me it's more important to check that the owner of the bank account of the mandate is the same as the owner of the mandate.

If there is a multi-company compatibility check to be done between res.partner.bank and res.partner it should be done elswhere.  And even that I'm not sure is useful. The more I think of it, the more I think the company_id field on res.partner.bank is useless, and if a multi-company record rule is needed on res.partner.bank (there is none in v12 AFAICS), it must be based on the company_id of the res.partner.bank owner.

etobella:
If a partner is only related to a company, its bank account should be related to the same company.
I also think that the field is useless, but the field is defined on odoo base module
https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_bank.py#L90
https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/res_bank.py#L88
and there are rules for it
https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/security/base_security.xml#L60
https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/security/base_security.xml#L48
So I think we should make it work properly using this base.
Also, I think this approach is defined for the case when a partner works for all the companies but every company uses different bank account.

anajuaristi:
IMHO the same bank account of a partner could be used for different companies in a multicompany environment
And different mandates could exists to allow different companies to charge on that bank account. So... I also think the company_id is usseless on Bank account (as said by @etobella) but it's necesary on mandate so the constraint should be removed as I explain on #644

On this aproach I think all cases can be covered and flexibility would heavily be increased.

PR #643, opened 2019-10-29
[FIX] account_payment_order: Fix payment order report. Adapt to bootstrap4

sergio-teruel:
cc @Tecnativa TT20377


pedrobaeza (approved):
Fast-tracking this for solving the layout problem

PR #636, opened 2019-10-01
[MIG] account_payment_mode: Migration to 13.0

lembregtse:
First module to support sepa migration.

Basic changes only.

PR #649, opened 2019-11-08
[MIG] account_payment_partner: Migration to 13.0

rven:
Depends on:

- [x] account_payment_mode https://github.com/OCA/bank-payment/pull/636


rven:
Hi @pedrobaeza,

How can I add the dependency on another PR?
Is this possible with oca_dependencies.txt. If yes, how should it be referenced?

pedrobaeza:
I'm afraid it's not possible with oca_dependencies.txt. One technique is to add the commits of the other module here, and rebase after the merge. Another is to simply wait until the first module is merged, and then rebase. Last one is to modify temporary .travis.yml for download and inject the folder in the build path for having it available. I know some contributors have done it, but I don't have right now any example.

pedrobaeza:
You can rebase now

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # I think copying payment mode from invoice to refund by default
+    # is a good idea because the most common way of "paying" a refund is to
+    # deduct it on the payment of the next invoice (and OCA/bank-payment
+    # allows to have negative payment lines since March 2016)
+    @api.model
+    def _prepare_refund(
+        self, invoice, date_invoice=None, date=None, description=None, journal_id=None
+    ):
+        vals = super(AccountMove, self)._prepare_refund(
```
astirpe: Method `_prepare_refund()` does not exist anymore in V13. You may want to use method `_reverse_moves()` instead.

account_payment_partner/models/account_move_line.py
```diff
@@ -8,9 +8,9 @@ class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     payment_mode_id = fields.Many2one(
-        "account.payment.mode",
+        related="move_id.payment_mode_id",
```
astirpe: Why did you make it related?

astirpe (changes requested):

account_payment_partner/models/account_move_line.py
rven: It's now a computed field

account_payment_partner/models/account_move.py
rven: Indeed, the function is renamed to _reverse_move_vals

account_payment_partner/models/account_payment_mode.py
```diff
@@ -42,29 +42,7 @@ def account_invoice_company_constrains(self):
                 raise ValidationError(
                     _(
                         "You cannot change the Company. There exists "
-                        "at least one Invoice with this Payment Mode, "
-                        "already assigned to another Company."
-                    )
-                )
-
-    @api.constrains("company_id")
-    def account_move_line_company_constrains(self):
```
astirpe: This constraint shouldn't be removed

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+        comodel_name="account.payment.mode",
+        string="Payment Mode",
+        compute=_compute_payment_mode,
+        store=True,
+        readonly=True,
+        ondelete="restrict",
+        states={"draft": [("readonly", False)]},
+    )
+    bank_account_required = fields.Boolean(
+        related="payment_mode_id.payment_method_id.bank_account_required", readonly=True
```
astirpe: `readonly=True` is the default for related fields and can be removed here

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+        readonly=True,
+        ondelete="restrict",
+        states={"draft": [("readonly", False)]},
+    )
+
+    # I think copying payment mode from invoice to refund by default
+    # is a good idea because the most common way of "paying" a refund is to
+    # deduct it on the payment of the next invoice (and OCA/bank-payment
+    # allows to have negative payment lines since March 2016)
+    @api.model
```
astirpe: This decorator is wrong for `def _reverse_move_vals()`, must be removed.

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+    def _reverse_move_vals(self, default_values, cancel=True):
+        move_vals = super(AccountMove, self)._reverse_move_vals(
```
astirpe: You can remove the parameters of the `super()` here and everywhere else in the code.

account_payment_partner/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "version": "13.0.1.0.0",
+    "category": "Banking addons",
+    "license": "AGPL-3",
+    "summary": "Adds payment mode on partners and invoices",
+    "author": "Akretion, " "Tecnativa, " "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/bank-payment",
+    "depends": ["account_payment_mode"],
+    "data": [
+        "views/res_partner_view.xml",
+        "views/account_invoice_view.xml",
```
astirpe: You could rename this file to `account_move_view,xml`

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+    payment_mode_filter_type_domain = fields.Char(
+        compute="_compute_payment_mode_filter_type_domain"
+    )
+    partner_bank_filter_type_domain = fields.Many2one(
+        comodel_name="res.partner", compute="_compute_partner_bank_filter_type_domain"
+    )
+    payment_mode_id = fields.Many2one(
+        comodel_name="account.payment.mode",
+        string="Payment Mode",
+        compute=_compute_payment_mode,
```
astirpe: If you wrap the method name with quotes, like for example here `compute="_compute_payment_mode"`, you don't need to move the field definition below the compute method. This is valid in general: it's useful to limit the changes in the diff and ease the code review.

account_payment_partner/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+# Copyright 2014 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    # v8 fields : same without the _id suffix
```
astirpe: Is this comment still needed?

account_payment_partner/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+    supplier_payment_mode_id = fields.Many2one(
+        "account.payment.mode",
+        string="Supplier Payment Mode",
```
astirpe: `string="Supplier Payment Mode",` could be removed here, as it's the default value

account_payment_partner/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+        company_dependent=True,
+        domain="[('payment_type', '=', 'outbound')]",
+        help="Select the default payment mode for this supplier.",
+    )
+    customer_payment_mode_id = fields.Many2one(
+        "account.payment.mode",
+        string="Customer Payment Mode",
```
astirpe: string="Customer Payment Mode", could be removed here, as it's the default value

account_payment_partner/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+There is nothing to configure.
```
astirpe: You an remove file `CONFIGURE.rst`, it's useless

astirpe (changes requested):
Only code review, not tested

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @api.depends("partner_id", "type")
+    def _compute_partner_bank_filter_type_domain(self):
+        for move in self:
+            if move.type in ("out_invoice", "in_refund"):
+                move.partner_bank_filter_type_domain = move.bank_partner_id
+            elif move.type in ("in_invoice", "out_refund"):
+                move.partner_bank_filter_type_domain = move.commercial_partner_id
+            else:
+                move.partner_bank_filter_type_domain = False
```
astirpe, line 53: Since you added these two methods, it would be nice if you can write some tests for the code coverage.

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,150 @@
...
+    @api.depends("partner_id", "company_id")
+    def _compute_payment_mode(self):
+        for move in self:
+            if move.partner_id:
+                if move.type == "in_invoice":
+                    move.payment_mode_id = move.with_context(
+                        force_company=move.company_id.id
+                    ).partner_id.supplier_payment_mode_id
+                elif move.type == "out_invoice":
+                    # No bank account assignation is done here as this is only
```
astirpe: If you look at the code of the previous version, this comment is meant to be for the bank assignation, so it should be moved into `def _compute_invoice_partner_bank`

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,150 @@
...
+
+            def get_bank_id():
+                return move.commercial_partner_id.bank_ids.filtered(
+                    lambda b: b.company_id == move.company_id or not b.company_id
+                )[:1]
+
+            if move.partner_id:
+                pay_mode = move.with_context(
+                    force_company=move.company_id.id
+                ).partner_id.supplier_payment_mode_id
```
astirpe: `payment_mode_id` is already passed in the depends(...), so this should be:

`pay_mode = move.payment_mode_id`

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,150 @@
...
+                    else:
+                        move.invoice_partner_bank_id = False
+                elif (
+                    pay_mode
+                    and pay_mode.payment_type == "outbound"
+                    and not pay_mode.payment_method_id.bank_account_required
+                ):
+                    self.invoice_partner_bank_id = False
+                elif not self.payment_mode_id:
+                    self.invoice_partner_bank_id = False
```
astirpe: The last three lines should be fixed, to prevent a singleton error:

```python
                    move.invoice_partner_bank_id = False
                elif not move.payment_mode_id:
                    move.invoice_partner_bank_id = False
```

Moreover, according to the previous version of the module, if `self.type` is not equal to `'in_invoice'`, the bank assignation should not be done. So this entire block of code could be replaced with `self.invoice_partner_bank_id = False`.

astirpe (changes requested):

account_payment_partner/models/account_move.py
dynapps: supplier_payment_mode_id != payment_mode_id

account_payment_partner/models/account_move.py
dynapps: Done!

rven: Done!

account_payment_partner/models/account_move.py
rven: Done!

account_payment_partner/models/account_move.py
rven: Done!

account_payment_partner/models/account_move.py
rven: Comment is moved and a little bit adjusted

ThomasBinsfeld (approved):

astirpe (approved):

pedrobaeza:
There's a report in Odoo with a possible problem of this module:

https://github.com/odoo/odoo/issues/42149

ghost:
V13 Js error on Accounting -> Journal Items

When i try open journal item from **Accounting -> Journal Items** view show .js error.
after activating debug mode with assets, console show error **"Error: Unknown field state in domain**"
in /web/static/src/js/core/domains.js line 68

error reason is ccount_payment_partner\models\account_move_line.py

`payment_mode_id = fields.Many2one(
        comodel_name="account.payment.mode",
        compute="_compute_payment_mode",
        store=True,
        ondelete="restrict",
        index=True,
        states={"draft": [("readonly", False)]},
    )`

i comment states and error fixed.

rven:
@pedrobaeza @bumblebeey Change is done!

SolidGlenn:
Hi, 
Having trouble on the 13.0 branch.
following error:
odoo.tools.convert.ParseError: "Element '<xpath expr="//p[@t-if='o.invoice_payment_term_id']">' cannot be located in parent view

Error context:
View `report_invoice_payment_mode`
[view_id: 2627, xml_id: n/a, model: n/a, parent_id: 901]" while parsing /opt/odoo/oca-addons/bank-payment/account_payment_partner/views/report_invoice.xml:2, near
<odoo>
    </odoo>


Which is strange, since it actually exists in the report?

PR #650, opened 2019-11-08
[MIG] account_payment_sale: Migration to 13.0

rven:
Depends on account_payment_partner https://github.com/OCA/bank-payment/pull/649


account_payment_sale/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2016 Akretion - Alexis de Lattre
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SaleAdvancePaymentInv(models.TransientModel):
+    _inherit = "sale.advance.payment.inv"
+
+    @api.multi
```
astirpe: Remove all the `@api.multi` occurrences

account_payment_sale/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.payment_mode_id:
+            vals["payment_mode_id"] = self.payment_mode_id.id
+            if self.payment_mode_id.bank_account_link == "fixed":
+                vals[
+                    "partner_bank_id"
+                ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
+        return vals
+
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
```
astirpe: Would be better to replace this onchange method with a computed method for field `payment_mode_id`.
So `payment_mode_id` should have: `computed="_compute_payment_mode_id"`, `readonly=False`, `store=True`.

account_payment_sale/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    payment_mode_id = fields.Many2one(
+        "account.payment.mode",
+        string="Payment Mode",
```
astirpe: `string="Payment Mode",` can be removed.

account_payment_sale/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+        :param invoices: {group_key: invoice}
+        :param references: {invoice: order}
+        """
+        payment_vals_by_invoice = {}
+        for invoice in invoices.values():
+            payment_vals_by_invoice[invoice] = {
+                "payment_mode_id": invoice.payment_mode_id.id,
+                "partner_bank_id": invoice.partner_bank_id.id,
+            }
+        res = super()._finalize_invoices(invoices, references)
```
astirpe: Method `def _finalize_invoices()` doesn't exist in V13.

astirpe (changes requested):

account_payment_sale/models/sale_order.py
rven: This method could be removed, because there is no issue with the calling of _onchange_partner_id in the super method.
The method _create_invoices does not call this onchange anymore

ThomasBinsfeld (approved):

atchuthan (approved):
Code review :+1:

okuryan:
@rven are you going to close this PR? I see code review passed, so only few steps to go

MiquelRForgeFlow:
please, pass black again

MiquelRForgeFlow:
FYI Maybe https://github.com/odoo/odoo/commit/5f1f60c1f6655ab210f4c26c0f34c34bb102c03e commit will help in this PR

rven:
I run against the following problem:

> TestSaleOrder.test_sale_to_invoice_payment_mode_2
> AssertionError: account.payment.mode(21,) != account.payment.mode(22,)

This is caused by the computed field on account.move in module account_payment_partner
```python
        for move in self:
            if move.partner_id:
                if move.type == "in_invoice":
                    move.payment_mode_id = move.with_context(
                        force_company=move.company_id.id
                    ).partner_id.supplier_payment_mode_id
                elif move.type == "out_invoice":
                    move.payment_mode_id = move.with_context(
                        force_company=move.company_id.id
                    ).partner_id.customer_payment_mode_id
            else:
                move.payment_mode_id = False
```
This sets the payment mode to the payment mode of the partner, even if the payment_mode is passed in the _prepare_invoice method on account.move in module account_payment_sale
```python
    def _get_payment_mode_vals(self, vals):
        if self.payment_mode_id:
            vals["payment_mode_id"] = self.payment_mode_id.id
            if self.payment_mode_id.bank_account_link == "fixed":
                vals[
                    "invoice_partner_bank_id"
                ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
        return vals

    def _prepare_invoice(self):
        """Copy bank partner from sale order to invoice"""
        vals = super()._prepare_invoice()
        return self._get_payment_mode_vals(vals)
```
The _onchange (v12) methods are converted to computed (v13, store=False) methods.

Anybody any suggestion how to handle with this behaviour?

rven:
@pedrobaeza Any suggestion on the above described problem?

rven:
The problem is caused by the following
```python
    payment_mode_id = fields.Many2one(
        comodel_name="account.payment.mode",
        compute="_compute_payment_mode",
        store=True,
        ondelete="restrict",
        # readonly=False,
        states={"draft": [("readonly", False)]},
    )
```
When I set the _readonly_ property to False, it works fine.
But with the states property it doesn't work.

```python
    @api.model_create_multi
    @api.returns('self', lambda value: value.id)
    def create(self, vals_list):
        ...
                # protect non-readonly computed fields against (re)computation
                if field.compute and not field.readonly:
                    protected.update(self._field_computed.get(field, [field]))
        ...
```
The create method does not take the states property into account, when creating a new record.
The readonly is True in this case...

pedrobaeza:
Maybe you need `force_save="1"` in the view?

rven:
> Maybe you need `force_save="1"` in the view?

It is not in a view, order._create_invoices() triggers the create
```python
    def create_invoice_and_check(
        self, order, expected_payment_mode, expected_partner_bank
    ):
        order.action_confirm()
        order._create_invoices()
        invoice = order.invoice_ids
        self.assertEqual(len(invoice), 1)
        self.assertEqual(invoice.payment_mode_id, expected_payment_mode)
        self.assertEqual(invoice.invoice_partner_bank_id, expected_partner_bank)
```

rven:
I've got a solution for the above problem.
I moved the `states={"draft": [("readonly", False)]}` to the view definition `attrs="{'readonly': [('state', '!=', 'draft')]}"`
This way it is possible to set the `readonly=False` property on the fields.

Is this correct what I'm trying to do?

rven:
@pedrobaeza Should I create another PR for the changes on the `account_payment_partner` module?
```
diff --git a/account_payment_partner/models/account_move.py b/account_payment_partner/models/account_move.py
index 1f606e83..36dbf02e 100644
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -20,7 +20,7 @@ class AccountMove(models.Model):
         compute="_compute_payment_mode",
         store=True,
         ondelete="restrict",
-        states={"draft": [("readonly", False)]},
+        readonly=False,
     )
     bank_account_required = fields.Boolean(
         related="payment_mode_id.payment_method_id.bank_account_required", readonly=True
@@ -29,7 +29,7 @@ class AccountMove(models.Model):
         compute="_compute_invoice_partner_bank",
         store=True,
         ondelete="restrict",
-        states={"draft": [("readonly", False)]},
+        readonly=False,
     )
 
     @api.depends("type")
@@ -134,3 +134,6 @@ class AccountMove(models.Model):
             )
         # Return this as empty recordset
         return self.invoice_partner_bank_id
+
+    def create(self, vals_list):
+        return super().create(vals_list)
diff --git a/account_payment_partner/views/account_move_view.xml b/account_payment_partner/views/account_move_view.xml
index e6f819fb..36cbf356 100644
--- a/account_payment_partner/views/account_move_view.xml
+++ b/account_payment_partner/views/account_move_view.xml
@@ -28,6 +28,7 @@
                 <field name="payment_mode_id"
                        domain="[('payment_type', '=', payment_mode_filter_type_domain), ('company_id', '=', company_id)]"
                        widget="selection"
+                       attrs="{'readonly': [('state', '!=', 'draft')]}"
                        invisible="context.get('default_type') not in  ('out_invoice','out_refund','in_invoice','in_refund')"/>
                 <field name="commercial_partner_id" invisible="1"/>
                 <field name="bank_account_required" invisible="1"/>
@@ -37,7 +38,7 @@
             <field name="invoice_partner_bank_id" position="attributes">
                 <attribute name="domain">[('partner_id', '=', partner_bank_filter_type_domain),
                     '|',('company_id', '=', company_id),('company_id', '=', False)]</attribute>
-                <attribute name="attrs">{'required': [('bank_account_required', '=', True)]}</attribute>
+                <attribute name="attrs">{'required': [('bank_account_required', '=', True)], 'readonly': [('state', '!=', 'draft')]}</attribute>
                 <attribute name="context">{'default_partner_id':commercial_partner_id}</attribute>
             </field>
         </field>
```

rven:
> FYI Maybe [odoo/odoo@5f1f60c](https://github.com/odoo/odoo/commit/5f1f60c1f6655ab210f4c26c0f34c34bb102c03e) commit will help in this PR

@MiquelRForgeFlow The problem is casued by the following
```python
    def _create_invoices(self, grouped=False, final=False):
        """
        Create the invoice associated to the SO.
        :param grouped: if True, invoices are grouped by SO id. If False, invoices are grouped by
                        (partner_invoice_id, currency)
        :param final: if True, refunds will be generated if necessary
        :returns: list of created invoices
        """
        ...
        if invoice_vals['type'] in self.env['account.move'].get_outbound_types():
            invoice_bank_id = self.partner_id.bank_ids[:1]
        else:
            invoice_bank_id = self.company_id.partner_id.bank_ids[:1]

        invoice_vals['invoice_partner_bank_id'] = invoice_bank_id

        # Create invoices.
        moves = self.env['account.move'].with_context(default_type='out_invoice').create(out_invoice_vals_list)
```
The bank account is always recomputed in the _create_invoices even if it has been passed in the values.
Any suggestions how to fix this?

The only option I see is by doing this:
```python
class AccountMove(models.Model):
    _inherit = "account.move"

    @api.model
    def create(self, vals_list):
        if vals_list.get("payment_mode_id"):
            payment_mode = self.env["account.payment.mode"].browse(vals_list["payment_mode_id"])
            if payment_mode.bank_account_link == "fixed":
                vals_list[
                    "invoice_partner_bank_id"
                ] = payment_mode.fixed_journal_id.bank_account_id.id
        return super().create(vals_list)
```

pedrobaeza:
You can put here changes in `account_payment_partner` that are involved in this properly working.

rven:
> You can put here changes in `account_payment_partner` that are involved in this properly working.

Done!

MiquelRForgeFlow:
could you squash last two commits, please?

rven:
> could you squash last two commits, please?

Done!

astirpe (approved):
Code review OK (I didn't test it).

PR #654, opened 2019-11-21
[IMP] if none self.company_id set on invoice use user company

vrenaville:
* dev
If we create an invoice not from the interface, self.company_id is false, so use the user company_id instead

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -20,11 +20,15 @@ class AccountInvoice(models.Model):
 
     @api.onchange('partner_id', 'company_id')
     def _onchange_partner_id(self):
+        if self.company_id:
+            company = self.company_id
+        else:
+            company = self.env.user.company_id
         res = super(AccountInvoice, self)._onchange_partner_id()
         if self.partner_id:
             if self.type == 'in_invoice':
                 pay_mode = self.with_context(
-                    force_company=self.company_id.id
+                    force_company=company.id
                 ).partner_id.supplier_payment_mode_id
                 self.payment_mode_id = pay_mode
                 if (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
CIs red

account_payment_partner/models/account_invoice.py
```diff
@@ -20,11 +20,15 @@ class AccountInvoice(models.Model):
 
     @api.onchange('partner_id', 'company_id')
     def _onchange_partner_id(self):
+        if self.company_id:
+            company_id = self.company_id
```
jcoux: Change variable name (in other places too):
```suggestion
            company = self.company_id
```

jcoux (changes requested):

vrenaville:
@pedrobaeza I have missed to push my last mods, sorry for thaht

account_payment_partner/models/account_invoice.py
vrenaville: @jcoux I keep this variable name to be constant in the function, because of this : https://github.com/OCA/bank-payment/pull/654/files#diff-b7b06a5eb2ba2f2489eb2d427fc5023aR42
But I can change If you want

jcoux: I think it's better to change it.

Tonow-c2c (approved):

account_payment_partner/models/account_invoice.py
sbidoul: I agree with the suggested change: local variables and function arguments ending with `_id` should be reserved to variables containing ids, not record sets.

sbidoul:
LGTM, except the style comment.

jcoux (approved):

PR #658, opened 2019-12-10
[11.0] FIX: Order payments with sepa and non-sepa payment lines are now working.

Nooka10:
I move the sepa and charge_bearer fields from account_payment_order to account_payment_line (and bank_payment_line).
Now, an order that contains sepa and non sepa payment lines will work correctly (which is very useful for l10n_ch).

The payments are now grouped in multiple <PmtInf> inside one same pain001.
They are grouped depending on the 4 same fields as before (requested_date, priority, local_instrument, categ_purpose) plus the boolean is_sepa (true if the line is a sepa payment), the charge_bearer of the line and an index number.

With these changes, we don't need to group sepa payment and non-sepa payment in different  payments orders.

The sepa value is now calculated for every payment line and the charge_bearer value can also be set for each payment line.

### DIFF
--- a/account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
+++ b/account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
@@ -1,0 +1,73 @@
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    """Update database from previous versions, after updating module."""
+
+    if not version:
+        return
+    cr = env.cr
+
+    # update charge_bearer of account_payment_lines
+    plines = env['account.payment.line'].search([])
+    for pline in plines:
+        pline.compute_charge_bearer()
+
+    # update charge_bearer for all bank_payment_line
+    cr.execute(
+        """
+        UPDATE bank_payment_line as l 
+        SET charge_bearer = (
+           SELECT charge_bearer FROM account_payment_order as o 
+           WHERE o.id = l.order_id
+        )
+        """
+    )
+    # correct charge_bearer for sepa bank_payment_line (must be 'SLEV')
+    blines = env['bank.payment.line'].search([])
+    for bline in blines:
+        if bline.sepa:
+            bline.charge_bearer = 'SLEV'
+
+    # update charge_bearer of account_payment_lines
+    # cr.execute(
+    #     """
+    #     UPDATE account_payment_line as l
+    #     SET charge_bearer = (
+    #        SELECT charge_bearer FROM account_payment_order as o
+    #        WHERE o.id = l.order_id
+    #     )
+    #     """
+    # )
+    #
+    # # update charge_bearer of bank_payment_line
+    # cr.execute(
+    #     """
+    #     UPDATE bank_payment_line as l
+    #     SET charge_bearer = (
+    #        SELECT charge_bearer FROM account_payment_order as o
+    #        WHERE o.id = l.order_id
+    #     )
+    #     """
+    # )
+    #
+    # cr.execute(
+    #     """
+    #     SELECT
+    #       res_partner_bank.sanitized_acc_number
+    #     FROM
+    #       public.account_payment_line,
+    #       public.account_payment_order,
+    #       public.account_journal,
+    #       public.res_partner_bank
+    #     WHERE
+    #       account_payment_line.order_id = account_payment_order.id AND
+    #       account_payment_order.journal_id = account_journal.id AND
+    #       account_journal.bank_account_id = res_partner_bank.id;
+    #     """
+    # )
+    # bank_accounts = cr.dictfetchall()
+    #
--- a/account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
+++ b/account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
@@ -3,6 +3,7 @@
 
 from openupgradelib import openupgrade
 
+
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     """Update database from previous versions, after updating module."""
@@ -18,56 +19,12 @@ def migrate(env, version):
 
     # update charge_bearer for all bank_payment_line
     cr.execute(
-        """
-        UPDATE bank_payment_line as l 
-        SET charge_bearer = (
-           SELECT charge_bearer FROM account_payment_order as o 
-           WHERE o.id = l.order_id
-        )
-        """
-    )
+        "UPDATE bank_payment_line as l "
+        "SET charge_bearer = ("
+        "SELECT charge_bearer FROM account_payment_order as o "
+        "WHERE o.id = l.order_id)")
     # correct charge_bearer for sepa bank_payment_line (must be 'SLEV')
     blines = env['bank.payment.line'].search([])
     for bline in blines:
         if bline.sepa:
             bline.charge_bearer = 'SLEV'
-
-    # update charge_bearer of account_payment_lines
-    # cr.execute(
-    #     """
-    #     UPDATE account_payment_line as l
-    #     SET charge_bearer = (
-    #        SELECT charge_bearer FROM account_payment_order as o
-    #        WHERE o.id = l.order_id
-    #     )
-    #     """
-    # )
-    #
-    # # update charge_bearer of bank_payment_line
-    # cr.execute(
-    #     """
-    #     UPDATE bank_payment_line as l
-    #     SET charge_bearer = (
-    #        SELECT charge_bearer FROM account_payment_order as o
-    #        WHERE o.id = l.order_id
-    #     )
-    #     """
-    # )
-    #
-    # cr.execute(
-    #     """
-    #     SELECT
-    #       res_partner_bank.sanitized_acc_number
-    #     FROM
-    #       public.account_payment_line,
-    #       public.account_payment_order,
-    #       public.account_journal,
-    #       public.res_partner_bank
-    #     WHERE
-    #       account_payment_line.order_id = account_payment_order.id AND
-    #       account_payment_order.journal_id = account_journal.id AND
-    #       account_journal.bank_account_id = res_partner_bank.id;
-    #     """
-    # )
-    # bank_accounts = cr.dictfetchall()
-    #
--- a/account_banking_pain_base/models/account_payment_line.py
+++ b/account_banking_pain_base/models/account_payment_line.py
@@ -185,7 +185,7 @@ class AccountPaymentLine(models.Model):
              "by debtor : all transaction charges are to be borne by the "
              "debtor.")
     sepa = fields.Boolean(
-        compute='compute_sepa', readonly=True, string="SEPA Payment Line")
+        compute='_compute_sepa', readonly=True, string="SEPA Payment Line")
 
     @api.multi
     @api.onchange('sepa')
@@ -202,7 +202,7 @@ class AccountPaymentLine(models.Model):
         'order_id.company_partner_bank_id.acc_type',
         'currency_id',
         'partner_bank_id.acc_type')
-    def compute_sepa(self):
+    def _compute_sepa(self):
         for bpline in self:
             bpline._compute_sepa_line()
 
--- a/account_banking_pain_base/models/account_payment_line.py
+++ b/account_banking_pain_base/models/account_payment_line.py
@@ -177,18 +177,29 @@ class AccountPaymentLine(models.Model):
         track_visibility='onchange',
         help="Following service level : transaction charges are to be "
              "applied following the rules agreed in the service level "
-             "and/or scheme (SEPA Core messages must use this). Shared : "
-             "transaction charges on the debtor side are to be borne by "
-             "the debtor, transaction charges on the creditor side are to "
-             "be borne by the creditor. Borne by creditor : all "
-             "transaction charges are to be borne by the creditor. Borne "
+             "and/or scheme (SEPA Core messages must use this). "
+             "\nShared : transaction charges on the debtor side are to be "
+             "borne by the debtor, transaction charges on the creditor side "
+             "are to be borne by the creditor. \n Borne by creditor : all "
+             "transaction charges are to be borne by the creditor. \nBorne "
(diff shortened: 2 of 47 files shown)


davidwul:
This PR would remove the need for 
- [this one on Swiss localisation](https://github.com/OCA/l10n-switzerland/pull/456) (pending)
- [this one on bank-payement](https://github.com/OCA/bank-payment/pull/587) (closed)

pedrobaeza:
I  don't like to move at payment line level `charge_bearer` and force to set it each time. `sepa` field, being computed, is more suitable to be at line level, but this will introduce a side effect: if a payment order has several lines and one of them has a bad IBAN account (by user mistake), it will pass unaware by the payment order creator.

davidwul:
There is a default value for Charge_bearer (SLEV) so there is no need to set it for each line. 
This allows to set the OUR or BEN for a specific line instead of having to do a special payment order when you need to do that.
Regarding the side effect you mention, I don't get what this PR introduce that change the visibility of user mistake, can you elaborate?

pedrobaeza:
> There is a default value for Charge_bearer (SLEV) so there is no need to set it for each line.

Yes, there is when you don't want to use that one.

> Regarding the side effect you mention, I don't get what this PR introduce that change the visibility of user mistake, can you elaborate?

Previously, the file was not exportable if that happened.

davidwul:
Regarding Charge_Bearer, for us the usual process is to not change this SLEV. It's the exception to set it. Is it different else where? If really the case, we could add a selection field at the order level to set all line with this default.

Regarding the side effect, I now understand, What about adding the account type at the payement line level to make it more visual if not iban? Do you have other suggestion?

To be able to group sepa and non sepa payement in the same order is a real plus for CH. Unfortunately there is no way to do this in our localisation.

pedrobaeza:
I would introduce at payment method level a flag for allowing this mixed things.

davidwul:
We did some test on a non-CH environment.
If there is a mistake in the IBAN, it is indeed possible de generate the pain file, it will just not be a SEPA order. We could simulate that by adding the "SEPA order"field back on the payement order and set it true when all payment line are SEPA. This would have the same behavior as now.

pedrobaeza:
OK, I think it can be enough with that and the bearer at both lines and header (it can be a computed with inverse). Any way, this change is huge, so I need the opinion from @sbidoul and @alexis-via on this matter.

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -153,7 +153,6 @@ def check_eur_currency_sct(self):
         self.assertEqual(agrolait_pay_line1.communication, 'F1341')
         self.payment_order.draft2open()
         self.assertEqual(self.payment_order.state, 'open')
-        self.assertEqual(self.payment_order.sepa, True)
```
sbidoul, line 156: don't we want to replace this assertion by a check on the lines?

account_banking_pain_base/models/account_payment_line.py
```diff
@@ -166,3 +166,40 @@ class AccountPaymentLine(models.Model):
...
+             "by debtor : all transaction charges are to be borne by the "
+             "debtor.")
+    sepa = fields.Boolean(
+        compute='compute_sepa', readonly=True, string="SEPA Payment Line")
+
+    @api.multi
+    @api.depends(
+        'currency_id',
+        'partner_bank_id.acc_type')
+    def compute_sepa(self):
```
sbidoul:
```suggestion
    def _compute_sepa(self):
```

account_banking_pain_base/models/account_payment_line.py
```diff
@@ -166,3 +166,40 @@ class AccountPaymentLine(models.Model):
...
+             "applied following the rules agreed in the service level "
+             "and/or scheme (SEPA Core messages must use this). Shared : "
+             "transaction charges on the debtor side are to be borne by "
+             "the debtor, transaction charges on the creditor side are to "
+             "be borne by the creditor. Borne by creditor : all "
+             "transaction charges are to be borne by the creditor. Borne "
+             "by debtor : all transaction charges are to be borne by the "
+             "debtor.")
+    sepa = fields.Boolean(
+        compute='compute_sepa', readonly=True, string="SEPA Payment Line")
```
sbidoul:
```suggestion
        compute='_compute_sepa', readonly=True, string="SEPA Payment Line")
```

account_banking_sepa_credit_transfer/models/account_payment_order.py
```diff
@@ -85,40 +84,46 @@ def generate_payment_file(self):
             priority = line.priority
             local_instrument = line.local_instrument
             categ_purpose = line.category_purpose
+            line.compute_sepa()
```
sbidoul: this is suspicious: computed fields should update themselves automatically if the dependencies are correct

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -230,7 +228,6 @@ def test_usd_currency_sct(self):
         self.assertEqual(asus_pay_line1.communication, 'Inv9032')
         self.payment_order.draft2open()
         self.assertEqual(self.payment_order.state, 'open')
-        self.assertEqual(self.payment_order.sepa, False)
```
sbidoul, line 233: You'll probably want to test the equivalent flag on the line, instead of completely removing the assertion.

account_banking_sepa_direct_debit/models/account_payment_order.py
```diff
@@ -69,6 +68,8 @@ def generate_payment_file(self):
             transactions_count_a += 1
             priority = line.priority
             categ_purpose = line.category_purpose
+            line.compute_sepa()
```
sbidoul: This is suspicious: you should normally not need to call compute methods manually.

account_payment_order/models/account_payment_order.py
```diff
@@ -25,7 +25,7 @@ def domain_journal_id(self):
             return [('id', 'in', jrl_ids)]
 
     name = fields.Char(
-        string='Number', readonly=True, copy=False)  # v8 field : name
+        string='Number', readonly=False, copy=False)  # v8 field : name
```
sbidoul: This change seems unrelated to the scope of the PR.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -230,7 +194,7 @@ def generate_start_payment_info_block(
             instruction_priority = etree.SubElement(
                 payment_type_info, 'InstrPrty')
             instruction_priority.text = priority
-        if self.sepa:
+        if eval_ctx.get('sepa'):
```
sbidoul, line 264: Is it necessary to pass this flag via eval_ctx? Is it not better to pass it as a function argument?

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -50,31 +31,14 @@ class AccountPaymentOrder(models.Model):
...
-                if pline.partner_bank_id.acc_type != 'iban':
-                    sepa = False
-                    break
-            sepa = order.compute_sepa_final_hook(sepa)
-            self.sepa = sepa
-
-    @api.multi
-    def compute_sepa_final_hook(self, sepa):
-        self.ensure_one()
-        return sepa
```
sbidoul, line 77: I personally don't use this hook but I suppose it was there for a reason. Not sure how to preserve it though.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -50,31 +31,14 @@ class AccountPaymentOrder(models.Model):
...
+    @api.model
+    def _prepare_bank_payment_line(self, paylines):
+        result = super()._prepare_bank_payment_line(paylines)
+        result.update({
+            'sepa': paylines[0].sepa,
+            'charge_bearer': paylines[0].charge_bearer
+        })
```
sbidoul, line 80: A comment that explain why it is safe to use `paylines[0]` would be helpful (do we have guarantee that all payline have the same value for these two fields?).

sbidoul: BTW, why is there a need to write the `sepa` field when it is actually a computed field?

sbidoul (commented):
I made a first superficial review. Some comments below.

sbidoul:
I suppose this requires a migration script?

account_banking_pain_base/models/account_payment_order.py
Nooka10: I think it's not needed anymore because the way the sepa field is computed has changed.

It was used in l10n_ch (https://github.com/OCA/l10n-switzerland/blob/0e006e5bcbff4c64bfc5fee8dd28123d0b139d31/l10n_ch_pain_base/models/account_payment_order.py#L13) to deactivate the sepa, but as I change the way the sepa field is now computed.

I don't know if it's used somewhere else. Do you prefer if I let this hook just in case?

account_payment_order/models/account_payment_order.py
Nooka10: Oups, it's a mistake. That was only for test purposes. I Changed it back in my last commit https://github.com/OCA/bank-payment/commit/2e1b8baa27bf83493f14237de0248ccda14ca931#diff-ef3721e9aed5856752bc942dd6e96dc0R28

account_banking_pain_base/models/account_payment_order.py
Nooka10: Indeed, sepa is not needed anymore. I corrected it in my last commit (https://github.com/OCA/bank-payment/pull/658/commits/1b5fd8797020563bc7b9c35a5539333631874a28).

Paylines[0] is just fine because paylines are now grouped also by sepa and charge_bearer values so all paylines have the same charge_bearer (https://github.com/OCA/bank-payment/pull/658/commits/2c14f5d7b7252a670b78f78c25b3c4f850e39405#diff-572f2cb9ee06aba671bb3532e23ca120R43).

account_banking_pain_base/models/account_payment_order.py
Nooka10: I did this to avoid changing the signature of the function, which could be a problem for modules using bank-payment like l10n modules...

But If you think I can do it, it'll be probably better to do it with an argument, yes.

Nooka10:
> I suppose this requires a migration script?

I add it in my last commit (https://github.com/OCA/bank-payment/pull/658/commits/408c595bb0c43e18b51ca90da48a63fa2e76df12)

account_banking_sepa_credit_transfer/tests/test_sct.py
Nooka10: Right. I added some checks in my commit https://github.com/OCA/bank-payment/pull/658/commits/dcd653f0bb578aa573ed6e6b5f7c924e2b3c4cf8

account_banking_sepa_credit_transfer/tests/test_sct.py
Nooka10: Right. I added some checks in my commit https://github.com/OCA/bank-payment/pull/658/commits/dcd653f0bb578aa573ed6e6b5f7c924e2b3c4cf8

davidwul:
Thanks @sbidoul for the high level review. 
Is there other remarks on the process or on the code written?
Is there anything doable to move forward this PR?

ecino (approved):
This one would be a good change for allowing mixing sepa and non-sepa payments in a same order, which is something possible in Switzerland. It eases our processes by allowing handling them all together regardless their type.

ecino:
@sbidoul Is it good for you now?

account_banking_pain_base/models/account_payment_order.py
sbidoul: Well, the breaking change is there anyway (one needs to pass sepa in the context for it to work), so adding an additional keyword argument is going to be more readable.

account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,30 @@
...
+    """Update database from previous versions, after updating module."""
+
+    if not version:
+        return
+    cr = env.cr
+
+    # update charge_bearer of account_payment_lines
+    plines = env['account.payment.line'].search([])
+    for pline in plines:
+        pline.compute_charge_bearer()
```
sbidoul, line 18: This should not be necessary, since it is a new field, Odoo will compute it automatically.

account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,30 @@
...
+    cr.execute(
+        "UPDATE bank_payment_line as l "
+        "SET charge_bearer = ("
+        "SELECT charge_bearer FROM account_payment_order as o "
+        "WHERE o.id = l.order_id)")
+    # correct charge_bearer for sepa bank_payment_line (must be 'SLEV')
+    blines = env['bank.payment.line'].search([])
+    for bline in blines:
+        if bline.sepa:
+            bline.charge_bearer = 'SLEV'
```
sbidoul, line 30: I would do this in SQL too: `UPDATE bank_payment_line SET charge_bearer = 'SLEV' where sepa`?
That said, if there was incorrect data, it's maybe not the role of this PR to fix it?

sbidoul (commented):
Mixing sepa and non-sepa payment is a good thing to have indeed.

I made another review round. Technically this looks good but I've not had a chance to test in real life, and I really don't know if there are restrictions to such mixing (standard-wise of local restrictions).

@alexis-via do you have experience or information to share in this area?

account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
ecino: In fact the field is not a computed field: it has 'SLEV' by default which may not be right for every lines.

account_banking_pain_base/migrations/11.0.2.0.0/post-migrate.py
ecino: `sepa` is a computed field so it's not present in database.

ecino: In fact charge_bearer is a new field in bank.payment.line so this is needed.

ecino:
@sbidoul I reviewed the migration script and in my opinion it is correct and doesn't need any change (see my individual comments). I made the change you suggested about the method signature and made another pull request here : #689

PR #663, opened 2019-12-31
[12.0][FIX]Structured Communication

luc-demeyer:
Fix traceback when adding invoice with structured communication.


luc-demeyer:
This PR fixes the traceback but we could also solve this in a different way by remove the 'structured' communication type from the invoice. Structured communication types should imho being added by addon modules (e.g. 'ISO' type by pain module).

ThomasBinsfeld (approved):

PR #666, opened 2020-01-08
[FIX] Allow default values for payment mode in multi-company env

CasVissers-360ERP:
If using ir.default for payment mode in a multi-company environment the payment mode of a sister company could be loaded if superuser (uid 1) was active in this sister company. Resulting in the error message:
_("The company of the invoice %s does not match with that of the payment mode") % rec.name)

This PR fixes this issue.

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -18,6 +18,9 @@ class AccountInvoice(models.Model):
         readonly=True)
     partner_bank_id = fields.Many2one(ondelete='restrict')
 
+    def _get_refund_common_fields(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please indicate steps to reproduce the problem. I don't think populating `company_id` in the refund fields is the solution to your problem.

CasVissers-360ERP:
@pedrobaeza I would like to create a runbot video, but OCA runbot seems not accessible?

pedrobaeza:
You can use the "Try me on runbot" button. Any way, I still prefer written steps than a hard to follow video. And as said, a technical explanation why this solves the problem is also welcome (included as comment/docstring of the method as well for future references).

PR #668, opened 2020-01-14
[IMP] account_payment_partner: move search of payment mode in separat…

i-vyshnevska:
refactor methods a little bit to allow flexibility for customization

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -22,11 +22,9 @@ class AccountInvoice(models.Model):
     def _onchange_partner_id(self):
         res = super(AccountInvoice, self)._onchange_partner_id()
         if self.partner_id:
+            self.payment_mode_id = self._get_payment_mode()
+            pay_mode = self.payment_mode_id
             if self.type == 'in_invoice':
-                pay_mode = self.with_context(
-                    force_company=self.company_id.id
-                ).partner_id.supplier_payment_mode_id
-                self.payment_mode_id = pay_mode
                 if (
                     pay_mode and
                     pay_mode.payment_type == 'outbound' and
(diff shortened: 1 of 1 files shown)


account_payment_partner/models/account_invoice.py
```diff
@@ -22,11 +22,9 @@ class AccountInvoice(models.Model):
     def _onchange_partner_id(self):
         res = super(AccountInvoice, self)._onchange_partner_id()
         if self.partner_id:
+            self.payment_mode_id = self._get_payment_mode()
```
pedrobaeza: Shouldn't this go out also of `if self.partner_id`?

account_payment_partner/models/account_invoice.py
```diff
@@ -40,19 +38,26 @@ def _onchange_partner_id(self):
...
-                ).partner_id.customer_payment_mode_id
         else:
             self.payment_mode_id = False
             if self.type == 'in_invoice':
                 self.partner_bank_id = False
         return res
 
+    def _get_payment_mode(self):
+        # find appropriate payment mode
+        # in cale of out-invoice bank account assignation is done here as this is only
```
TDu:
```suggestion
        # in case of out-invoice bank account assignation is done here as this is only
```

account_payment_partner/models/account_invoice.py
TDu: Looks like `self.partner_id` is set to False in the else case. So I would think not.

TDu (approved):

account_payment_partner/models/account_invoice.py
i-vyshnevska: indeed it can be outside, and simpler to reuse
updated

StefanRijnhart:
Looks good! Restarting runbot. Would you consider including https://github.com/OCA/bank-payment/pull/711/files?

PR #674, opened 2020-01-29
[13.0][MIG] account_payment_purchase

MiquelRForgeFlow:
Standard migration.

- [x] Includes a cherry-pick of https://github.com/OCA/bank-payment/pull/603.


astirpe:
Could you apply the change proposed here?
https://github.com/OCA/bank-payment/pull/603

We have many customers who don't use stock.

MiquelRForgeFlow:
> Could you apply the change proposed here?

Ok, when your change is approved, then I will add them or forward port them.

astirpe:
Well, in #603 the codecov/patch is 0% of diff hit and it's marked as red, but it doesn't decrease the coverage, so could be considered green.
And even if it will be not merged before this one, still it's a requirement that our customers need.

account_payment_purchase/models/account_invoice.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.onchange("purchase_vendor_bill_id", "purchase_id")
+    def _onchange_purchase_auto_complete(self):
+        new_mode = self.purchase_id.payment_mode_id.id or False
+        new_bank = self.purchase_id.supplier_partner_bank_id.id or False
+        super(AccountMove, self)._onchange_purchase_auto_complete()
+        res = {}
```
astirpe:
```python
res = super()._onchange_purchase_auto_complete() or {}
```

and at the end the method should always `return res`.

astirpe (changes requested):

misern2:
Could you please fix conflicts ?

MiquelRForgeFlow:
@misern2 done :)

misern2:
@MiquelRForgeFlow  LGTM. Functional test.  There's some issue on travis. (precommit)

MiquelRForgeFlow:
> There's some issue on travis. (precommit)

Fixed.

misern2 (approved):

HviorForgeFlow (approved):

PR #679, opened 2020-02-19
[IMP] account_payment_order_return: import journal items without invoice

oihane:
When you try to add to a payment order a journal item from a expense it does not show them if they don't have an invoice. This solves the problem.

It should be backported to v11

### DIFF
--- a/account_payment_order_return/wizards/account_payment_line_create.py
+++ b/account_payment_order_return/wizards/account_payment_line_create.py
@@ -2,6 +2,7 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo import api, fields, models
+from odoo.models import expression
 
 
 class AccountPaymentLineCreate(models.TransientModel):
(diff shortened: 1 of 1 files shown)


cubells (changes requested):
It's good to me.

But I'd do two changes more:

1) update version of the module
2) change information here: https://github.com/OCA/bank-payment/blob/12.0/account_payment_order_return/readme/DESCRIPTION.rst because not only invoices can be added now.

pedrobaeza:
Version is now changed by the bot, so no need to do it manually.

cubells:
> Version is now changed by the bot, so no need to do it manually.

Perfect!

pedrobaeza:
See https://odoo-community.org/blog/the-oca-blog-1/post/the-oca-github-bot-can-now-merge-pull-requests-92

cubells:
Better action is this: "Give some love to contributors."  :smile: 

:heart: :heart: :heart:

oihane:
@cubells I'm not 100% sure but, as this module is an extension of the `account_payment_order` in this case the description is correct, as it only adds the possibillity of including returning invoices. It's a fix, because it does not import expense's journal items correctly if this module is installed.

> 2. change information here: https://github.com/OCA/bank-payment/blob/12.0/account_payment_order_return/readme/DESCRIPTION.rst because not only invoices can be added now.

pedrobaeza:
Can you please add a test that exercises the problem for avoiding regressions?

pedrobaeza:
And rebase...

PR #680, opened 2020-02-20
[12.0][IMP] account_payment_order: change 'bill reference' column value in reports

ernestotejeda:
Change 'bill reference' column value in reports for supplier invoices

Cc @Tecnativa TT22069


carlosdauden (approved):

pedrobaeza (approved):

rafaelbn (approved):
Reviewed :+1: 

Change in the PDF report `number` by `reference` which is the real number of the real invoices

PR #682, opened 2020-02-27
[FIX] 12.0 refactor account_payment_order tests

PieterPaulussen:
@pedrobaeza Since you were the last person to update these tests, I would appreciate a small review.
In a multicompany setup, the search results on `account.journal` are not accurate enough because the record rules do not apply to the superuser.
I also did some cleaning in the setupclass method since there were duplicate class variables set there.

### DIFF
--- a/account_payment_order/README.rst
+++ b/account_payment_order/README.rst
@@ -99,6 +99,7 @@ Contributors
 * Jose María Alzaga <jose.alzaga@aselcis.com>
 * Meyomesse Gilles <meyomesse.gilles@gmail.com>
 * Carlos Dauden
+* Pieter Paulussen <pieterpaulussen@code-source.be>
 
 Maintainers
 ~~~~~~~~~~~
--- a/account_payment_order/demo/payment_demo.xml
+++ b/account_payment_order/demo/payment_demo.xml
@@ -20,12 +20,12 @@
 
 <record id="account_payment_mode.payment_mode_outbound_ct1" model="account.payment.mode">
     <!-- Credit Transfer to Suppliers -->
-    <field name="default_journal_ids" search="[('type', 'in', ('purchase', 'purchase_refund'))]"/>
+    <field name="default_journal_ids" search="[('type', 'in', ('purchase', 'purchase_refund')), ('company_id', '=', ref('base.main_company'))]"/>
 </record>
 
 <record id="account_payment_mode.payment_mode_inbound_dd1" model="account.payment.mode">
     <!-- Direct Debit of customers -->
-    <field name="default_journal_ids" search="[('type', 'in', ('sale', 'sale_refund'))]"/>
+    <field name="default_journal_ids" search="[('type', 'in', ('sale', 'sale_refund')), ('company_id', '=', ref('base.main_company'))]"/>
 </record>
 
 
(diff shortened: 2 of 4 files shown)


pedrobaeza:
It's very difficult to understand the problem if you mix stylistic changes with the real ones. Please restore again the `self = cls` line and let `self`, although I know that's not the best idiomatic, but it was done the same for reducing the diff (TransactionCase was switched to SavepointCase for increasing performance). If you want still to apply that change, at least do it in a separate commit.

When done, I will come back and check.

PieterPaulussen:
@pedrobaeza Thanks for the feedback, I split the changes into 2 separate commits as you suggested

pedrobaeza (approved):
Perfect, thanks

pedrobaeza (changes requested):
Oh, not so perfect. Please don't undo the split in 2 classes: Base and the other. You are breaking the inheritability of that class for using in other modules tests.

pedrobaeza:
Example of something that you have broken:

https://github.com/OCA/account-reconcile/blob/f93f6d6415e60959c60565c190f5d7911a8e8cde/account_reconcile_payment_order/tests/test_account_reconcile_payment_order.py#L5

and not only that, this is done on purpose for not repeating the tests in the inherited test.

PieterPaulussen:
I was unaware of the usage of that base case. Stupid of me not to check.

pedrobaeza (approved):
It's a good habit any way to split your tests in 2 classes if others can reuse your setup: one base with that setup, and another for the tests themselves.

PieterPaulussen:
@pedrobaeza Found another little snag during the tests [here](https://github.com/OCA/bank-payment/pull/682/files#diff-de4893858393574c4b1e0d1038cd93afR24) and [here](https://github.com/OCA/bank-payment/pull/682/files#diff-de4893858393574c4b1e0d1038cd93afR30-R34) . The installation of the demo data adds all existing journals and ignores the addition of the domain defined on the field definition of `default_journal_ids`. There are 2 possible solutions of course. Either specify the journals with a domain on the main company in the demo data, or just update it in the tests. I chose the second because it seemed trivial.

pedrobaeza:
I don't get the problem, but I see problematic to remove the company from the demo data.

PieterPaulussen:
I believe these domains [here](https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/demo/payment_demo.xml#L23) and [here](https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/demo/payment_demo.xml#L28) don't take into account the default domain defined on the field [here](https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/models/account_payment_mode.py#L33)
Given a standard installation where you just have 1 company and install this module, this will always work. But when multiple companies are already setup, all journals including those from other companies are included. This leads to the problem where moves are selected outside of the current user company for which payment orders are generated for. Does that explain the problem a bit more?

pedrobaeza:
Then I think the patch is to include in the search also the main_company as part of that domain.

PieterPaulussen:
@pedrobaeza Updated the domain as well. Feel free to review my changes again. Thanks!

pedrobaeza (approved):
Thanks for the work!

carlosdauden (approved):

PR #692, opened 2020-04-02
[FIX] account_payment_partner: payment_mode_id default not set for Refund

atchuthan:
When we try to create a refund for an invoice, we get an error that payment_mode_id default is not set


atchuthan:
@rven @astirpe @ThomasBinsfeld

astirpe (approved):
LGTM

ThomasBinsfeld (approved):

PR #695, opened 2020-04-10
[13.0][FIX] account_payment_order: multicompany rule

astirpe:
As per migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0


pedrobaeza (approved):

pedrobaeza:
Let's fast-track it as not having them will affect more people:

/ocabot merge

PR #697, opened 2020-04-20
[12.0][FIX] make tests compatible with runbot

hbrunn:
the current tests rely on a chart of accounts existing when they run, which is generally not the case. On travis, this works because there we do `odoo -i account` first and only then `odoo -i ourmodules --test-enable ...`. The `account` module has all its tests post install because then whatever module is selected by the [hook](https://github.com/OCA/OCB/blob/12.0/addons/account/__init__.py#L13) has been installed and we have a COA. So I propose to do the same here, this way those tests also work on a plain runbot, or on odoo.sh.

The implied group is necessary so that https://github.com/OCA/OCB/blob/12.0/addons/account/tests/account_test_users.py#L15 can create a bank. I think this makes sense anyways, account managers should be allowed to do anything concerning accounting by default.

We need the last commit in `fields_view_get` to have a correct `fields` dictionary, as otherwise tests fail in https://github.com/OCA/OCB/blob/12.0/odoo/models.py#L5343 where core expects fields used in a view to show up in the `fields` dict.

Without the above, if you run `odoo -d empty_db -i account_payment_partner --test-enable --stop-after-init`, it actually deadlocks because the cursor from the [exception](https://github.com/OCA/bank-payment/blob/12.0/account_payment_partner/tests/test_account_payment_partner.py#L34) blocks the cursor created for the next test. This should probably be rewritten to raise `unittest.SkipTest` early in that case, or inherit from [AccountingTestCase](https://github.com/OCA/OCB/blob/12.0/addons/account/tests/account_test_classes.py#L9). The latter I started, but the diff would become huge and what I did here we'd need anyways, so I start with that.

### DIFF
--- a/account_payment_order/demo/payment_demo.xml
+++ b/account_payment_order/demo/payment_demo.xml
@@ -28,5 +28,9 @@
     <field name="default_journal_ids" search="[('type', 'in', ('sale', 'sale_refund')), ('company_id', '=', ref('base.main_company'))]"/>
 </record>
 
+<!-- without this, account tests fail /-->
+<record id="account.group_account_manager" model="res.groups">
+    <field name="implied_ids" eval="[(4, ref('group_account_payment'))]" />
+</record>
 
 </odoo>
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -132,5 +132,8 @@ class AccountMoveLine(models.Model):
                 doc.remove(elem)
             for elem in doc.xpath("//field[@name='credit']"):
                 doc.remove(elem)
-            result['arch'] = etree.tostring(doc)
+            arch, fields = self.env['ir.ui.view'].postprocess_and_fields(
+                self._name, doc, view_id,
+            )
+            result.update(arch=arch, fields=fields)
         return result
--- a/account_payment_order/tests/test_account_payment.py
+++ b/account_payment_order/tests/test_account_payment.py
@@ -1,6 +1,7 @@
 # Copyright 2019 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import unittest
 from odoo.tests.common import SavepointCase
 
 
@@ -25,6 +26,8 @@ class TestAccountPayment(SavepointCase):
         # Journals
         cls.bank_journal = cls.account_journal_model.search(
             [('type', '=', 'bank')], limit=1)
+        if not cls.bank_journal:
+            raise unittest.SkipTest('No journals defined')
         cls.bank_journal.inbound_payment_method_ids = [
             (6, 0, [cls.inbound_payment_method_01.id,
                     cls.inbound_payment_method_02.id])]
--- a/account_payment_order/tests/test_payment_mode.py
+++ b/account_payment_order/tests/test_payment_mode.py
@@ -1,6 +1,7 @@
 # © 2017 Creu Blanca
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import unittest
 from odoo.tests.common import TransactionCase
 from odoo.exceptions import ValidationError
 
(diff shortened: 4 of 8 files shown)


NL66278 (approved):
:+1: LGTM

pedrobaeza:
It's better to put each patch in one PR, as there are some of them that can be debatable.

Starting with, I'm afraid the post-install patch is not correct, but a side-effect of your way to do the CI (or we can call it also a limitation of current testing framework), but at the end, you have to follow the same way that Odoo/OCA Travis is testing: first install dependencies that are not going to be tested, and the install and test at the same time what you want to test. If not, post-install is causing other undesirable side effects to the rest that are doing this in the proper way: some tests will fail due to the new possible injected behaviors of other modules that are being installed in your DB.

hbrunn:
I was part of writing the travis scripts, and it was never a choice doing it like that, but a result of some issues with the versions back then.

I am aware of the disadvantages of post install tests, but given any test in https://github.com/OCA/OCB/blob/12.0/addons/account/tests/ does it like that, I don't see why the OCA should go and declare that wrong. I can't find a convention about that btw.

pedrobaeza:
The problem comes when you do integration tests and they include these tests. It's not easy to go this way, but some blockings in other PRs from @lmignon and others have forced us to follow that way if we want some things to be merged.

lmignon:
My main problem is we write code in addons in case other addons they don't depend on are installed because the tests are marked as pot-install. 
I maintain that it is not a good practice to run all tests in "post_install" for OCA modules. I can understand that in some rare cases this may be justified but it should not be the rule.

hbrunn:
so I remove the post install flags, replace the exceptions that mess up everything with SkipTest and you're happy?

The side effect will be that those tests never run on non-travis CIs (which I maintain is the standard: runbot), but I can live with that.

pedrobaeza:
Why don't you split this PR for each of the commits so each of them can go its own way?

hbrunn:
because this is the make-it-work-on-runbot-PR, not a couple of unrelated commits

pedrobaeza:
I prefer some unrelated commits to be merged soon than a big one blocked by one of the questions, but as you wish.

hbrunn:
@lmignon @pedrobaeza I removed the post install flags and added `unittest.SkipTest()` where appropriate. With this we all can live well I think.

account_payment_order/security/payment_security.xml
```diff
@@ -14,6 +14,10 @@
 <data noupdate="1">
 
 
+<record id="account.group_account_manager" model="res.groups">
```
pedrobaeza: This one is not correct IMO and I have cases where account managers shouldn't be able to perform orders. Put that admin user belongs to this, but not the implied group.

hbrunn: that's why it's noupdate and you configure your case the way you need it. As a default, I think this is very sensible.

pedrobaeza: The same applies to you, so please leave it as it was.

hbrunn: as written above, this is not for my private amusement, but has a technical reason:

> The implied group is necessary so that https://github.com/OCA/OCB/blob/12.0/addons/account/tests/account_test_users.py#L15 can create a bank. I think this makes sense anyways, account managers should be allowed to do anything concerning accounting by default.

I moved this to demo data then. But I still believe it's very undesirable for the general case that installing this module takes away permissions from account managers.

pedrobaeza: For the technical reason, for me it's as simple as adding the group to the user in the setUp of your tests.

For the functional ones, I let other to give their opinion, but for me it's OK as it is.

NL66278:
@pedrobaeza Can you live with this as it is now?

pedrobaeza:
OK

NL66278:
@pedrobaeza Would you approve it and merge it then? I have no authority in this repo.

pedrobaeza:
I'm not comfortable approving it as I don't check it deeply, but let's merge it:

/ocabot merge patch

NL66278:
@pedrobaeza Thank you!

PR #671, opened 2020-01-21
[MIG] account_banking_sepa_credit_transfer: Migration to 13.0

axelpriem:
Depends on account_banking_pain_base #670 
(and indirectly on account_payment_order #651, account_payment_partner #649 , account_payment_mode)

PR #699, opened 2020-04-22
[13.0][MIG] account_banking_sepa_credit_transfer: Migration to 13.0

ValentinVinagre:
This PR supersedes https://github.com/OCA/bank-payment/pull/671

- [x] Depends https://github.com/OCA/bank-payment/pull/700


sergio-teruel:
@ValentinVinagre Please can you check travis??

ValentinVinagre:
Hi @sergio-teruel I'm waiting this fix https://github.com/OCA/bank-payment/pull/700 to be approved.

sergio-teruel:
Ok. Thanks.. @carlosdauden Can you review https://github.com/OCA/bank-payment/pull/700?

ValentinVinagre:
Hi @sergio-teruel @HaraldPanten full green :)

HaraldPanten:
I think there's a kind of "bug" with runbots these days. I've seen it in other PR's. Runbot is working but it fails. And when you try to check all logs a 404 error appears. 

Look at https://github.com/OCA/sale-workflow/pull/1154

@pedrobaeza do you know if we are missing something? Any clue of how could we solve this? Thanks in advance.

ValentinVinagre:
Hi @sergio-teruel, @pedrobaeza 
Do you know any solution for the runbot not to get error 404?
Thanks,

pedrobaeza:
I can't say. Other times, rebuilding several times has made it.

kurogeek:
I just checked on the runbot. I tried to install the "account_banking_sepa_credit_transfer" on runbot system. There was an error about missing module "account_banking_pain_base". I think you might need to include the module in your repo.

DeepinScreenshot_select-area_20200609100947

ValentinVinagre:
> I just checked on the runbot. I tried to install the "account_banking_sepa_credit_transfer" on runbot system. There was an error about missing module "account_banking_pain_base". I think you might need to include the module in your repo.
> 
> DeepinScreenshot_select-area_20200609100947

Done, thanks :)

ValentinVinagre:
@HaraldPanten could you review please?

HaraldPanten (commented):
It's functionally working fine. But I'd like to comment:

As invoices are now account.moves, user must set a "reference" field in invoices, or "communication" field in bank.payment.lines wouldn't be set. If communication fields are not set, an error is occurring as a mandatory field hasn't been set. 

In previous versions, communication field was inherited from invoice numbers if I'm right. But now, it inherits the "reference" field.

Do you think that any info should be shown in invoices when user selects a payment mode related to a "sepa_credit_transfer" payment method? I was thinking about that, as I don't think we should set "reference" field as mandatory.

See attached images:

Captura de pantalla 2020-06-09 a las 12 55 19

This is my suggestion:

Captura de pantalla 2020-06-09 a las 12 44 47


What do you think @ValentinVinagre @pedrobaeza @sergio-teruel ?

THX

ValentinVinagre:
Hi @pedrobaeza , @HaraldPanten ,@sergio-teruel 
I think that the problem is here, for some reason the code detects the 'ref' field isn't empty, and doesn't choose the "name" field.

https://github.com/OCA/bank-payment/blob/080bad77d6eedb031004a993743ebf5356821102/account_payment_order/models/account_move_line.py#L52-L64

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -0,0 +1,366 @@
...
+        cls.partner_asus.company_id = cls.main_company.id
+        cls.partner_c2c.company_id = cls.main_company.id
+        cls.env.user.write(
+            {
+                "company_ids": [(6, 0, cls.main_company.ids)],
+                "company_id": cls.main_company.id,
+            }
+        )
+        cls.env.ref(
+            "l10n_generic_coa.configurable_chart_template"
```
carlosdauden: Can we use this without having l10n_generic_coa dependency in the manifest?

ValentinVinagre: I think it is correct as it is, if it is added in the manifest the "l10n_generic_coa" will always be installed but it is only used in the test.

pedrobaeza: No, it's not, as this is pre-configuring for the first company that CoA there, provoking problems. Why previous instruction is not working?

ValentinVinagre: I don't understand, should I change this?

pedrobaeza: What is the problem with the code in v12?

ValentinVinagre: IMO the code is correct, this code fragment is the same in v.12. If it isn't correct, please explain more the error, because I don't understand the previous comments.

pedrobaeza: You're right. It was like that in v12, but it's not the best approach, as the module may be missing and this will fail. Let's keep it this way until further analysis.

HaraldPanten:
> It's functionally working fine. But I'd like to comment:
> 
> As invoices are now account.moves, user must set a "reference" field in invoices, or "communication" field in bank.payment.lines wouldn't be set. If communication fields are not set, an error is occurring as a mandatory field hasn't been set.
> 
> In previous versions, communication field was inherited from invoice numbers if I'm right. But now, it inherits the "reference" field.
> 
> Do you think that any info should be shown in invoices when user selects a payment mode related to a "sepa_credit_transfer" payment method? I was thinking about that, as I don't think we should set "reference" field as mandatory.
> 
> See attached images:
> 
> Captura de pantalla 2020-06-09 a las 12 55 19
> 
> This is my suggestion:
> 
> Captura de pantalla 2020-06-09 a las 12 44 47
> 
> What do you think @ValentinVinagre @pedrobaeza @sergio-teruel ?
> 
> THX



> Hi @pedrobaeza , @HaraldPanten ,@sergio-teruel
> I think that the problem is here, for some reason the code detects the 'ref' field isn't empty, and doesn't choose the "name" field.
> 
> https://github.com/OCA/bank-payment/blob/080bad77d6eedb031004a993743ebf5356821102/account_payment_order/models/account_move_line.py#L52-L64

Hi @pedrobaeza @carlosdauden what do you think about these suggestions?

Thanks!

HaraldPanten:
Could you have a look at our last comments @pedrobaeza @carlosdauden ?

THX!

pedrobaeza:
@ValentinVinagre please give me permissions to this branch for pushing the solution for `l10n_generic_coa` hidden dependency.

pedrobaeza:
About the communication, it shouldn't be handled here, as it's something independent from this module, but a general question on `account_payment_order`. Please put an issue with the detailed steps to reproduce the problem.

ValentinVinagre:
> ase give me permissions to this branch for pushing the solution for `l10n_generic_coa` hidden dependency.

Done.

pedrobaeza:
@sergio-teruel @carlosdauden now it's independent from `l10n_generic_coa`.

pedrobaeza (approved):

HaraldPanten:
> About the communication, it shouldn't be handled here, as it's something independent from this module, but a general question on `account_payment_order`. Please put an issue with the detailed steps to reproduce the problem.

Done 

Related issue, RFO: https://github.com/OCA/bank-payment/issues/719

carlosdauden (approved):

PR #686, opened 2020-03-06
[13.0][MIG] account_banking_sepa_direct_debit

misern2:
~~Depends on  #670~~

PR #700, opened 2020-04-22
[13.0][FIX] account_payment_order: fix get partner_bank_id in account move lines

### LINKED ISSUE #701: [13.0] User can't "Add to payment/debit orders" the invoices from tree view (account_payment_order)
I was reviewing https://github.com/OCA/bank-payment/pull/686 and I realized that this functionality is not working for V13 as it was in previous versions.

User can't add invoices to payment/debit orders from the tree view but it's still possible to add them one by one from the form view. IMHO it was an interesting feature that should be maintained.

Do you think that would be appropriate to suggest a [FIX] for that?

Thx.

ValentinVinagre:
Fix get partner_bank_id in account move lines.

**Case:**
The bank account isn't set properly in the payment order lines, because the bank account is only got from account.move.line or the first bank account in the partner_id, but never from the invoice (account.move).

**Solution:**
With this fix, the bank account will be get from:

1. account.move.line
2. account.move
3. first bank account in the partner_id

This fix contain too https://github.com/OCA/bank-payment/issues/701
This fix contain too: fix the attachment file name. https://github.com/OCA/bank-payment/pull/686

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -2,7 +2,7 @@
 # © 2014 Serv. Tecnol. Avanzados - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
 from odoo.fields import first
 
 
@@ -12,6 +12,9 @@ class AccountMoveLine(models.Model):
     partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
         string="Partner Bank Account",
+        compute="_compute_partner_bank_id",
+        readonly=False,
+        store=True,
         help="Bank account on which we should pay the supplier",
     )
     bank_payment_line_id = fields.Many2one(
@@ -22,6 +25,23 @@ class AccountMoveLine(models.Model):
         inverse_name="move_line_id",
         string="Payment lines",
     )
+
+    @api.depends(
+        "move_id", "move_id.invoice_partner_bank_id", "move_id.payment_mode_id"
+    )
+    def _compute_partner_bank_id(self):
+        for ml in self:
+            if (
+                ml.move_id.type in ("in_invoice", "in_refund")
+                and not ml.reconciled
+                and ml.payment_mode_id.payment_order_ok
+                and ml.account_id.internal_type in ("receivable", "payable")
+                and not any(
+                    p_state in ("draft", "open", "generated")
+                    for p_state in ml.payment_line_ids.mapped("state")
+                )
+            ):
+                ml.partner_bank_id = ml.move_id.invoice_partner_bank_id.id
 
     def _prepare_payment_line_vals(self, payment_order):
         self.ensure_one()
--- a/account_payment_order/views/ir_attachment.xml
+++ b/account_payment_order/views/ir_attachment.xml
@@ -12,12 +12,7 @@
                         <field name="name" />
                     </h1>
                     <group name="main">
-                        <field
-                            name="datas"
-                            filename="store_fname"
-                            string="Generated File"
-                        />
-                        <field name="store_fname" invisible="1" />
+                        <field name="datas" filename="name" string="Generated File" />
                         <label for="create_uid" string="Created by" />
                         <div name="creation_div">
                             <field name="create_uid" readonly="1" class="oe_inline" />
--- a/account_payment_order/views/account_invoice_view.xml
+++ b/account_payment_order/views/account_invoice_view.xml
@@ -57,7 +57,6 @@
         name="Add to Payment/Debit Order"
         res_model="account.invoice.payment.line.multi"
         binding_model="account.move"
-        binding_views="form"
         view_mode="form"
         target="new"
         id="account_invoice_create_account_payment_line_action"


ValentinVinagre:
Hi @pedrobaeza @carlosdauden @Tardo 
Can you check this fix? is related to 13 MIG account banking sepa credit transfer.
Thanks a lot,

pedrobaeza:
I'm not sure about the patch. If `self.partner_bank_id` is not set, the problem is that the transfer from `invoice_partner_bank_id` to the line has not been correct.

ValentinVinagre:
The problem I found appears when I run the test. "account.move.line" doesn't have invoice_partner_bank_id set. With this patch this problem is corrected. 
But I don't know if this patch needs to be applied when "account.move.line" is created (invoice validate).
But I agree to change that, if there is any other better way.

pedrobaeza:
Is the test failing here in Travis?

ValentinVinagre:
The test fails when check the partner_bank_id in the module "account banking sepa credit transfer" if this patch are not applied. https://github.com/OCA/bank-payment/pull/699
Here is the link:
https://travis-ci.org/github/OCA/bank-payment/jobs/678673610

Header explains the fail.

HaraldPanten:
Hi @ValentinVinagre ,

Is it OK for you, to add https://github.com/OCA/bank-payment/issues/701 in this PR, please?

Thank you in advance! 👍

ValentinVinagre:
Hi @HaraldPanten ,
no problem to fix, but I don't know if I need open other PR for resolve the https://github.com/OCA/bank-payment/issues/701 or I can add the solution in this fix, @pedrobaeza how should I proceed?

Thanks,

ValentinVinagre:
This fix contain too https://github.com/OCA/bank-payment/issues/701

ValentinVinagre:
Hi @pedrobaeza ,
Have you been able to review this problem?

Thank you,

HaraldPanten (approved):
Functional review 👍 

IMO, both fixes are working fine.

HaraldPanten (approved):
Tested again with the new fix in local environment.

Functional review OK. 👍

HaraldPanten:
Hi @pedrobaeza could you have a look? This fix is blocking several dependencies.

Thanks in advance!

pedrobaeza:
This has 3 things, and I don't agree with the main one: we shouldn't change the way payment lines are prepared, but see why move line is not having propagated the bank account from invoice (account.move) to the AR/AP account.move.line. This is the patch to do.

ValentinVinagre:
I will check to see if I can fix it in another way.

ValentinVinagre:
Hi @pedrobaeza ,
I see in v.12 that fields are set in https://github.com/OCA/bank-payment/blob/d5f9c89399a4c8fbd55b876e7d26039c6f80c17b/account_payment_order/models/account_invoice.py#L51 but this functionality has not been implemented in the migration to v.13.

I added this field when the user posts the "invoices". Is this solution more suitable?

pedrobaeza:
Please don't do it on post, but on account.move.line creation, where there are hooks for that, and I would say that the same hook is used for updating values. You can put also a computed writable field if you prefer it. See for example my refactoring using this new feature in this module, that saves tons of code:

https://github.com/OCA/commission/pull/237/commits/081678a50f98c7a75ffe13809e4875558e60c7d0

ValentinVinagre:
Hi @pedrobaeza ,
done.

account_payment_order/views/ir_attachment.xml
```diff
@@ -12,12 +12,7 @@
                         <field name="name" />
                     </h1>
                     <group name="main">
-                        <field
```
StefanRijnhart, line 15: Can I ask if this alternative attachment view is in need of a value for the field `priority` so that it does not compete with the main attachment form view?

account_payment_order/models/account_move_line.py
```diff
@@ -23,6 +26,25 @@ class AccountMoveLine(models.Model):
         string="Payment lines",
     )
 
+    @api.depends(
+        "move_id", "move_id.invoice_partner_bank_id", "move_id.payment_mode_id"
```
StefanRijnhart, line 30: Isn't the partner_bank_id redundant now that Odoo added an invoice_partner_bank_id on the move level?

ValentinVinagre: I don't think so, I explain one situation:
You have to pay your employees, so you will create an "account move" to generate this movement. Once you have to pay it, you want to associate the "account.move.line" to a payment order, so you have to be able to associate a "bank account" to an "account.move.line".

account_payment_order/views/ir_attachment.xml
ValentinVinagre: it's possible but this FIX have 3 fix's and I don't want to add more. If the view haver any problem open other fix please.

StefanRijnhart: Yes, I was wondering about adding this second fix but given that you are modifying this view you could either split off all view fixes into a separate PR or add the priority + view selection in this PR.

account_payment_order/models/account_move_line.py
StefanRijnhart: Yes, good point.

account_payment_order/models/account_move_line.py
```diff
@@ -12,6 +12,9 @@ class AccountMoveLine(models.Model):
     partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
         string="Partner Bank Account",
+        compute="_compute_partner_bank_id",
```
StefanRijnhart, line 15: I think making this a computed field could maybe lead to unexpected results. Would you consider, instead of updating the move line's partner_bank_id directly to change the logic here: https://github.com/OCA/bank-payment/blob/13.0/account_payment_order/models/account_move_line.py#L58 and fetch either the move line's partner_bank_id when it's set, or the move's invoice_partner_bank_id, or else any of the partner's banks?

ValentinVinagre: That solution is the first solution he presents, but he wants to solve the reason why he did not transfer the bank account, therefore it was not well received. Now the third solution is presented, you can read the solution conversation where this solution is recommended.

StefanRijnhart: Oh sh*t that was your first proposal but @pedrobaeza did not like it. Well, then I will refrain from reviewing this PR further (refer to deleted commit https://github.com/OCA/bank-payment/commit/b7b08ac44672bf51968e974998d5e6aca2877c22)

account_payment_order/views/ir_attachment.xml
ValentinVinagre: If you tell me I should change, I can add the changes in this solution 👍

account_payment_order/models/account_move_line.py
ValentinVinagre: Haha no problem 👍

account_payment_order/views/ir_attachment.xml
StefanRijnhart: Thanks, but you are right about combining too many different fixes in the same PR. I've proposed this priority thing here: https://github.com/OCA/bank-payment/pull/712/files

HaraldPanten (approved):
👍 IMO. What do you think @StefanRijnhart ?

StefanRijnhart:
> +1 IMO. What do you think @StefanRijnhart ?

Like I said, I like the earlier version better. Seems to me that the move's invoice_partner_bank_id serves a different purpose than the move line's partner_bank_id and trying to keep them synced might lead to unexpected behaviour. But I'm refraining from review so that I'm not blocking the merge of this.

ValentinVinagre:
Hi @pedrobaeza ,
Could you review? I change the "partner_bank_id" field by adding a computed method.

HaraldPanten:
Hi @pedrobaeza we applied the changes that you suggested. Is there anything else we need to do before merging this FIX? It's blocking other PR's. Could you have a look?

Thanks!

account_payment_order/models/account_move_line.py
```diff
@@ -23,6 +26,25 @@ class AccountMoveLine(models.Model):
         string="Payment lines",
     )
 
+    @api.depends(
+        "move_id", "move_id.invoice_partner_bank_id", "move_id.payment_mode_id"
+    )
+    def _compute_partner_bank_id(self):
+        for ml in self:
+            res = None
```
pedrobaeza: Better to avoid possible side effects using a falsy but not expected value

```suggestion
            res = False
```

account_payment_order/models/account_move_line.py
```diff
@@ -23,6 +26,25 @@ class AccountMoveLine(models.Model):
...
+                and not ml.reconciled
+                and ml.payment_mode_id.payment_order_ok
+                and ml.account_id.internal_type in ("receivable", "payable")
+                and not any(
+                    p_state in ("draft", "open", "generated")
+                    for p_state in ml.payment_line_ids.mapped("state")
+                )
+            ):
+                res = ml.move_id.invoice_partner_bank_id.id
+            ml.partner_bank_id = res
```
pedrobaeza: Anyway, I think you mustn't assign always the bank account, but only when matching the criteria. If not, you are emptying the field in the move line is linked to a payment line for example.

pedrobaeza (changes requested):
And the other unrelated changes?

account_payment_order/models/account_move_line.py
ValentinVinagre: done.

ValentinVinagre:
@pedrobaeza changes done.

pedrobaeza:
And what about the other unrelated changes? Why are they here?

ValentinVinagre:
The unrelated changes are: 
This fix contain too #701 
This fix contain too: fix the attachment file name. #686

I put all the changes in the same PR because they are very small fixes.
If you prefer I can create other 2 PR's with this fixes, and only leave the 'partner_bank_id' fix in this PR.

pedrobaeza:
Well, they can go in the same PR being small, but different commits explaining each fix. Current commit message is:

- Very short. Please include the use case and description you have put in the PR main comment. When browsing commit history, you don't see the PRs where they were commented.
- Only referring to the first fix, not the others. But as said, it's preferable in this case to have a commit for each fix.

ValentinVinagre:
> Well, they can go in the same PR being small, but different commits explaining each fix. Current commit message is:
> 
> * Very short. Please include the use case and description you have put in the PR main comment. When browsing commit history, you don't see the PRs where they were commented.
> * Only referring to the first fix, not the others. But as said, it's preferable in this case to have a commit for each fix.

Done in 3 commits and each one has been commented.

pedrobaeza (approved):
Thanks for the efforts. Now it's all homogenized and well conducted.

PR #709, opened 2020-05-15
[13.0][FIX] account_banking_mandate: Set Force company

etobella:
Missing `force_company`

### DIFF
--- a/account_banking_mandate/models/account_move.py
+++ b/account_banking_mandate/models/account_move.py
@@ -35,6 +35,7 @@ class AccountMove(models.Model):
         for onchange_method, changed_fields in list(onchanges.items()):
             if any(f not in vals for f in changed_fields):
                 invoice = self.new(vals)
+                invoice = invoice.with_context(force_company=invoice.company_id.id,)
                 getattr(invoice, onchange_method)()
                 for field in changed_fields:
                     if field not in vals and invoice[field]:


pedrobaeza:
In v13 this is done other way I think: https://www.odoo.com/documentation/13.0/howtos/company.html

account_banking_mandate/models/account_move.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Marçal Isern <marsal.isern@qubiq.es>
+ # Copyright 2020 Marçal Isern <marsal.isern@qubiq.es>
```
StefanRijnhart: unintended whitespace?

StefanRijnhart:
> In v13 this is done other way I think: https://www.odoo.com/documentation/13.0/howtos/company.html

Can you clarify @pedrobaeza ? That page still mentions force_company, and so does the code https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_property.py#L185

pedrobaeza:
Yeah, my fault. The change is from saas-13.2/v14, that `with_company` is introduced.

pedrobaeza:
FYI: https://www.odoo.com/documentation/master/howtos/company.html

pedrobaeza:
Any way, CI is red. I plan to work on moving these modules to computed writable fields during Spanish OCA Days, so if you want to wait (and if you participate in the CS), we can make directly the switch for both.

StefanRijnhart:
Thanks for the clarification. I assume that if pre-commit nags are fixed quickly this can be merged in the meantime.

etobella:
Now it should be green

StefanRijnhart (approved):

pedrobaeza (approved):

PR #711, opened 2020-05-15
[FIX] Onchange methods can be called on newIds that have no company yet

StefanRijnhart:
The correct payment mode is not fetched here: https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L472

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -22,9 +22,12 @@ class AccountInvoice(models.Model):
     def _onchange_partner_id(self):
         res = super(AccountInvoice, self)._onchange_partner_id()
         if self.partner_id:
+            company_id = (self.company_id.id or
+                          self.env.context.get('force_company') or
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I don't get the use case where you have the partner and not the company (being that one set on a default value), but any way, this is harmless, so approving.

StefanRijnhart:
I linked to it: https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L472-L475. In sale.order's `_prepare_invoice` the onchange is called on a new() that has no value for `company_id`, i.e. `self.company_id` = an empty recordset.

pedrobaeza:
Ah, I see.. all this clunky stuff due to onchanges. `common.test.Form` class is an advance in this (or `onchange_helper`), but it's not used here. Luckily for v14 we will get rid off all of this thanks to the whole refactoring to computed writable.

astirpe (approved):
LGTM

PR #713, opened 2020-05-15
[12.0][FIX] Form view of mandate had 2 "type" fields

alexis-via:


account_banking_sepa_direct_debit/views/account_banking_mandate_view.xml
```diff
@@ -17,12 +17,13 @@
...
             <field name="scheme" attrs="{'invisible': [('format', '!=', 'sepa')],
                                          'required': [('format', '=', 'sepa')]}"/>
-            <field name="type" attrs="{'invisible': [('format', '!=', 'sepa')],
-                                       'required': [('format', '=', 'sepa')]}"/>
             <field name="recurrent_sequence_type"
                    attrs="{'invisible': ['|', ('type', '=', 'oneoff'), ('format', '!=', 'sepa')],
                            'required': [('type', '=', 'recurrent')]}"/>
         </field>
+        <field name="type" position="attributes">
+            <attribute name="attrs">{'required': [('format', '=', 'sepa')]}</attribute>
```
pedrobaeza, line 25: What about the `'invisible': [('format', '!=', 'sepa')]` part?

alexis-via: Well, type field is now part of account_banking_mandate. So I would be strange to hide it in an upper-level module.

pedrobaeza: Yes, makes sense

pedrobaeza (approved):

HaraldPanten (approved):

PR #716, opened 2020-06-18
ADD 13.0 account_invoice_select_for_payment

gurneyalex:
This module allows to mark invoices as "selected for payment". This can be done
in the list view of invoices using a button in the first column of the view
which shows the selection status. This selection persists until a payment is
registered.


grindtildeath (approved):
A small test would be good, but LGTM :+1:

HaraldPanten (changes requested):
Hi @gurneyalex I would suggest some changes:

1- IMO you should show the column name in account.view_invoice_tree. Don't you think so? Find attached a screenshot:

Captura de pantalla 2020-06-19 a las 10 19 03


2- When an invoice has been set to "paid" status, I think that the "selected for payment" checkbox should be clickable, or should be invisible. What do you think?

3- Could you give me some details about what's the need/objective of this module? Aren't the due dates enough to solve this need? (Just asking 😇 )

simahawk:
@gurneyalex ping

gurneyalex:
@HaraldPanten  regarding the need / objective of this module: this has been developped for a customer with a large number of supplier invoices, where the decision for making a bill payable is taken manually after checking that the goods or services were received. Then typically once per month, a different person will take the invoices flagged for payment and issue a payment order for the various suppliers. The due dates are not enough, as you typically don't want to pay an invoice if the service was not fully delivered.

HaraldPanten:
THX for the explanation @gurneyalex . Then, I suggest you to add a name for the column in the tree view. WDYT?

pedrobaeza:
Why not using the existing check in account.move.line "Litigation" (or similar)? The module already excludes these ones by default.

gurneyalex:
@pedrobaeza are you referring to the "blocked" (No Follow-up) field? This field serves another purpose in EE's account_followup,  and I don't want to mix the concerns.

gurneyalex:
@HaraldPanten  AFAIK you cannot add a column header for a button column. My last commit is changing this to use a single button to toggle the state, which will hopefully make a clearer interface.

pedrobaeza:
Yeah, that one. OK if it doesn't serve for you. For me, not having EE account_followup, serves for the purpose.

HaraldPanten:
> @HaraldPanten AFAIK you cannot add a column header for a button column. My last commit is changing this to use a single button to toggle the state, which will hopefully make a clearer interface.

Huummm... Unfortunately it's out of my knowledge as well. Maybe @pedrobaeza could shed some light here 🤔

pedrobaeza:
If it's a field, you can use toggle_buton widget, that integrates well in tree views.

HaraldPanten (approved):
Looks nice! 👍

jgrandguillaume (approved):
LGTM thx for the addition

PR #723, opened 2020-08-04
[IMP] account_banking_sepa_direct_debit: test at install

yajo:
As a remainder from 8d1cb348fcc35c614dd1c7e3dc6c5fdab92268f2 (possibly trying to fix the wrong problem), this test was being ran as post install. This broke some integration environments without need.

- Removed post install mode.
- Improved to use `SavepointCase`.

@Tecnativa TT23969

### DIFF
--- a/account_banking_sepa_direct_debit/tests/test_sdd.py
+++ b/account_banking_sepa_direct_debit/tests/test_sdd.py
@@ -9,40 +9,38 @@ import time
 from lxml import etree
 
 
-class TestSDD(common.HttpCase):
-    at_install = False
-    post_install = True
-
-    def setUp(self):
-        super(TestSDD, self).setUp()
-        self.company = self.env['res.company']
-        self.account_model = self.env['account.account']
-        self.move_model = self.env['account.move']
-        self.journal_model = self.env['account.journal']
-        self.payment_order_model = self.env['account.payment.order']
-        self.payment_line_model = self.env['account.payment.line']
-        self.mandate_model = self.env['account.banking.mandate']
-        self.bank_line_model = self.env['bank.payment.line']
-        self.partner_bank_model = self.env['res.partner.bank']
-        self.attachment_model = self.env['ir.attachment']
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

pedrobaeza:
For reducing the diff, please don't rename all `self` occurrences by `cls`, just add a first line like:
```python
self = cls
```
I wonder if this will break other things, and we only having one DB in 10.0, not sure if it's worth.

yajo:
I prefer to have more diff but let code be more explicit.

pedrobaeza:
The code is the same explicit... Anyway, as we are moving the customer to upper versions, we can let this as PR and never merge.

pedrobaeza:
Look your inconsistency also: here you want to be "more explicit", and in OCA/maintainer-tools#466 you want to be less explicit? Choose one.

yajo:
https://github.com/OCA/maintainer-tools/pull/466 is unrelated. Also, the point that Tecnativa has more or less customers is not relevant for OCA when judging this PR. Personally, when a PR is not interesting to me because it doesn't affect our customers, I just unsubscribe and let other members review; you can do so if this is not interesting to you :man_shrugging: . In any case, please don't mix subjects.

Could we please focus on the PR? Is there any real blocking point, or can we merge? :roll_eyes:

pedrobaeza:
It was related about consistency in the principles used: you should be in general explicit or implicit, and Python already told us about being explicit, so it would be OK in this PR, but not in the other (that you remove the explicit part). However, this one can be considered an exception, as it's an old branch with a very specific case that can be solved with few code lines, so this is why I asked for not doing such diff.

Anyway, outside of the diff, I still think that this can produce side effects, so it's better to keep it as a PR as we only needed for one customer CI that is in process of being migrated, and we can add this PR to it.

pedrobaeza:
Let's close it for having a clean house in our tracker, but not removing the branch for their usage in the CI.

PR #724, opened 2020-08-04
[IMP] account_banking_pain_base: switch from @api.model to @api.multi

sebalix:
The calls are the following:
`generate_payment_file()` -> `generate_party_block(...)` -> `generate_party_agent(...)`

The first was decorated with `api.multi` (with ensure_one) and the others with `api.model`.

This change allows to get the reference of the current payment order in
`generate_party_block(...)` and `generate_party_agent(...)` methods (for
instance to modulate the generation of the file depending on the related
payment method configuration).

Used by: https://github.com/OCA/l10n-switzerland/pull/558

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -340,7 +340,6 @@ class AccountPaymentOrder(models.Model):
                 % self.company_id.name)
         return True
 
-    @api.model
     def generate_party_agent(
             self, parent_node, party_type, order, partner_bank, gen_args,
             bank_line=None):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
OK for me in the general principle, but:

- Don't add `self.ensure_one()`. Maybe you are passing a multi-record or want to in a future implementation. As current code is not limited to one record, it doesn't make sense.
- Remove `@api.model` instead of changing it to `@api.multi`. The default is this last one.

sebalix:
@pedrobaeza done

pedrobaeza (approved):

astirpe (approved):
Ok to remove the `self.ensure_one()` but then remember to adapt the code in `l10n_ch_pain_base` (https://github.com/OCA/l10n-switzerland/pull/558/files#diff-5d36e3d6e4ebf67a6a0490d4a52478d0R71).

yvaucher (approved):

PR #725, opened 2020-08-13
[FIX] 10.0 - prevent partner_bank_id to stay undefined

lmarion-source:
In some cases, mandate can exist without a partner_bank_id. In this case, local variable 'partner_bank_id' was never initialize, causing a bug. Now, 'partner_bank_id' is initialize to False

### DIFF
--- a/account_banking_mandate/models/account_move_line.py
+++ b/account_banking_mandate/models/account_move_line.py
@@ -20,6 +20,7 @@ class AccountMoveLine(models.Model):
         if payment_order.payment_type != 'inbound':
             return vals
         mandate = self.mandate_id
+        partner_bank_id = False
         if not mandate and vals.get('mandate_id', False):
             mandate = mandate.browse(vals['mandate_id'])
         if not mandate:


account_banking_mandate/models/account_move_line.py
```diff
@@ -20,6 +20,8 @@ def _prepare_payment_line_vals(self, payment_order):
         if payment_order.payment_type != 'inbound':
             return vals
         mandate = self.mandate_id
+        # Initilization of partner_bank_id
```
astirpe: Typo:

```suggestion
        # Initialization of partner_bank_id
```

astirpe (changes requested):

account_banking_mandate/models/account_move_line.py
astirpe: I think you can even remove this comment, as the code is self-explanatory

sbidoul (approved):
I'd remove the comment too. It is the kind of change that is best explained in the PR or commit message than a code comment.

astirpe (approved):

astirpe:
Could you squash the two commits together?

Issue #718, opened 2020-06-23
[12.0]account_payment_order_return


javierjcf:
https://github.com/OCA/bank-payment/blob/dc8a3ed1e0739c4b40ab59bfeb160541ca24dc22/account_payment_order_return/wizards/account_payment_line_create.py#L18

I think this is a bug. If i dont check the self.include_returned field, the condition is true, so  ('invoice_id.returned_payment', '=', False) is added to the domain.
I think is a problem, because a lot of account.moves.lines are not linked to an invoice, so invoice_id.returned_payment is False BUT because invoice_id is False, so this moves are not found.
I need to check always the include_returned field in order to found the moves.

Im not sure about all, but is a problem that already found

colmeirin:
Hello, I can confirm I experienced the same behaviour which is very odd with move lines that are not linked to invoices such as payrolls payments or expenses.

pedrobaeza:
Please add a PR that modifies the domain this way:
```python
domain += [
    '|',
    ('invoice_id', '=', False),
    ('invoice_id.returned_payment', '=', False)
]
```

colmeirin:
Fix has been merged. I think this can be close now @javierjcf

PR #726, opened 2020-08-19
fix domain in payment_order_return

colmeirin:
fix domain in payment_order_return in order to avoid problems with moves without associated invoice as discused at #718


pedrobaeza (approved):

AaronHForgeFlow (approved):
:+1:  Makes sense

PR #727, opened 2020-08-24
[10.0][IMP] set a filename according to the name of the mandate

lmignon:
In the same time, this change allows to preserve the original name of the scanned file.


rousseldenis (approved):

rafaelbn (approved):
👍 Thank! Sorry such a delay!

PR #728, opened 2020-09-07
[13.0][IMP] account_banking_sepa_direct_debit: Remove l10n_generic_coa hidden dependency in tests

pedrobaeza:
As we were setting a new company with its CoA, we need to load in it one, but using l10n_generic_coa doesn't serve, as we don't depend on that module (and we shouldn't), but we are invoking its XML-ID for loading the CoA, trusting in the good faith that the module is available and installed, which is not the case in some cases.

Luckily, Odoo provides already an inheritable test for having a minimal CoA that doesn't require to load any chart. This PR uses that test and performs the minimal changes to adapt the module tests to it.

@Tecnativa

### DIFF
--- a/account_banking_sepa_direct_debit/tests/test_sdd.py
+++ b/account_banking_sepa_direct_debit/tests/test_sdd.py
@@ -7,15 +7,18 @@ import base64
 from lxml import etree
 
 from odoo import fields
-from odoo.tests import common
 from odoo.tools import float_compare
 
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):

HaraldPanten (approved):
Looks reasonable

carlosdauden (approved):

PR #730, opened 2020-09-08
[FIX] _check_payment_mode_company_constrains

mileo:
Can breaks multi company environments, for example: https://travis-ci.org/github/OCA/l10n-brazil/jobs/722154732#L2375

### DIFF
--- a/account_payment_partner/models/account_invoice.py
+++ b/account_payment_partner/models/account_invoice.py
@@ -21,10 +21,8 @@ class AccountInvoice(models.Model):
     @api.onchange('partner_id', 'company_id')
     def _onchange_partner_id(self):
         res = super(AccountInvoice, self)._onchange_partner_id()
-        if self.partner_id:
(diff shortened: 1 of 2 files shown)


renatonlima (commented):
@mileo This is not necessary because in module account_payment_mode has one ir.rule for multi-company https://github.com/OCA/bank-payment/blob/061e1dcddadf6cd2a48177bee4075212a4a38a00/account_payment_mode/security/account_payment_mode.xml#L4

mileo:
Onchange of odoo/sale: https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L477-L490

Is called without company, than this can handle some unexpected behavior at multi company environment. 

When the company of the sales order is different from the user who confirms it.

marcelsavegnago (approved):

jpcarassato (approved):

luismalta (approved):

PR #731, opened 2020-09-10
FIX:add to action list view for payment order

cormaza:
Adds to action list view for payment order, in version 13 you must specify in which type of view the action can be displayed


pedrobaeza:
Hasn't this a default if not specified?

PR #732, opened 2020-09-11
[IMP] account_banking_pain_base: use account holder name in party block

robinkeunen:
For some partners, the account name holder can be different than the partner name.

If that alternative name (`acc_holder_name`) is set on the bank account, use it.

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -449,7 +449,7 @@ class AccountPaymentOrder(models.Model):
             party_type_label = _("Creditor name")
         elif party_type == 'Dbtr':
             party_type_label = _("Debtor name")
-        name = 'partner_bank.partner_id.name'
+        name = 'partner_bank.acc_holder_name or partner_bank.partner_id.name'
         eval_ctx = {'partner_bank': partner_bank}
         party_name = self._prepare_field(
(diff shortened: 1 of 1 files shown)


account_banking_pain_base/models/account_payment_order.py
```diff
@@ -449,7 +449,12 @@ def generate_party_block(
             party_type_label = _("Creditor name")
         elif party_type == 'Dbtr':
             party_type_label = _("Debtor name")
-        name = 'partner_bank.partner_id.name'
+
+        if partner_bank.acc_holder_name:
+            name = 'partner_bank.acc_holder_name'
+        else:
+            name = 'partner_bank.partner_id.name'
+
```
pedrobaeza: I think you can put an expression there:
```suggestion
        name = 'partner_bank.acc_holder_name or partner_bank.partner_id.name'
```
and please don't introduce empty lines inside a method.

robinkeunen: Yep, that works, here you go.

pedrobaeza (approved):

houssine78 (approved):
LGTM

bealdav (approved):

pedrobaeza:
/ocabot merge patch

Please forward port it to v13

PR #733, opened 2020-09-11
[IMP] ab_pain_base: use account holder name in party block

robinkeunen:
Port of https://github.com/OCA/bank-payment/pull/732

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -497,7 +497,7 @@ class AccountPaymentOrder(models.Model):
             party_type_label = _("Creditor name")
         elif party_type == "Dbtr":
             party_type_label = _("Debtor name")
-        name = "partner_bank.partner_id.name"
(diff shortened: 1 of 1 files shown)


astirpe (approved):

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -497,7 +497,7 @@ def generate_party_block(
             party_type_label = _("Creditor name")
         elif party_type == "Dbtr":
             party_type_label = _("Debtor name")
-        name = "partner_bank.partner_id.name"
+        name = 'partner_bank.acc_holder_name or partner_bank.partner_id.name'
```
pedrobaeza: This will fail pre-commit, as it uses black with double quote instead of simple.

robinkeunen: fixed.

pedrobaeza (approved):

PR #734, opened 2020-09-11
[13.0][FIX] account_payment_order: avoid cache miss in computed field

astirpe:
We should always ensure to assign something in a compute method

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -42,6 +42,8 @@ class AccountMoveLine(models.Model):
                 )
             ):
                 ml.partner_bank_id = ml.move_id.invoice_partner_bank_id.id
+            else:
+                ml.partner_bank_id = ml.partner_bank_id
 
     def _prepare_payment_line_vals(self, payment_order):
         self.ensure_one()


account_payment_order/models/account_move_line.py
```diff
@@ -42,6 +42,8 @@ def _compute_partner_bank_id(self):
                 )
             ):
                 ml.partner_bank_id = ml.move_id.invoice_partner_bank_id.id
+            else:
+                ml.partner_bank_id = False
```
pedrobaeza: Being stored=True, it shouldn't be needed. But anyway, it's better to assign the same previous value:
```suggestion
                ml.partner_bank_id = ml.partner_bank_id
```

pedrobaeza: Maybe the problem is the absence of the default?

pedrobaeza:
Thinking more on this, I believe the problem is the absence of the default, and with my first proposed change is still not going to work. Do you have any test to exercise the problem?

astirpe:
I'm not able to trigger the error with an automated test.
Anyway this is a well-known issue of V13 and is not fixable by adding a default. In Odoo V13 the computed fields were all fixed the same way, see for example [this fix](https://github.com/odoo/odoo/pull/48947/files) and [this other fix](https://github.com/odoo/odoo/pull/38463/files#diff-a6a62c953c1d9ceb1ccf6a3ec670578bR41-R42).

pedrobaeza:
I think the problem is the absence of a default value. Can it be?

pedrobaeza (approved):
Anyway, as we assign the same value, and this is avoiding the problem, it's OK for me.

CasVissers-360ERP (approved):
Faced the same issue. Fixed the problem.

PR #737, opened 2020-09-17
[12.0][FIX]account_banking_mandate

AaronHForgeFlow:
I cannot select a mandate of the partner in a manually created invoice. I can only select mandates for his/her company.

This issue happens when the partner is not under any company and you assign a mandate to it. However, after some days, the partner is set under a company, but the mandate is still under the partner.

It is not possible to change the partner of the mandate. The only solution is to delete the mandate and create a new one but users don't want that.

This solution does not seem harmful to me, however I am not sure if it is worth it to develop something that changes the partner of the mandate when doing such changes in the partner.

### DIFF
--- a/account_banking_mandate/views/account_invoice_view.xml
+++ b/account_banking_mandate/views/account_invoice_view.xml
@@ -13,7 +13,7 @@
     <field name="arch" type="xml">
         <field name="partner_bank_id" position="after">
             <field name="mandate_id"
-                domain="[('partner_id', '=', commercial_partner_id), ('state', '=', 'valid')]"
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
This change leads to payment lines with mandates that do not match the partner of the outstanding accounting entry. That seems undesirable to me.
I think demoting a partner from commercial partner to non-commercial partner is not supported at all in Odoo in the accounting department. For instance, I don't think it is possible to reconcile accounting entries between the original partner and the new commercial partner. So I am not sure we should include support for these corner cases.

pedrobaeza:
Yeah, I agree we shouldn't change this on OCA, so closing.

pedrobaeza:
The workaround is to allow to change the partner's mandate.

PR #739, opened 2020-09-28
[13.0] [IMP] Account payment order one line per invoice

### LINKED ISSUE #738: [13.0] Account payment order grouped creditors entry
**Steps:**

1. Create a payment order for 3 supplier invoices from the same creditor (3 x 100USD)
2. Payment mode is to group transactions
3. Create and validate the payment 
4. Reconciliation booking will contain 4 lines (300 debit / 100 x 3 credit)
5. Unreconcile the invoice from the from view

**Behavior:**
All invoices are unreconciled because complete reconciliation is cancelled.

**Desired behavior:**
Reconciliation booking will contain 6 lines (100x3 debit / 100 x 3 credit)
Only one invoice is unreconciled when unreconciling

CasVissers-360ERP:
[IMP] Create account.move.line per account.move and reconcile per account.move.line to avoid big reconciliations to be cancelled

Fixes #738

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -1,5 +1,6 @@
 # © 2014-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
 # © 2014 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# © 2020 360 ERP - Cas Vissers
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
@@ -24,6 +25,10 @@ class AccountMoveLine(models.Model):
         comodel_name="account.payment.line",
         inverse_name="move_line_id",
         string="Payment lines",
+    )
+    transient_payment_line_id = fields.Many2one(
+        'account.payment.line',
+        string='Inverse Payment Line',
     )
 
     @api.depends(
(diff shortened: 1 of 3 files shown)


pedrobaeza:
This must be an extra option, as the rest don't want this as default behavior.

CasVissers-360ERP:
> This must be an extra option, as the rest don't want this as default behavior.

So when paying 10 invoices grouped, unreconciling one of the invoices should unreconcile all? This is desired behaviour? If so, I will cancel this PR immediately and just build a custom module to fix it.

pedrobaeza:
What I don't get is that a lot of people (including me) is using this suite since a lot (not in v13 yet), without such problem, so is this something specific of v13? Seeing code, I don't think so, so maybe something in your procedure is not correct at all, but what I don't see is to unfold into a lot of account.move.line if not needed.

CasVissers-360ERP:
Now I understand. It's v13 only. In this video I show v12 vs v13:
https://drive.google.com/file/d/1taxChr_d8szEVYA_vDbpzxXlFmYIP5GM/view

CasVissers-360ERP:
@pedrobaeza 
Maybe it's actually a base error? Unreconciling one invoice in a reconciliation works in v12 but doesn't in v13. Maybe I should report it to Odoo first? What do you think?

CasVissers-360ERP:
Reported this to Odoo first, maybe they will fix it in base accounting:
https://drive.google.com/file/d/1aXCJGZEzUKhdFpJdM4pEWh6xf4kNmmbD/view

CasVissers-360ERP:
Fixed by Odoo in:
https://github.com/odoo/odoo/commit/55a5eaf367aaaf9421d2ee003f359ffef7ffae99

PR #742, opened 2020-10-15
[13.0][account_payment_mode] Update "development_status": "Production/Stable"

rafaelbn:
Update "development_status": "Production/Stable"

### DIFF
--- a/account_payment_mode/__manifest__.py
+++ b/account_payment_mode/__manifest__.py
@@ -4,6 +4,7 @@
 {
     "name": "Account Payment Mode",
     "version": "13.0.1.1.1",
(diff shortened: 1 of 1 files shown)


HaraldPanten:
Hi @rafaelbn why this changes?

THX

pedrobaeza:
Because it complies with the rules for that, but nobody has figured out till now to put it:

/ocabot merge patch

HaraldPanten:
> Because it complies with the rules for that, but nobody has figured out till now to put it:
> 
> /ocabot merge patch

That's true, so... the next question is why isn't everybody following these standards in many modules 🤷‍♂️

PR #743, opened 2020-10-16
[14.0] [MIG] account_payment_mode

misern2:
Standard Migration


alexis-via:
I propose a more complete migration here https://github.com/OCA/bank-payment/pull/744

misern2:
@alexis-via You're right. Your PR is more complete :+1:

PR #746, opened 2020-10-22
[FIX] account_banking_mandate: Allow inheritability

carlosdauden:
@Tecnativa TT26356


pedrobaeza (approved):
This is not an improvement, but a fix, as it's breaking inheritance

pedrobaeza:
I fast-track due to the severity for modules like `account_financial_risk`:

/ocabot merge patch

PR #745, opened 2020-10-17
[14.0][MIG] account_payment_partner

misern2:
Standrad Migration

PR #749, opened 2020-10-24
[14.0][MIG] migration of module account_payment_order

Oscar-Valverde:
### Proposed changes
I've migrated the module account_payment_order and its functionalities from Odoo 13.0 to 14.0. I've checked if all unit tests are still running.

### Types of changes
* [ ]  Bugfix (non-breaking change which fixes an issue)
* [ ]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [x]  Migration Update

### Checklist
* [x]  I have read the CONTRIBUTING doc
* [x]  I have signed the CLA
* [x]  Lint and unit tests pass locally with my changes
* [x]  I have added tests that prove my fix is effective or that my feature works
* [x]  I have updated the necessary documentation

### Further comments
This module depends on the account_payment_partner module. That means that there should be a migration of it to 14.0. Because there is already a PR for migration I've added the changes of the PR to my local environment to test my migration.

This PR depends on the current migration PR:

https://github.com/OCA/bank-payment/pull/745


alexis-via:
I propose to use check_company=True on M2O fields with company_id in source and destination, as well as     _check_company_auto = True on the new classes that we define (account.payment.order, account.payment.line, bank.payment.line). I tested it and it works well. I started using it in my PR to migrate account_banking_mandate to v14, cf https://github.com/OCA/bank-payment/pull/758

StefanRijnhart:
Thanks! Tested successfully. Agree with Alexis' comment about using Odoo's mechanisms for company consistency.

Not blocking, but I'm wondering if we should replace the tabs for bank payment lines and journal entries on the payment order form by buttons in a button box on the form.

image

pedrobaeza:
Superseded by #769

PR #752, opened 2020-11-06
[12.0][IMP] account_payment_order: Add compatibility with partner_bank_active

carlosdauden:
@Tecnativa TT26023

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -59,8 +59,14 @@ class AccountMoveLine(models.Model):
             # in this case
         if payment_order.payment_type == 'outbound':
             amount_currency *= -1
-        partner_bank_id = self.partner_bank_id.id or first(
-            self.partner_id.bank_ids).id
+        partner_bank_id = self.partner_bank_id.id
+        if not partner_bank_id:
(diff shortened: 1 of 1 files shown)


account_payment_order/models/account_move_line.py
```diff
@@ -59,8 +59,14 @@ def _prepare_payment_line_vals(self, payment_order):
             # in this case
         if payment_order.payment_type == 'outbound':
             amount_currency *= -1
-        partner_bank_id = self.partner_bank_id.id or first(
-            self.partner_id.bank_ids).id
+        partner_bank_id = self.partner_bank_id.id
+        if not partner_bank_id:
+            bank_ids = self.partner_id.bank_ids
```
pedrobaeza, line 64: But having the module, is this returning full list even with archived banks?

pedrobaeza: Maybe there's a context `active_test=False` around?

pedrobaeza: Seen.

pedrobaeza (approved):

rafaelbn (approved):
Thanks! :+1:

victoralmau (approved):

PR #754, opened 2020-11-24
[12.0][FIX] Fixed small bug in compute_sepa function

wpichler:
sepa flag should get set on the current order - not on self. Causes a bug when called with multiple records


astirpe (approved):

rousseldenis (approved):
Trivial.

@OCA/accounting-maintainers

sergiocorato (approved):

pedrobaeza (approved):

PR #755, opened 2020-12-01
[13.0][FIX] account_banking_mandate: fixes singleton error when posting some records

rescola:
There's an error when posting more than one entry.

### DIFF
--- a/account_banking_mandate/models/account_move.py
+++ b/account_banking_mandate/models/account_move.py
@@ -21,7 +21,8 @@ class AccountMove(models.Model):
     )
 
     def post(self):
-        self.line_ids.mandate_id = self.mandate_id
+        for record in self:
+            record.line_ids.write({"mandate_id": record.mandate_id})
         return super(AccountMove, self).post()
 
     @api.model


sergioliva7 (approved):

danimv5 (approved):
Functional review! LGTM

brendapaniagua:
Functional review done!!

ValentinVinagre (changes requested):
Hi,
Code: LGTM
Commit: Please change the name by [FIX] Banking Mandate: post more than one entry, and explain the error, ex(other fix): It should not be possible to delete a bank account linked to mandate(s).

Thanks.

account_banking_mandate/models/account_move.py
```diff
@@ -21,7 +21,8 @@ class AccountMove(models.Model):
     )
 
     def post(self):
-        self.line_ids.mandate_id = self.mandate_id
+        for sel in self:
+            sel.line_ids.mandate_id = sel.mandate_id
```
pedrobaeza: No need of the loop nor a mapped on relational fields in v13:

```suggestion
        self.line_ids.write{"mandate_id": sel.mandate_id}
```

pedrobaeza (changes requested):
And please put a proper commit and PR title following conventions. Proposal:
```
[13.0][FIX] account_banking_mandate: Propagation of the mandate with multi-recordset

There's an error when posting more than one entry.
```
My doubt now is why propagating to all journal items instead only the AR/AP ones.

rescola:
@ValentinVinagre @pedrobaeza thanks! It's my first PR

pedrobaeza:
Please squash commits into one (not using GitHub interface) and name properly according what I have put. You need to use git in local, as the GitHub options doesn't allow this.

account_banking_mandate/models/account_move.py
rescola: HI @pedrobaeza I was trying, i think it needs to loop, because if you select 10 account.moves with diferent mandate in each one. It have to look for each one and write in the specific account.move. Is not the same for all the records.
Then the first code is good or:

    def post(self):
        for sel in self:
            sel.line_ids.write({"mandate_id": sel.mandate_id})
        return super(AccountMove, self).post()

What do you think?

pedrobaeza: Yeah, totally, forget my distraction. Maybe it's better to use a more distinguishable variable name, but yours is correct:
```python
for record in self:
    record.line_ids.write({"mandate_id": record.mandate_id})
```

rescola:
> And please put a proper commit and PR title following conventions. Proposal:
> 
> ```
> [13.0][FIX] account_banking_mandate: Propagation of the mandate with multi-recordset
> 
> There's an error when posting more than one entry.
> ```
> 
> My doubt now is why propagating to all journal items instead only the AR/AP ones.

@pedrobaeza hi! now I think is ok. Can you check please?

pedrobaeza (approved):

PR #756, opened 2020-12-06
14.0 mig account payment sale

dnplkndll:


alexis-via:
I don't understand why the create() of account.move in made in account_payment_sale (it was introduced in v13).
Missing check_company on the M2O field added on sale.order.
I'll make a PR to fix this.

alexis-via:
and the field invoice_partner_bank_id doesn't exist in v14. It has been renamed !
I'll fix all this.

alexis-via:
And there is now a prepare method in sale.advance.payment.inv, let's use it. I'll fix that too.

alexis-via:
Superseeded by https://github.com/OCA/bank-payment/pull/759
Don't consider that updating the version number is enough to port an OCA module to a new version. There is much more than that to do !

PR #758, opened 2020-12-23
[MIG][14.0] account_banking_mandate

alexis-via:


StefanRijnhart:
Thanks for all your work in these module ports again! Can I take the opportunity of asking what the use is of having the mandate_id on the move line as well as on the move, especially if the move's mandate_id is propagated to the move line's mandate ids. Would it make more sense not to propagate them on `_post` but instead use the one on the move line if it is set with a fallback on the mandate on the move if that one is set?

alexis-via:
@StefanRijnhart I agee with you, it's strange to have mandate_id on both move and move.line. I think we should have it on the move only, and re-organise the code to read it there. I'll do that in the coming days/weeks.

pedrobaeza:
Can you please check Travis?

joao-p-marques:
I think Travis was failing because `account_payment_order` was not migrated yet. It was merged 2 days ago, so re-running the CI should unblock it @pedrobaeza

pedrobaeza:
Yes, restarted and now green. Please review.

StefanRijnhart:
I can file a PR into this one for my suggestion if needed.

account_banking_mandate/__manifest__.py
```diff
@@ -1,13 +1,13 @@
-# Copyright 2014 Compassion CH - Cyril Sester <csester@compassion.ch>
-# Copyright 2014 Tecnativa - Pedro M. Baeza
-# Copyright 2015-16 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
-# Copyright 2017 Tecnativa - Carlos Dauden <carlos.dauden@tecnativa.com>
+# Copyright 2014-2020 Compassion CH - Cyril Sester <csester@compassion.ch>
+# Copyright 2014-2020 Tecnativa - Pedro M. Baeza
+# Copyright 2015-2020 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
+# Copyright 2017-2020 Tecnativa - Carlos Dauden <carlos.dauden@tecnativa.com>
```
joao-p-marques, line 4: You do you change the Copyright from the previous changes? This should be left as is, and you should only add your Copyright with the current year

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -1,6 +1,6 @@
-# Copyright 2014 Compassion CH - Cyril Sester <csester@compassion.ch>
-# Copyright 2014 Tecnativa - Pedro M. Baeza
-# Copyright 2015-16 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
+# Copyright 2014-2020 Compassion CH - Cyril Sester <csester@compassion.ch>
+# Copyright 2014-2020 Tecnativa - Pedro M. Baeza
+# Copyright 2015-2020 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
```
joao-p-marques, line 3: Same here

joao-p-marques (commented):
@alexis-via could you squash the two migration commits into one?

Also, please check the comments bellow.

Otherwise, looks good :+1:

pedrobaeza:
@StefanRijnhart that would be good for finishing this.

StefanRijnhart:
I've created a draft https://github.com/OCA/bank-payment/pull/770. I've processed  @joao-p-marques's comments so we can use it as a superseding PR unless @alexis-via disagrees.

account_banking_mandate/README.rst
```diff
@@ -0,0 +1,103 @@
...
+
+This module depends on :
+* account_payment
+
+This module is part of the OCA/bank-payment suite.
+
+Usage
+=====
+
+To use this module, see menu "Invoicing/Accounting > Payments > Debit Orders"
```
CarlosRoca13, line 53:
```suggestion
To use this module, see menu "Invoicing/Accounting > Customers > Debit Orders"
```

joao-p-marques:
I think we can close this in favour of #770, as @alexis-via seems unresponsive. @pedrobaeza can you do it?

PR #761, opened 2020-12-24
[14.0] [MIG] account_banking_pain_base: Migration to v14

CarlosRoca13:
cc @ Tecnativa TT26412

This migration depends on:

- [x] https://github.com/OCA/bank-payment/pull/769

please @carlosdauden @pedrobaeza review this

PR #762, opened 2020-12-26
[14.0][MIG] account_banking_pain_base

alexis-via:


pedrobaeza:
We already did the port in #761, but both are still with Travis red.

alexis-via:
Sorry for the double migration. I started a few days ago, but I didn't push before migrating SDD and SCT.
FYI, I still plan to make more improvements, in particular on the computed is_sepa field: I want to add an additional check on the IBAN prefix, to make sure that the country is part of the SEPA zone (for example, Turkey uses IBAN, but is not part of the SEPA zone, cf the map on https://en.wikipedia.org/wiki/Single_Euro_Payments_Area).
And I would also like to rework the XML generation.

I don't want to make a simple port, I want to go further than that. But, for those who need to go live on January 1st 2021, they can start to use this code.

pedrobaeza:
OK, what kind of rework on XML generation are you thinking about?

CarlosRoca13:
Please, close this PR and add the new features in other PR that makes the Improvement. https://github.com/OCA/bank-payment/pull/761 was merged yesterday

PR #773, opened 2021-01-14
[14.0] [MIG] account_banking_sepa_direct_debit: Migration to v14

CarlosRoca13:
cc @Tecnativa TT26414

This PR superseds https://github.com/OCA/bank-payment/pull/763

Please, @joao-p-marques  @pedrobaeza can you review this?


account_banking_sepa_direct_debit/models/account_banking_mandate.py
```diff
@@ -61,47 +67,50 @@ def _compute_display_name2(self):
 
     @api.onchange("partner_bank_id")
     def mandate_partner_bank_change(self):
-        for mandate in self:
```
joao-p-marques: Why removing this loop? Shouldn't you do the same for all the records if it is a multi record set? I can see that in another implementation of this it is done: https://github.com/OCA/bank-payment/blob/14.0/account_banking_mandate/models/account_banking_mandate.py#L189

joao-p-marques (approved):
Code review :+1: Just a small comment

Can you squash the administrative commits to reduce the commit noise?

account_banking_sepa_direct_debit/models/account_banking_mandate.py
pedrobaeza: onchanges are supposed to be always singleton. The loop doesn't hurt, but it's not needed as well.

account_banking_sepa_direct_debit/tests/__init__.py
```diff
@@ -1,2 +1,2 @@
-from . import test_sdd
+# from . import test_sdd  # TODO port those tests
```
pedrobaeza: This file must be readded and fixed

pedrobaeza (changes requested):

CarlosRoca13:
The model TestAccountMultiCompanyNoChartCommon has dissapeared on 14.0, so I recover the field and method definition used on this tests from 13.0. Please, @pedrobaeza can you review it again?

account_banking_sepa_direct_debit/tests/test_sdd.py
```diff
@@ -7,18 +7,36 @@
...
-)
 
-
-class TestSDDBase(TestAccountMultiCompanyNoChartCommon):
+class TestSDDBase(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         self = cls
+        self.company_B = self.env["res.company"].create({"name": "Company B"})
```
pedrobaeza: The `self = cls` was a trick for not creating a big diff when the switch to SavepointCase was onde, but you shouldn't use `self` in new lines, and don't change it in existing ones. If any thing to do, is the contrary one:  change all existing `self` to `cls` and remove the alias.

account_banking_sepa_direct_debit/tests/test_sdd.py
```diff
@@ -111,6 +129,123 @@ def setUpClass(cls):
...
+                "user_type_id": user_type_revenue.id,
+                "reconcile": True,
+                "company_id": cls.company_B.id,
+            }
+        )
+
+    @classmethod
+    def setUpAccountJournal(cls):
+        """ Set up some journals: sale, purchase, ... """
+        cls.journal_purchase = cls.env["account.journal"].create(
```
pedrobaeza: I don't think all these journals are really needed here.

pedrobaeza (changes requested):
Please rename the commit message accordingly the changes done.

Why the uncovered lines?

CarlosRoca13:
@pedrobaeza I removed the not needed journals and `self = cls` alias.

pedrobaeza (approved):

PR #781, opened 2021-02-05
[12.0][IMP] Account Payment Order usability improvement

luismalta:
This PR aims to add improvements in the usability of the Account Payment Order.
Developed together with @mileo

### DIFF
--- a/account_payment_order/models/account_invoice.py
+++ b/account_payment_order/models/account_invoice.py
@@ -52,22 +52,7 @@ class AccountInvoice(models.Model):
         return res
 
     @api.multi
-    def _prepare_new_payment_order(self, payment_mode=None):
-        self.ensure_one()
-        if payment_mode is None:
-            payment_mode = self.env['account.payment.mode']
-        vals = {
-            'payment_mode_id': payment_mode.id or self.payment_mode_id.id,
-        }
-        # other important fields are set by the inherit of create
-        # in account_payment_order.py
-        return vals
-
-    @api.multi
     def create_account_payment_line(self):
-        apoo = self.env['account.payment.order']
-        result_payorder_ids = []
-        action_payment_type = 'debit'
         for inv in self:
             if inv.state != 'open':
                 raise UserError(_(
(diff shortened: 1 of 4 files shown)


mileo:
CC: @mbcosta @renatonlima @marcelsavegnago @rvalyi

marcelsavegnago:
@luismalta Can you implement more testing to make everything green?

rvalyi (changes requested):
I didn't analyse everything yet, but I really don't think imposing that `account.move.line` inherit from `mail.thread` by default is any good idea:
https://github.com/OCA/bank-payment/pull/781/commits/43fa37450d589ad0af197f7ed90b4a1a61eaf60c

Historically, mail.thread has been really a performance hostspot because of the way it overrides the create and write ORM methods and then JOIN in many2many tables.

Many companies easily have hundred of thousands of account.move.line, this would not scale well I think.

Also, that can easily be added as an extra module in the case one would really want to and knows exactly what he is doing performance wise.

Finally as I said in another PR, I'm OK to help review, but I don't think the module maintainers and users will be OK to do such a big change in the 12.0 version. IMHO, such large changes (other than the mail.thread thing) are much more welcome in the newest versions rather than legacy ones.

cc @alexis-via @pedrobaeza

pedrobaeza:
I agree about not making account.move.line a thread. The rest should be explained:

- What is doing with words (not with code).
- Why it improves the usability.

alexis-via:
I agree with @rvalyi account.move.line should NOT inherit from mail.thread.

luismalta:
Hello guys. Explaining better what was done in the code of this PR of the mail.thread questions. In general, @mileo identified the need to centralize some features in the AML, the need to have more control over flows and better assign some responsibilities. For this, the following actions were carried out:

- Moved the functionality of create_account_payment_line to AML to decrease multiple accesses of dot notation;
- Functionality to remove a payment line from a payment order that is still in draft;
- Added the payment line field in the AML view, as well as buttons for manually adding the AML to a payment order or removing the related payment line.

rvalyi:
@pedrobaeza in versions 13 and 14, we have a mail.thread at the account.move level. Isn't that enough to track the history of the AMLs? What about the performance in companies with hundreds of thousands of AMLs?
Few of the AML are payment lines, shouldn't we have a design that would track on these lines instead of all the AML?
Will people really browse the AML form page detail to read the mail.thread messages? cc @alexis-via @sebastienbeau @legalsylvain

legalsylvain (changes requested):
HI.

Thanks @rvalyi for the ping.

> * What is doing with words (not with code).
> * Why it improves the usability.

:+1: Don't understand what are the improvement.

> Also, that can easily be added as an extra module in the case one would really want to and knows exactly what he is doing performance wise.

:+1: to add it in a separate module. 

inheriting from ``mail.thread`` should not be done on odoo core model specially in existing OCA modules IMO.
If desired, please put the feature in a separate and explicite module. ``account_move_line_mail_thread``.

adding such dependency will impact performance (don't know how many), starting with the new index created on the table, that should be maintained and have a cost.

https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_thread.py#L112

pedrobaeza:
Sorry, @rvalyi, I wrote bad. I wanted to say that I **DON'T** want thread in aml as well.

mileo:
Hi Guys,

The account_payment_order module has some usability issues, this issues can occur in all the versions, however, they could occur a little differently in version 13/14 due to the removal of the invoice. 

I will describe what happens in version 12.0 and then we can define how to solve this problem in other versions.

The problems:

    1. When canceling an invoice that one of its accounting entries is present in a payment.order, the user receives a nonsense error.

image

image

The error is the same, regardless of the status of the payment.order.

    2. When accessing an account.move.line contained in a payment order, it is not possible to find which payment order it belongs to, regardless of the status of the payment order.

image

**With the improvements added to this PR, flows are improved as follows:**

    1. When canceling an invoice that one of your accounting entries is contained in a payment order in draft, the payment.line is automatically removed;
    2. When canceling an invoice that one of its accounting entries is contained in a payment order not in draft, the user receives the following notice:

image


    3. A tab was created in aml, to list the payment order lines that the record is associated with.

image

If you try to add again:

image

If you click on cancel, the line is removed

image

If you try to cancel:

image

The invoice thead keeps the same:

image

It is worth mentioning that the idea and placing the email thread in the AML was to keep these messages associated with the AML, however we did not consider the performance issue.

Another point of attention is that some methods have been moved from the invoice to accout.move.line, as they are the responsibility of aml and allow other modules to better manage more complex flows.

Thanks for the review and sorry I didn't explain the PR objectives in advance.

pedrobaeza:
@mileo thanks for the extra explanations. Now we get more the purpose of this, but I'm afraid it's a mistake to automatically remove an account.move.line from an existing payment order, as the order may be already uploaded to banks or other subsequent things happened, as the counterpart generation and reconciliation.

I'm OK raising a more informative error instead of the FK constraint error though.

About refactoring the code to be on account.move.line, I prefer to not do that, as this has the risk of introducing bugs with not too much real gain. I don't think reducing the dot access supposes too much performance gain, or have you measured it?

mileo:
Hi, @pedrobaeza 

Thanks for the review 

> @mileo thanks for the extra explanations. Now we get more the purpose of this, but I'm afraid it's a mistake to automatically remove an account.move.line from an existing payment order, as the order may be already uploaded to banks or other subsequent things happened, as the counterpart generation and reconciliation.

The automatic removal only occurs when the state of the payment order is draft, on the other states user get the UserError.

> 
> I'm OK raising a more informative error instead of the FK constraint error though.
> 
> About refactoring the code to be on account.move.line, I prefer to not do that, as this has the risk of introducing bugs with not too much real gain. I don't think reducing the dot access supposes too much performance gain, or have you measured it?

I did not take measurements, but this can be done if you think it is necessary.

The refactoring of the reduction with the reduction of access to points and the consequent responsibility of aml to make decisions is important for the Brazilian location, but can be apply to other modules too.

In Brazil we have a type of payment order that has a future due date and can be changed after it has been sent to the bank, for several reasons, **to do that we need to send payment.line back to the bank with some modifications.**

Some examples:

1.  The cancel of the already sent invoice, consequently the cancellation of the payment order. To do that we need to create a simular payment.line with cancel action;
2. The user/partner request the change the due date / discount / amount / payment instructions / etc, to do that we need to create a simular payment.line with change action;
3. The user request the write-off of it by financial agreement, to do that we need to create a simular payment.line with Write-off action;
4. After the due date, payment can be paid through another payment method, consequently it must be canceled, however with an action that was paid with other method.
5. It is also possible to send an "aml" manual, without an invoice, consolidating multiple late payments, already canceled or blocked for payment.

I can stay all day writing here, Brazil is absurdly complex....

In summary, it is very important that we can delegate decisions to the AML, because all of these complex flows happen there. Increasing control over the re-sending of aml due to the change and allowing the simplified creation of payment.lines.

There is another use case that can be beneficial for other localizations is the association of this functionality with, for example, the **account_due_list** module, where it would be possible to select one or multiple and add them to a payment order easily.

pedrobaeza:
Even if removing them only on draft payment orders for me is not correct, as one can assume the invoices you put on an order stay the same, and confirm the order thinking that, finding later that an invoice was removed. To raise the informative error is the only solution for me.

About the refactoring of the code, we need real improvement for doing it in such relative old version for not compromising stability without a real reason.

mileo:
> Even if removing them only on draft payment orders for me is not correct, as one can assume the invoices you put on an order stay the same, and confirm the order thinking that, finding later that an invoice was removed. To raise the informative error is the only solution for me.
> 
> About the refactoring of the code, we need real improvement for doing it in such relative old version for not compromising stability without a real reason.

Fine, i will split this pull request in small ones and port it to the 13/14

PR #784, opened 2021-02-12
[13.0][IMP] account_payment_order: done status button

AaronHForgeFlow:
Honestly, I am just starting to use this module. However I don't see a way to set the payment order to state "Done". I really don't know why this status was included and how possibly the payment order can reach  than status without that button. This is more a "question" than a Pull request, just if someone can clarify this I will close it. Regards.

### DIFF
--- a/account_payment_order/views/account_payment_order.xml
+++ b/account_payment_order/views/account_payment_order.xml
@@ -49,8 +49,14 @@
                     <button
                         name="action_done_cancel"
                         type="object"
-                        states="uploaded"
+                        states="uploaded,done"
                         string="Cancel Payments"
+                    />
+                    <button
+                        name="action_done"
+                        type="object"
+                        states="uploaded"
+                        string="Set Done"
                     />
                     <field name="state" widget="statusbar" />
                 </header>


luc-demeyer:
@AaronHForgeFlow 
imho the payment order should move to done when
1) in case of transfer account, this account account is reconciled by bank statement processing.
2) in case of no transfer account all account.move.lines have been reconciled or partially reconciled.
The second case is technically a bit harder to implement.
At interim the unused 'done' state is very confusing for the end users hence I think we should merge this PR.
Even when we have implemented the automatic 'Set Done' logic having a button to allow an authorized user to force this state is usefull.

luc-demeyer (approved):

pedrobaeza:
I would remove the state instead.

hbrunn (approved):
removing the state would make some of my users quite unhappy, I'm currently working on implementing what @luc-demeyer proposed

HaraldPanten (approved):
I'd remove the button as well, but I have nothing against applying the change that @AaronHForgeFlow is suggesting. 👍

pedrobaeza:
OK, please move it to v14 as well for consistency:

/ocabot merge minor

PR #785, opened 2021-02-17
[14.0][FIX] account_banking_sepa_direct_debit: Fix crash caused by typo in print_report_name

alexis-via:
And one small improvement: add print button on mandate form view.

### DIFF
--- a/account_banking_sepa_direct_debit/views/account_banking_mandate_view.xml
+++ b/account_banking_sepa_direct_debit/views/account_banking_mandate_view.xml
@@ -12,6 +12,9 @@
         <field name="model">account.banking.mandate</field>
         <field name="inherit_id" ref="account_banking_mandate.view_mandate_form" />
         <field name="arch" type="xml">
+            <button name="validate" position="before">
+                <button name="print_report" type="object" string="Print" />
+            </button>
             <field name="partner_id" position="after">
                 <field
                     name="scheme"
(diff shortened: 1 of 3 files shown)


account_banking_sepa_direct_debit/views/account_banking_mandate_view.xml
```diff
@@ -12,6 +12,15 @@
         <field name="model">account.banking.mandate</field>
         <field name="inherit_id" ref="account_banking_mandate.view_mandate_form" />
         <field name="arch" type="xml">
+            <button name="validate" position="before">
+                <button
+                    name="print_report"
+                    type="object"
+                    string="Print"
+                    class="btn-default"
+                    states="draft,valid"
```
pedrobaeza: Once valid, it should be shown IMO, or at least not as highlighted.

alexis-via: ok, I agree. I removed states=""

pedrobaeza: Agh, I made a typo, I wanted to say `shouldn't be shown`, but well if you put it without states, but not highlighted, it's OK.

pedrobaeza: Please remove then `class="btn-default"`

alexis-via: btn-default is NOT highlighted (it's highligthed when you have btn-primary).

pedrobaeza: Ah, OK, sorry for the confusion.

pedrobaeza (approved):

HaraldPanten (approved):

PR #788, opened 2021-03-05
[14.0] account_banking_mandate: fix inherit of partner view

alexis-via:
On the inherit of partner view: the xpath uses button[@name='%(base.action_res_partner_bank_account_form) which is located in account.view_partner_property_form and NOT in account.partner_view_buttons on v14.


alexis-via:
cf https://github.com/odoo/odoo/blob/14.0/addons/account/views/partner_view.xml#L163

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it as straight-forward:

/ocabot merge minor

PR #793, opened 2021-03-15
[13.0][FIX] account_payment_order_return: Fix tests

victoralmau:
It's in progress.

Please @joao-p-marques  can you review it?

@Tecnativa TT28774

### DIFF
--- a/account_payment_order_return/tests/test_account_payment_order_return.py
+++ b/account_payment_order_return/tests/test_account_payment_order_return.py
@@ -11,42 +11,38 @@ from odoo.tests.common import Form
 class TestAccountPaymentOrderReturn(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
-        super(TestAccountPaymentOrderReturn, cls).setUpClass()
-        cls.account_type = cls.env["account.account.type"].create(
+        super().setUpClass()
+        cls.a_income = cls.env["account.account"].create(
             {
-                "name": "Test Account Type",
-                "type": "receivable",
-                "internal_group": "income",
-            }
-        )
-        cls.a_receivable = cls.env["account.account"].create(
-            {
-                "code": "TAA",
-                "name": "Test Receivable Account",
-                "reconcile": True,
-                "internal_type": "receivable",
-                "user_type_id": cls.account_type.id,
+                "code": "TIA",
+                "name": "Test Income Account",
+                "user_type_id": cls.env.ref("account.data_account_type_revenue").id,
             }
         )
         cls.partner = cls.env["res.partner"].create({"name": "Test Partner 2"})
-        cls.journal = cls.env["account.journal"].create(
-            {"name": "Test Journal", "type": "bank", "code": "Test"}
+        cls.sale_journal = cls.env["account.journal"].create(
+            {"name": "Test Sale Journal", "type": "sale", "code": "Test"}
+        )
+        cls.bank_journal = cls.env["account.journal"].create(
+            {"name": "Test Bank Journal", "type": "bank"}
         )
         move_form = Form(
-            cls.env["account.move"].with_context(default_type="out_invoice")
+            cls.env["account.move"].with_context(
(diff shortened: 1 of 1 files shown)


joao-p-marques:
Some fixes added to the original commit by @victoralmau 

The tests now pass, but I had to force the value `True` to the `returned_payment` field in the invoices, which should be done automatically when creating a `payment.return`. I think it is a problem with the reconciliation done right before that operation. 

WDYT @pedrobaeza?

pedrobaeza:
Something is not OK if the flag is not auto-changed.

joao-p-marques (approved):

joao-p-marques:
@pedrobaeza as pointed with @victoralmau's latest change, the problem was with the registration of the payment, so it is now fixed by using a Form directly to create an `account.payment`

Thanks :+1:

account_payment_order_return/tests/test_account_payment_order_return.py
```diff
@@ -83,42 +83,27 @@ def test_global(self):
...
-                            "amount": self.payment_line.credit,
-                        },
-                    )
-                ],
-            }
-        )
+        payment_return_form = Form(self.env["payment.return"])
+        payment_return_form.journal_id = self.bank_journal
+        with payment_return_form.line_ids.new() as line_form:
+            line_form.move_line_ids.clear()
```
pedrobaeza: Why do you add this? It's a new record, so no need to clear empty lines

account_payment_order_return/tests/test_account_payment_order_return.py
```diff
@@ -83,42 +83,27 @@ def test_global(self):
...
+                self.payment.move_line_ids.filtered(
+                    lambda x: x.account_id.internal_type == "receivable"
+                )
+            )
+        self.payment_return = payment_return_form.save()
         self.payment_return.action_confirm()
         wizard.include_returned = False
         wizard.populate()
-        self.assertTrue((wizard.move_line_ids - prev_move_lines), 1)
+        self.assertEquals(len(wizard.move_line_ids), 0)
```
pedrobaeza, line 108: This is not resilient, as there can be other items, so please use the same formula as previously (you can add a comment in code for avoiding next time to forget why this is done this way).

pedrobaeza (changes requested):

account_payment_order_return/tests/test_account_payment_order_return.py
joao-p-marques: @pedrobaeza it is true. However, the problem previously became from the fact that the wizard was aggregating data from existing journals and partners. Now that we create those in the tests, there shouldn't be such a big issue. 
However, it is still better to make it more resilient.

victoralmau: Finally it's not necessary to check if there can be other items because is used journals (bank and sales) created previously and it's not possible to exists other records, right?

pedrobaeza: OK, it's very rare that there's something that alters the behavior, although not impossible: imagine a previously installed module creates several journal items per payment on that newly created journal. Anyway, let's keep it this way now.

pedrobaeza (approved):

joao-p-marques:
@pedrobaeza seems like the merge command did not work. Is @OCA-git-bot working?

pedrobaeza:
@sbidoul it seems the bot is stuck again.

sbidoul:
It seems the nightly jobs are taking longer than usual, and progressing very slowly. Not sure why yet.

pedrobaeza:
Is there the possibility of having 2 queues or something for working independently?

sbidoul:
Yes we already have 2 queues, but we'd need to play with priorities.

sbidoul:
I gave it a little boost. It should catch up in an hour or so.

pedrobaeza:
Thanks for the look, Stéphane

PR #796, opened 2021-03-25
[13.0][FIX] account_payment_partner + account_payment_sale: Fix multi-company issues

joao-p-marques:
We were always propagating the payment_mode from SO to Invoice, but that can introduce issues on multi-company environments.
We should check that before propagating.

@Tecnativa
TT28906

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_payment_partner/models/res_partner.py
+++ b/account_payment_partner/models/res_partner.py
@@ -11,12 +11,14 @@ class ResPartner(models.Model):
     supplier_payment_mode_id = fields.Many2one(
         "account.payment.mode",
         company_dependent=True,
+        check_company=True,
         domain="[('payment_type', '=', 'outbound')]",
         help="Select the default payment mode for this supplier.",
     )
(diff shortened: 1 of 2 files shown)


joao-p-marques:
This fixes integration tests. 
Would it also make sense to add a constraint in `account_payment_partner` to avoid associating a payment mode to the partner with a different company in the first place?

pedrobaeza:
This is trying to ignore problems derived from a bad setup, but this should be:

- Prevented in first place in the model (there's an option `check_company=True` for that).
- Do the proper data setup in the failing test.

joao-p-marques:
Thanks for the tips @pedrobaeza 

I think the latest change would fix it then, along with https://github.com/OCA/sale-workflow/pull/1526

Can you check please?

victoralmau (approved):

pedrobaeza (approved):

PR #797, opened 2021-03-29
[13.0][FIX] account_payment_order: Fix tests

joao-p-marques:
On multi-company settings, we need to make sure we are searching and creating objects with the correct company associated.
(Includes part of https://github.com/OCA/bank-payment/pull/786)

Create account and bank journal instead of searching.
Add proper expected taxes to that account to allow predicting the final value.

@Tecnativa
TT28423
TT28962

ping @victoralmau @pedrobaeza

### DIFF
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -14,25 +14,30 @@ class TestPaymentOrderInboundBase(SavepointCase):
     def setUpClass(cls):
         self = cls
         super().setUpClass()
+        self.env.user.company_id = self.env.ref("base.main_company").id
         self.inbound_mode = self.env.ref(
             "account_payment_mode.payment_mode_inbound_dd1"
         )
-        self.invoice_line_account = self.env["account.account"].search(
-            [
-                (
-                    "user_type_id",
-                    "=",
-                    self.env.ref("account.data_account_type_revenue").id,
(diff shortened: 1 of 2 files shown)


account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -14,25 +14,30 @@ class TestPaymentOrderInboundBase(SavepointCase):
...
+                "user_type_id": self.env.ref("account.data_account_type_revenue").id,
+                "tax_ids": [
+                    (4, self.env.ref("l10n_generic_coa.1_sale_tax_template").id)
+                ],
+            }
         )
-        self.journal = self.env["account.journal"].search(
-            [("type", "=", "bank")], limit=1
+        self.journal = self.env["account.journal"].create(
+            {"name": "Test journal", "type": "bank"}
```
victoralmau:
```suggestion
            {"name": "Test journal", "code": "BANK", "type": "bank"}
```

victoralmau (changes requested):

account_payment_order/tests/test_payment_order_inbound.py
joao-p-marques: Ok, change made, but why is this necessary? I see that it automatically fills that field of not defined.

victoralmau (approved):

PR #798, opened 2021-03-30
[13.0][FIX] account_payment_order: Fix tests

joao-p-marques:
Create tax instead of referencing generic one (disappears after installing l10n module)

Missing from #797 

ping @victoralmau @pedrobaeza

### DIFF
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -23,9 +23,6 @@ class TestPaymentOrderInboundBase(SavepointCase):
                 "name": "Test account",
                 "code": "TEST1",
                 "user_type_id": self.env.ref("account.data_account_type_revenue").id,
-                "tax_ids": [
-                    (4, self.env.ref("l10n_generic_coa.1_sale_tax_template").id)
-                ],
             }
         )
         self.journal = self.env["account.journal"].create(
@@ -67,6 +64,7 @@ class TestPaymentOrderInboundBase(SavepointCase):
                 invoice_line_form.quantity = 1
                 invoice_line_form.price_unit = 100.0
                 invoice_line_form.account_id = self.invoice_line_account
+                invoice_line_form.tax_ids.clear()
         invoice = invoice_form.save()
         invoice_form = Form(invoice)
         invoice_form.payment_mode_id = self.inbound_mode


victoralmau (approved):

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -18,14 +18,15 @@ def setUpClass(cls):
...
+        )
         self.invoice_line_account = self.env["account.account"].create(
             {
                 "name": "Test account",
                 "code": "TEST1",
                 "user_type_id": self.env.ref("account.data_account_type_revenue").id,
-                "tax_ids": [
-                    (4, self.env.ref("l10n_generic_coa.1_sale_tax_template").id)
-                ],
+                "tax_ids": [(4, taxes.id)],
```
pedrobaeza: I don't really see the usage of this, so I think you can simply remove the tax.

joao-p-marques: The problem is that, when you install a different `l10n` module, it applies a different tax by default, resulting in different total amounts, as we had in https://github.com/OCA/account-reconcile/pull/370

So, either we read the total amount there instead of hardcoding it, but that is a little but against the nature of the tests (https://github.com/OCA/account-reconcile/pull/370#discussion_r602435371) or we ensure we assign the correct tax here, to always result in the same value.

joao-p-marques: This tests are used in the base for others, BTW, so that's why we don't see the usage here, but there: https://github.com/OCA/account-reconcile/pull/370

pedrobaeza: But associating a tax to an account is something very weird and not needed for the test. Just change the amount in the other test.

joao-p-marques: The problem is that that total amount will be different whether a localization module is installed or not: if l10n_es is installed, for example, no taxes will be applied, and the total amount there would be 200, but if it isn't, the default of 15% tax will be applied and the total value will be 230, and I also haven't tested with other localization modules, so the value might vary again. 
The only solution if we remove the taxes here would be to use a not hardcoded value there, as I had in the 1st version of https://github.com/OCA/account-reconcile/pull/370

pedrobaeza: No if you remove this line AFAIK. We have to always hardcode values in tests. I don't get the problem...

joao-p-marques:
Changes made @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Check if needed in v14

joao-p-marques:
> Check if needed in v14

I opened a PR applying the same logic in v14. When `account_reconcile_payment_order` is migrated, https://github.com/OCA/account-reconcile/pull/370 will be included, as no one seems to have started the migration yet.

PR #801, opened 2021-04-07
[13.0][IMP] account_payment_order: add the posibility to choose the payment…

ValentinVinagre:
… date when adding a transaction line manually and it hasn't a due date.

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -85,6 +85,7 @@ class AccountMoveLine(models.Model):
             "communication_type": communication_type,
             "currency_id": currency_id,
             "amount_currency": amount_currency,
+            "date": False,
             # date is set when the user confirms the payment order
         }
         return vals
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -295,7 +295,7 @@ class AccountPaymentOrder(models.Model):
                 payline.draft2open_payment_line_check()
                 # Compute requested payment date
                 if order.date_prefered == "due":
-                    requested_date = payline.ml_maturity_date or today
+                    requested_date = payline.ml_maturity_date or payline.date or today
                 elif order.date_prefered == "fixed":
                     requested_date = order.date_scheduled or today
                 else:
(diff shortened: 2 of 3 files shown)


ValentinVinagre:
@HaraldPanten could you review it?

HaraldPanten (approved):
Tested in local environment

ValentinVinagre:
Could you review? @OCA/banking-maintainers

manuelcalerosolis (approved):

account_payment_order/models/account_move_line.py
```diff
@@ -85,6 +85,7 @@ def _prepare_payment_line_vals(self, payment_order):
             "communication_type": communication_type,
             "currency_id": currency_id,
             "amount_currency": amount_currency,
+            "date": False,
```
pedrobaeza, line 88: Why forcing this to False?

ValentinVinagre: If I'm not mistaken, because it was possible that I put today's date

pedrobaeza: I see that is a related field: https://github.com/OCA/bank-payment/blob/00031b13ff311faed526df86c5fdd132d922b294/account_payment_order/models/bank_payment_line.py#L65, so all of this seems weird.

ValentinVinagre: And if in the future a default is added in the field, wouldn't it be better to set the date to false?
It should work if it is removed, do I remove it?

pedrobaeza: But the question is not the default or not, but the related part.

ValentinVinagre: In the original version:
`# date is set when the user confirms the payment order` 
I think it is better to set the field False, just in case.

pedrobaeza: But if this field is related, setting to anything is trickling the system.

ValentinVinagre: I think the field you indicated before is not correct. The correct field is this:
https://github.com/OCA/bank-payment/blob/00031b13ff311faed526df86c5fdd132d922b294/account_payment_order/models/account_payment_line.py#L62

pedrobaeza: Oh, right. Sorry for the confusion.

pedrobaeza:
/ocabot merge patch

Should be fw-port it to v14

HaraldPanten:
> /ocabot merge patch
> 
> Should be fw-port it to v14

Will you do that? Or do you prefer us to open a new PR?

pedrobaeza:
All yours!

PR #802, opened 2021-04-14
[14.0][FIX] account_payment_order + account_payment_partner + account_payment_order_return: Fix tests

joao-p-marques:
After https://github.com/odoo/odoo/commit/fcaa54939e9a4f0dd5e47cd0ccffe7aa24bd451c, we need to specify the invoice date for vendor bills to be able to post it.
Also, after https://github.com/odoo/odoo/commit/80c28189ae6b10ad17c4de2a781233a8ffe663a6 we can't post an invoice that is already posted.

@victoralmau @pedrobaeza can you review please?

### DIFF
--- a/account_payment_order/tests/test_payment_order_outbound.py
+++ b/account_payment_order/tests/test_payment_order_outbound.py
@@ -4,6 +4,7 @@
 
 from datetime import date, datetime, timedelta
 
+from odoo import fields
 from odoo.exceptions import UserError, ValidationError
 from odoo.tests.common import TransactionCase
 
@@ -48,6 +49,7 @@ class TestPaymentOrderOutbound(TransactionCase):
                 "payment_mode_id": self.env.ref(
                     "account_payment_mode.payment_mode_outbound_ct1"
                 ).id,
+                "invoice_date": fields.Date.today(),
                 "invoice_line_ids": [
                     (
                         0,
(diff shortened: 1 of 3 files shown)


victoralmau (approved):

victoralmau:
Good job @joao-p-marques.
Please, check `account_payment_order_return` tests and add commit to fix it.

joao-p-marques:
Thanks!
It is not needed in `account_payment_order_return`, as the invoices created there are customer invoices, so it isn't affected by this change.

joao-p-marques:
@victoralmau it should be fixed now, thanks!

victoralmau (approved):

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -48,6 +48,7 @@ def _create_supplier_invoice(self):
                 "payment_mode_id": self.env.ref(
                     "account_payment_mode.payment_mode_outbound_ct1"
                 ).id,
+                "invoice_date": datetime.today(),
```
pedrobaeza: It's better to have a specific date. Why do you want to have it today? In any case, you should use `fields.Date.today()` instead.

joao-p-marques: Ok, thanks @pedrobaeza 
I changed it to use `fields.Date.today()` instead. I opted for that as that was the default date applied by Odoo before its removal: https://github.com/odoo/odoo/blob/fcaa54939e9a4f0dd5e47cd0ccffe7aa24bd451c/addons/account/models/account_move.py#L111
So, for minimal impact in the actual result, I think it would be better to use that.

PR #803, opened 2021-04-14
[FIX][13.0] account_payment_order_return frezze time to run properly …

i-vyshnevska:
…date depandant test

### DIFF
--- a/account_payment_order_return/tests/test_account_payment_order_return.py
+++ b/account_payment_order_return/tests/test_account_payment_order_return.py
@@ -3,6 +3,8 @@
 # Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0
 
+from freezegun import freeze_time
+
 from odoo import fields
 from odoo.tests import common
 from odoo.tests.common import Form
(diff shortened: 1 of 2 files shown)


pedrobaeza:
As I have commented other times, I prefer to not introduce new requirements (even in tests) for a module already in the repo. This will break CIs that don't have this requirement set. Please set instead the value in the corresponding field to a fixed value for avoiding the problem, that I suppose comes from the combination of timezones + test running around midnight.

i-vyshnevska:
> As I have commented other times, I prefer to not introduce new requirements (even in tests) for a module already in the repo. This will break CIs that don't have this requirement set. Please set instead the value in the corresponding field to a fixed value for avoiding the problem, that I suppose comes from the combination of timezones + test running around midnight.

freezegun was introduced in odoo14 as a magor way to handle dates in tests

pedrobaeza:
But this is 13.0

i-vyshnevska:
@pedrobaeza sorry but IMO it doesn't worth fixing tests through haking the timezone when you have a ready tool that is standard even for the newest version, that is not considered with our time budget

pedrobaeza:
I'm not saying to hack timezone, but to fix the needed field. Anyway, this is not failing unless being around midnight, so you can let it as is if not committing at that hours.

pedrobaeza:
It may be enough to use `fields.Date.context_today(self.env.user)` in the affected field.

pedrobaeza:
I did the needed fix in #925, so closing.

Razz-786:
How can i hack imo

PR #805, opened 2021-04-21
[13.0][IMP] account_payment_order - move date should be bank line date when offsetting_account is bank_account

baimont:
Between Odoo v10.0 and Odoo v13.0, date fields on account.move and account.move.line became readonly. In both versions, date field on account.move.line is a related of the account.move one. This means that when creating our move in v10.0, date from account.move would take value of account.move.line. This is not true anymore with the readonly attributes. This is a way to return to that behavior.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -416,6 +416,7 @@ class AccountPaymentOrder(models.Model):
         return True
 
     def _prepare_move(self, bank_lines=None):
+        move_date = False
         if self.payment_type == "outbound":
             ref = _("Payment order %s") % self.name
         else:
@@ -424,6 +425,8 @@ class AccountPaymentOrder(models.Model):
             ref += " - " + bank_lines.name
         if self.payment_mode_id.offsetting_account == "bank_account":
             journal_id = self.journal_id.id
+            if bank_lines:
+                move_date = bank_lines[0].date
         elif self.payment_mode_id.offsetting_account == "transfer_account":
             journal_id = self.payment_mode_id.transfer_journal_id.id
         vals = {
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please specify steps to reproduce the problem.

sanrav:
Steps to test :

* Configure a bank journal with a bank account

image

* Configure a payment mode "SEPA Credit Transfer to suppliers" with Offsetting account = Bank account

image

image

* Create a vendor bill and post it

image

* Create a payment order with Due date as Payment Execution Date Type

image

image

* Confirm payments, generate xml file and confirm the upload => journal entry is created with accounting date = payment date

image

sanrav (approved):
Functional testing OK

pedrobaeza:
I think the problem is that you are not putting on "Move Option", "One move per payment date", and that's the problem. This PR is not correct, assigning randomly the date of the first line.

sanrav:
I tested also with "One move per payment date" and the issue is the same. The accounting entry takes the today's date instead of payment date. Tests done on runbot OCA V13.

If you choose "One move per payment line", there is one accounting entry per payment line. So, one payment date.
If you choose "One move per payment date", we can have more than one payment line grouped into one accounting entry. But as they are grouped by payment date, we will have the same payment date on each line. So, I don't see the issue to choose randomly the date of the first line.

pedrobaeza:
@carlosdauden didn't you add such option of propagating the date?

carlosdauden:
I have commented this problem sometime, but since our clients did not open any issue (most of our clients use offsetting_account == "transfer_account" or they have a fix date of payment) I have not found the time to do it.

carlosdauden (approved):

pedrobaeza (approved):

pedrobaeza:
Thanks for the confirmation and the contribution:

/ocabot merge patch

Can you fw-port it to v14?

PR #806, opened 2021-04-23
[13.0][IMP] account_payment_order - allow inheritance of requested_date computation when confirming a payment order

baimont:
is necessary to https://github.com/OCA/account-payment/pull/410

### DIFF
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -154,3 +154,18 @@ class AccountPaymentLine(models.Model):
             raise UserError(
                 _("Missing Partner Bank Account on payment line %s") % self.name
             )
+
+    def get_requested_date(self, default_date):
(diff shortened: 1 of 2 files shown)


account_payment_order/models/account_payment_line.py
```diff
@@ -154,3 +154,18 @@ def draft2open_payment_line_check(self):
             raise UserError(
                 _("Missing Partner Bank Account on payment line %s") % self.name
             )
+
+    def get_requested_date(self, today):
```
apineux: If it should always be computed with today therefore the today should be in the method and not a parameter.
If you put a parameter to be more generic therefore you should call this parameter date, default_date, ref_date or something more generic than today.

manuelcalerosolis (approved):

adrienpeiffer (approved):

PR #807, opened 2021-04-23
[13.0][FIX] account_payment_purchase: Do not warn if no purchase

AaronHForgeFlow:
Forward port of https://github.com/OCA/bank-payment/pull/652

If the user set no purchase order on an invoice, it makes no sense
to warn them about details being different.

The onchange for the purchase_id field is not only run when the user
sets a purchase, but also when they open a new, empty form view.
Normally this causes no problems, but in some contexts the user
may want to define defaults for the payment term or bank account.
If these are defined, they will not be False on a new draft invoice,
but there will be no purchase, so the code interpreted the values
as having "changed".

We avoid this by simply not checking for changes if no purchase has
been set on the invoice.

cc @ForgeFlow @MiquelRForgeFlow

### DIFF
--- a/account_payment_purchase/models/account_invoice.py
+++ b/account_payment_purchase/models/account_invoice.py
@@ -10,6 +10,9 @@ class AccountMove(models.Model):
 
     @api.onchange("purchase_vendor_bill_id", "purchase_id")
     def _onchange_purchase_auto_complete(self):
+        if not self.purchase_id:
+            # User did not add a purchase order, no need to warn
+            return super()._onchange_purchase_auto_complete()
         new_mode = self.purchase_id.payment_mode_id.id or False
         new_bank = self.purchase_id.supplier_partner_bank_id.id or False
         res = super()._onchange_purchase_auto_complete() or {}


AaronHForgeFlow:
Put your hands up if you want forward port to 14.0

account_payment_purchase/models/account_invoice.py
```diff
@@ -10,9 +10,13 @@ class AccountMove(models.Model):
 
     @api.onchange("purchase_vendor_bill_id", "purchase_id")
     def _onchange_purchase_auto_complete(self):
+        new_purchase = self.purchase_id
         new_mode = self.purchase_id.payment_mode_id.id or False
         new_bank = self.purchase_id.supplier_partner_bank_id.id or False
         res = super()._onchange_purchase_auto_complete() or {}
+        if not new_purchase:
```
pedrobaeza: Why not directly?:
```suggestion
        if not self.purchase_id:
```

pedrobaeza (commented):
~~Anyway, I don't get the use case where this happens~~ Got it

account_payment_purchase/models/account_invoice.py
AaronHForgeFlow: because the _onchange_purchase_auto_complete method sets self.purchase_id to False

pedrobaeza: And do it before at the beginning

pedrobaeza: OK, and can't be done at the beginning, returning super in that case?

AaronHForgeFlow:
> Anyway, I don't get the use case where this happens

It happens when creating the vendor bill manually from the PO. The payment_mode_id is always blank when you create a manual vendor bill from the PO, because self.purchase_id is deleted in the onchange method.

pedrobaeza:
> It happens when creating the vendor bill manually from the PO. The payment_mode_id is always blank when you create a manual vendor bill from the PO, because self.purchase_id is deleted in the onchange method.

Yes, sorry, I saw it at that very moment.

pedrobaeza (approved):

MiquelRForgeFlow (approved):
LGTM

account_payment_purchase/models/account_invoice.py
AaronHForgeFlow: Yes, I tested and I can do it before the onchange. Just I moment and I will push the changes.

account_payment_purchase/models/account_invoice.py
```diff
@@ -10,6 +10,9 @@ class AccountMove(models.Model):
 
     @api.onchange("purchase_vendor_bill_id", "purchase_id")
     def _onchange_purchase_auto_complete(self):
+        if not self.purchase_id:
+            # User did not add a purchase order, no need to warn
+            return super()._onchange_purchase_auto_complete()
```
MiquelRForgeFlow, line 15:
```suggestion
            return super()._onchange_purchase_auto_complete() or {}
```
🤔

pedrobaeza: No that was needed in this method for adding the warning, but returning directly is not needed, whatever the return is.

AaronHForgeFlow:
I just tested and it is not necessary in version14. The creation of bills from the purchase order will not trigger that method in v14.

pedrobaeza:
Good to know

PR #809, opened 2021-04-28
[13.0] [account_payment_purchase][fix] Only issue a warning message if the PO had a non-blank payment mode or bank.

JordiBForgeFlow:
cc @AaronHForgeFlow @JordiMForgeFlow

### DIFF
--- a/account_payment_purchase/models/account_invoice.py
+++ b/account_payment_purchase/models/account_invoice.py
@@ -16,14 +16,18 @@ class AccountMove(models.Model):
         new_mode = self.purchase_id.payment_mode_id.id or False
         new_bank = self.purchase_id.supplier_partner_bank_id.id or False
         res = super()._onchange_purchase_auto_complete() or {}
-        if self.payment_mode_id and self.payment_mode_id.id != new_mode:
+        if self.payment_mode_id and new_mode and self.payment_mode_id.id != new_mode:
             res["warning"] = {
                 "title": _("Warning"),
                 "message": _("Selected purchase order have different payment mode."),
             }
             return res
         self.payment_mode_id = new_mode
-        if self.invoice_partner_bank_id and self.invoice_partner_bank_id.id != new_bank:
+        if (
+            self.invoice_partner_bank_id
+            and new_bank
+            and self.invoice_partner_bank_id.id != new_bank
+        ):
             res["warning"] = {
                 "title": _("Warning"),
                 "message": _("Selected purchase order have different supplier bank."),
--- a/account_payment_purchase/tests/test_account_payment_purchase.py
+++ b/account_payment_purchase/tests/test_account_payment_purchase.py
@@ -13,14 +13,20 @@ class TestAccountPaymentPurchase(SavepointCase):
         cls.journal = cls.env["account.journal"].create(
             {"name": "Test journal", "code": "TEST", "type": "general"}
         )
+        cls.payment_method_out = cls.env["account.payment.method"].create(
+            {
+                "name": "Test payment method",
+                "code": "test",
+                "payment_type": "outbound",
+                "bank_account_required": True,
+            }
+        )
         cls.payment_mode = cls.env["account.payment.mode"].create(
             {
                 "name": "Test payment mode",
                 "fixed_journal_id": cls.journal.id,
                 "bank_account_link": "variable",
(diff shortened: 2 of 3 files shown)


JordiMForgeFlow (approved):

AaronHForgeFlow (approved):

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
A test should be adapted accordingly.

JordiMForgeFlow:
Adapted the tests:
- Currently in the tests the payment mode was related to an "inbound" payment method, which should actually be an "outbound".
- The last test was failing because the payment mode in the invoice was being set to False. When doing so, the method to compute the invoice_partner_bank_id was triggered. This compute method returns False when the payment mode is related to an "inbound" payment method, and also when the payment mode is False, so the invoice_partner_bank_id was actually being set to False in the _onchange_purchase_auto_complete. As it was False, then the warning was not being triggered.
- The second test was modified because, as now it is using a payment method having bank account required, the invoice partner bank id needed to be set. This is solved by calling the onchange so the supplier bank id is set in the PO.

JordiMForgeFlow:
Adapted also the tests for account_payment_purchase_stock with the same criteria.

JordiBForgeFlow:
@alexis-via @pedrobaeza @sbidoul do you agree with this?

sbidoul:
Sounds reasonable.

pedrobaeza:
I don't get the steps to reproduce the problem, and the changes in tests don't reflect them either.

JordiMForgeFlow:
The changes are aimed at preventing showing the warning to the user when no payment mode or supplier bank was set in the PO.

Due to this change, the last test was failing because in that test the payment mode was set to False (both in purchase and purchase2). Then, when performing the `result = invoice._onchange_purchase_auto_complete()`, the payment mode of the invoice was also being set to False (https://github.com/OCA/bank-payment/blob/13.0/account_payment_purchase/models/account_invoice.py#L25), and this was triggering the method to compute the invoice_supplier_bank_id (https://github.com/OCA/bank-payment/blob/13.0/account_payment_partner/models/account_move.py#L70). You can see from the compute that, as the payment mode was False, the invoice partner bank was also set to False, so no warning was being issued due to the changes.

To fix the test, we need to make sure that the compute of the invoice partner bank could actually return a bank_id, so that we are really comparing two non empty partner bank.

That is why first we need to have a payment mode and also the payment type should be outbound.
Also, IMO, if we are assigning a payment mode for a PO it should be a payment mode related to an outbound payment method.

pedrobaeza:
Thanks for the explanations, @JordiMForgeFlow, but what I ask for is the steps to reproduce the problem, which are the detailed steps I have to put in Odoo UI before applying this patch for having such warning message.

About the changes in tests, triggering manually onchanges is a recipe for the future failure, so if you are modifying them, better to use `Form` and mimic all the things that happen at the UI.

JordiBForgeFlow:
Imagine that you have a number of PO's for which you have not yet defined a
payment method. Then, at the time of receiving the bill you want to
indicate it. Since you had no preference in the payment method in the PO,
adding one in the bill should not raise any error.

By the way, before this change if you had a method in the PO and you set it
to blank in the bill you will get no warning. To me this is less justified!

JordiMForgeFlow:
The tests trigger manually the onchange because the error is catched in the result of the onchange. Perhaps it would be better to use a UserError. What do you think?

pedrobaeza (approved):
I still think that the tests can be done with `Form` for avoiding false positive/negative, but it's a secondary thing in this PR.

JordiMForgeFlow:
@pedrobaeza just out of curiosity, in this scenario where we are triggering the warning as the return of the onchange method, how could it be catched in the tests using the `Form`?

pedrobaeza:
I was not talking about that, but not having to launch manually `onchange` for having consistent data, which you are doing here:

https://github.com/OCA/bank-payment/pull/809/files#diff-ce106a1a1362d1523ddf5f577513202fea2a2945b58f6e214a0ca524b1efd850R84

I don't see in your modifications about catching some warning.

Anyways, it's true that I have checked if with `Form`, when assigning a value, there's a way to catch the returned warning of an onchange. Something to investigate, so if you please :pray:

PR #810, opened 2021-05-04
[13.0][IMP] account_payment_partner: Add refund_payment_mode_id to set in reverse moves.

victoralmau:
Add `refund_payment_mode_id` to assign any payment mode to set auto-compute in reverse moves (`in_refund` and `out_refund`).

@Tecnativa TT29536

### DIFF
--- a/account_payment_partner/__manifest__.py
+++ b/account_payment_partner/__manifest__.py
@@ -1,6 +1,7 @@
 # Copyright 2014 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
 # Copyright 2014 Tecnativa - Pedro M. Baeza
 # Copyright 2018 Tecnativa - Carlos Dauden
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
@@ -9,7 +10,7 @@
     "category": "Banking addons",
     "license": "AGPL-3",
     "summary": "Adds payment mode on partners and invoices",
-    "author": "Akretion, " "Tecnativa, " "Odoo Community Association (OCA)",
+    "author": "Akretion, Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/bank-payment",
     "depends": ["account_payment_mode"],
     "data": [
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -1,5 +1,6 @@
 # Copyright 2014-16 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
 # Copyright 2014 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
@@ -57,14 +58,26 @@ class AccountMove(models.Model):
         for move in self:
             move.payment_mode_id = False
             if move.partner_id:
+                partner = move.with_context(force_company=move.company_id.id).partner_id
                 if move.type == "in_invoice":
-                    move.payment_mode_id = move.with_context(
-                        force_company=move.company_id.id
-                    ).partner_id.supplier_payment_mode_id
+                    move.payment_mode_id = partner.supplier_payment_mode_id
                 elif move.type == "out_invoice":
-                    move.payment_mode_id = move.with_context(
-                        force_company=move.company_id.id
-                    ).partner_id.customer_payment_mode_id
+                    move.payment_mode_id = partner.customer_payment_mode_id
+                elif (
+                    move.type in ["out_refund", "in_refund"] and move.reversed_entry_id
+                ):
+                    move.payment_mode_id = (
+                        move.reversed_entry_id.payment_mode_id.refund_payment_mode_id
+                    )
+                elif not move.reversed_entry_id:
+                    if move.type == "out_refund":
+                        move.payment_mode_id = (
+                            partner.customer_payment_mode_id.refund_payment_mode_id
+                        )
+                    elif move.type == "in_refund":
+                        move.payment_mode_id = (
+                            partner.supplier_payment_mode_id.refund_payment_mode_id
+                        )
 
     @api.depends("partner_id", "payment_mode_id")
     def _compute_invoice_partner_bank(self):
@@ -90,13 +103,9 @@ class AccountMove(models.Model):
                         bank_id = get_bank_id()
             move.invoice_partner_bank_id = bank_id
 
-    # I think copying payment mode from invoice to refund by default
-    # is a good idea because the most common way of "paying" a refund is to
-    # deduct it on the payment of the next invoice (and OCA/bank-payment
-    # allows to have negative payment lines since March 2016)
     def _reverse_move_vals(self, default_values, cancel=True):
         move_vals = super()._reverse_move_vals(default_values, cancel=cancel)
-        move_vals["payment_mode_id"] = self.payment_mode_id.id
+        move_vals["payment_mode_id"] = self.payment_mode_id.refund_payment_mode_id.id
         if self.type == "in_invoice":
             move_vals["invoice_partner_bank_id"] = self.invoice_partner_bank_id.id
         return move_vals
--- a/account_payment_partner/models/account_payment_mode.py
+++ b/account_payment_partner/models/account_payment_mode.py
@@ -1,5 +1,6 @@
 # Copyright 2017 ForgeFlow S.L.
-# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# Copyright 2018 Tecnativa - Carlos Dauden
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
 from odoo import _, api, fields, models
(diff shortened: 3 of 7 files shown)


account_payment_partner/models/account_move.py
```diff
@@ -90,13 +99,8 @@ def get_bank_id():
                         bank_id = get_bank_id()
             move.invoice_partner_bank_id = bank_id
 
-    # I think copying payment mode from invoice to refund by default
-    # is a good idea because the most common way of "paying" a refund is to
-    # deduct it on the payment of the next invoice (and OCA/bank-payment
-    # allows to have negative payment lines since March 2016)
     def _reverse_move_vals(self, default_values, cancel=True):
         move_vals = super()._reverse_move_vals(default_values, cancel=cancel)
-        move_vals["payment_mode_id"] = self.payment_mode_id.id
```
carlosdauden, line 99: I think that we should obtain the refund payment mode from the invoice payment mode.
If we do not assign anything, default refund partner payment mode will be set.

You can add a test in which the invoice has a different payment mode than the default customer payment mode to check what I say.

carlosdauden (changes requested):

account_payment_partner/models/account_move.py
pedrobaeza: There are 2 cases:

- Doing a refund from an invoice. In that case, indeed the payment mode from original invoice should be checked and use the refund payment mode for that one.
- Doing a manual refund or invoicing a sales order which results in a refund. In this case (which is the one I had in mind and the customer user case), the payment mode should be extracted from the partner, and the test should be done in the partner onchange.

carlosdauden: Of course, that's why I make the point here inside the _reverse_move_vals method, which is used in case 1.

account_payment_partner/models/account_move.py
```diff
@@ -65,6 +66,14 @@ def _compute_payment_mode(self):
                     move.payment_mode_id = move.with_context(
                         force_company=move.company_id.id
                     ).partner_id.customer_payment_mode_id
+                elif move.type == "out_refund":
+                    move.payment_mode_id = move.with_context(
+                        force_company=move.company_id.id
+                    ).partner_id.customer_payment_mode_id.refund_payment_mode_id
```
carlosdauden: Another option can be to use `reversed_entry_id` field to evaluate if get payment mode from it or from partner.

carlosdauden:
Why remove _reverse_move_vals changes?

victoralmau:
> Why remove _reverse_move_vals changes?

According to https://github.com/OCA/bank-payment/pull/810#discussion_r625590721

pedrobaeza:
@victoralmau I didn't say you have to remove it, only which behavior you need to follow.

victoralmau:
Sorry, changes done.

account_payment_partner/__manifest__.py
```diff
@@ -1,15 +1,16 @@
 # Copyright 2014 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
 # Copyright 2014 Tecnativa - Pedro M. Baeza
 # Copyright 2018 Tecnativa - Carlos Dauden
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Account Payment Partner",
-    "version": "13.0.1.0.3",
+    "version": "13.0.1.1.0",
```
pedrobaeza: Why changing manually the module version if no migration script is added?

account_payment_partner/models/account_move.py
```diff
@@ -65,6 +66,18 @@ def _compute_payment_mode(self):
                     move.payment_mode_id = move.with_context(
                         force_company=move.company_id.id
                     ).partner_id.customer_payment_mode_id
+                elif (
+                    move.type in ["out_refund", "in_refund"] and move.reversed_entry_id
+                ):
+                    move.payment_mode_id = move.reversed_entry_id.payment_mode_id
```
pedrobaeza: Is this executed when doing the reverse?

account_payment_partner/models/account_payment_mode.py
```diff
@@ -24,6 +25,11 @@ class AccountPaymentMode(models.Model):
     show_bank_account_chars = fields.Integer(
         string="# of digits for customer bank account"
     )
+    refund_payment_mode_id = fields.Many2one(
+        comodel_name="account.payment.mode",
+        domain="[('payment_type', '!=', payment_type)]",
+        string="Refund payment mode",
```
pedrobaeza:
```suggestion
        string="Payment mode for refunds",
        help="This payment mode will be used when doing refunds coming from the current payment mode.",
```

account_payment_partner/tests/test_account_payment_partner.py
```diff
@@ -391,12 +396,25 @@ def test_partner_onchange(self):
         self.assertEqual(
             self.supplier_invoice.invoice_partner_bank_id, self.supplier_bank
         )
+        vals = {"partner_id": self.customer.id, "type": "out_refund"}
+        invoice = self.move_model.new(vals)
+        self.assertEquals(invoice.payment_mode_id, self.supplier_payment_mode)
+        vals = {"partner_id": self.supplier.id, "type": "in_refund"}
+        invoice = self.move_model.new(vals)
+        self.assertEquals(invoice.payment_mode_id, self.customer_payment_mode)
+
```
pedrobaeza: Don't add empty lines. If you want to split things, put a comment and also explain the "section".

pedrobaeza (changes requested):
Exercise a test with the reversing of an invoice of each type.

account_payment_partner/models/account_move.py
victoralmau: Now, some tests check it.

pedrobaeza: This is not correct. We have to take the payment mode for the refunds of the reversed entry payment mode. Anyways, I'm not sure if we have to do this one here, as this is only intended for manual refunds.

account_payment_partner/models/account_move.py
pedrobaeza: This one is still missing to fill the refund with the refund payment mode from the original invoice payment mode.

pedrobaeza (changes requested):

account_payment_partner/models/account_move.py
victoralmau: Exactly, these code is related to manual refunds. 
Exists too in `_reverse_move_vals()` function `payment_mode_id` to keep previous value.

victoralmau:
Thanks @pedrobaeza , changes done.

account_payment_partner/models/account_move.py
pedrobaeza: But that's incorrect and the reason why we are doing all these changes, for avoiding a payment mode whose money direction is the contrary. The comment in the code saying that it's better to keep the same payment mode is also incorrect.

victoralmau: OK, i have changed to `refund_payment_mode_id`. Is correct now right?

account_payment_partner/models/account_move.py
```diff
@@ -65,6 +66,20 @@ def _compute_payment_mode(self):
...
+                        move.reversed_entry_id.payment_mode_id.refund_payment_mode_id
+                    )
+                elif move.type == "out_refund" and not move.reversed_entry_id:
+                    move.payment_mode_id = move.with_context(
+                        force_company=move.company_id.id
+                    ).partner_id.customer_payment_mode_id.refund_payment_mode_id
+                elif move.type == "in_refund" and not move.reversed_entry_id:
+                    move.payment_mode_id = move.with_context(
+                        force_company=move.company_id.id
+                    ).partner_id.supplier_payment_mode_id.refund_payment_mode_id
```
pedrobaeza: You can refactor this code for better reading (not black formatted):
```suggestion
                elif not move.reversed_entry_id:
                    partner = move.with_context(force_company=move.company_id.id).partner_id
                    if move.type == "out_refund"
                        move.payment_mode_id = partner.customer_payment_mode_id.refund_payment_mode_id
                    elif move.type == "in_refund":
                       move.payment_mode_id = partner.supplier_payment_mode_id.refund_payment_mode_id
```

victoralmau:
Thanks @pedrobaeza changes done.

pedrobaeza (approved):

carlosdauden (approved):

account_payment_partner/models/account_move.py
```diff
@@ -57,14 +58,26 @@ def _compute_payment_mode(self):
         for move in self:
             move.payment_mode_id = False
             if move.partner_id:
+                partner = move.with_context(force_company=move.company_id.id).partner_id
                 if move.type == "in_invoice":
-                    move.payment_mode_id = move.with_context(
-                        force_company=move.company_id.id
-                    ).partner_id.supplier_payment_mode_id
+                    move.payment_mode_id = partner.supplier_payment_mode_id
```
carlosdauden, line 63: Why do you remove force_company?

account_payment_partner/models/account_move.py
```diff
@@ -57,14 +58,26 @@ def _compute_payment_mode(self):
...
                 elif move.type == "out_invoice":
-                    move.payment_mode_id = move.with_context(
-                        force_company=move.company_id.id
-                    ).partner_id.customer_payment_mode_id
+                    move.payment_mode_id = partner.customer_payment_mode_id
```
carlosdauden, line 65: Same

carlosdauden (changes requested):

account_payment_partner/models/account_move.py
pedrobaeza: It's already in https://github.com/OCA/bank-payment/pull/810/files#diff-560b5e454acd270e25223d56b85967eaec36fe9a08eac999124990650aa32f6cR61

carlosdauden: Ups, I see, sorry

carlosdauden (approved):

manuelcalerosolis (approved):

pedrobaeza:
/ocabot merge minor

To be fw-ported to v14

PR #816, opened 2021-05-17
[14.0][FIX] account_payment_order: several minor fixes

joao-p-marques:
**Change security group category**

Before this change, the group was creating a new section in each user's settings called "Other" with a field called "Other Extra Rights" that could select "Accounting / Payments" as an option
This places it in the correct location

---

**Make post move field invisible in payment modes**

When the generate_moves option is not selected, the post_move option should not be available

---

@Tecnativa
TT29182

ping @pedrobaeza @victoralmau


HaraldPanten (approved):
Small changes. Seems reasonable

victoralmau (approved):

PR #818, opened 2021-05-21
[9.0][FIX][MIG] sepa: migrate noupdate payment method settings

StefanRijnhart:
To get from https://github.com/OCA/bank-payment/blob/8.0/account_banking_sepa_direct_debit/data/payment_type_sdd.xml#L8
to https://github.com/OCA/bank-payment/blob/9.0/account_banking_sepa_direct_debit/data/account_payment_method.xml#L12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ sudo: false
 cache: pip
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
This is something that I always do manually and each time I say: I have to develop the migration script, but never perform it. Thanks for materializing it.

pedrobaeza:
Fast-tracking being OpenUpgrade related thing and without consuming CI.

PR #821, opened 2021-06-02
[14.0][IMP]account_payment_order: payment date when adding transaction line manually

manuelregidor:
A payment date can be added when a transaction line is introduced manually and it doesn't have a due date.
This imp has already been applied to v13 module (https://github.com/OCA/bank-payment/pull/801)


HaraldPanten (approved):

HaraldPanten:
@pedrobaeza could you have a quick look at this one?

THX!

PR #822, opened 2021-06-08
[14.0] Improve computation of sepa boolean, update move generation, improve views

alexis-via:
Improve computation of sepa on account.payment.order: check IBAN is in SEPA zone
Update move line generation to get transfer account from bank journal
Update payment mode configuration accordingly (3 fields removed)
Several improvements in payment order tree and form view


pedrobaeza (commented):
Why removing the offsetting account + journal? I don't think the journal selected for the payment order serves for the same, as for SEPA orders, you select the bank journal, but you want to register the intermediate move on other journal + account

alexis-via:
@pedrobaeza Well, it's the philosophy of odoo v14!
- The offsetting account is the payment_debit_account_id/payment_credit_account_id of the bank journal
- the offsetting move is also the same bank journal.
This is what happens when you click on "Register Payment" on an invoice. I can understand that few people may prefer to have a different journal for the offsetting move... but most people will be OK with this I think.

pedrobaeza:
Please check Travis and reduce the number of commits.

account_banking_pain_base/views/account_payment_order.xml
```diff
@@ -21,5 +21,19 @@
                 />
             </field>
         </field>
-    </record>
```
NL66278, line 24: I think this shifting of the closing `</record>` is a mistake

account_banking_pain_base/views/account_payment_order.xml
```diff
@@ -21,5 +21,19 @@
...
+</record>
+    <record id="account_payment_order_tree" model="ir.ui.view">
+        <field name="name">pain.base.account.payment.order.tree</field>
+        <field name="model">account.payment.order</field>
+        <field
+            name="inherit_id"
+            ref="account_payment_order.account_payment_order_tree"
+        />
+        <field name="arch" type="xml">
+                <field name="description" position="after">
```
NL66278, line 33: And here indenta are wrong. DId you use tabs instead of spaces?

account_banking_pain_base/views/bank_payment_line_view.xml
```diff
@@ -17,4 +17,15 @@
             </field>
         </field>
     </record>
+    <record id="bank_payment_line_tree" model="ir.ui.view">
+        <field name="name">pain.base.bank.payment.line.tree</field>
+        <field name="model">bank.payment.line</field>
+        <field name="inherit_id" ref="account_payment_order.bank_payment_line_tree" />
+        <field name="arch" type="xml">
+            <field name="communication" position="after">
+                    <field name="priority" optional="hide" />
```
NL66278, line 26: Same here,

NL66278 (approved):
:+1: Functionally seems good to me. Some remarks on xml formatting. And some of our customers are missing the 'done' state that was set in the pas when payments were finally reconciled.

NL66278:
@alexis-via What is the work still being done on this?

alexis-via:
@NL66278 This is the branch I deploy for my v14 customers. There is no special work being done here at the moment, but this branch consolidate all the work I made so far to fix and improve bank-payment for v14.

account_banking_pain_base/views/account_payment_order.xml
alexis-via: As you know, the formatting is done by black/isort and all those automated tools. So, if you're not happy with XML reformatting, you can report this upstream :)

account_banking_pain_base/views/account_payment_order.xml
alexis-via: same answer

account_banking_pain_base/views/bank_payment_line_view.xml
alexis-via: same answer

alexis-via:
BTW, one thing left to do it to fix Travis. I'll try to find time to work on that in the coming days, to merge this PR.

account_banking_pain_base/views/account_payment_order.xml
NL66278: Strange, if I change the formatting to what I would expect, it is perfectly fine with prettier etc. Is not blocking of course, but it just struck me as odd.

alexis-via:
Tests are ok on my laptop... so I hope Travis will be green.

alexis-via:
Travis is green, so I propose to merge this.

ProcessControl-MTG:
Aunque la funcionalidad es similar, no sirve para lo mismo... La cuenta de compensación en muchos casos se utilizaba para las remesas al descuento contra la 431. Odoo no permite indicar esta cuenta en el diario de banco cómo cuenta transitoria...Por lo que el balance sería incorrecto. @alexis-via

rousseldenis:
> @pedrobaeza Well, it's the philosophy of odoo v14!
> 
> * The offsetting account is the payment_debit_account_id/payment_credit_account_id of the bank journal
> * the offsetting move is also the same bank journal.
>   This is what happens when you click on "Register Payment" on an invoice. I can understand that few people may prefer to have a different journal for the offsetting move... but most people will be OK with this I think.

@alexis-via We have a customer that requires a different journal than the bank one.

Could we revert this or find a way to keep both possible ?

PR #824, opened 2021-06-14
[9.0][FIX] account_payment_order: do not mangle states in migration

NL66278:
After a conversion from 7.0 to 12.0 of a customer we noticed that all the done payment orders had reverted to uploaded and all sent payment orders to generated.

In 9.0 for some reason the done state is missing, even though there is an button Done that sets the state to the non-existing done state,

This PR intends to rectify this problem.

### DIFF
--- a/account_payment_order/migrations/9.0.1.0.0/post-migration.py
+++ b/account_payment_order/migrations/9.0.1.0.0/post-migration.py
@@ -17,7 +17,7 @@ def map_payment_type(cr):
     openupgrade.map_values(
         cr,
         openupgrade.get_legacy_name('state'), 'state',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The same problem then will appear on 10.0. It's better to simply adapt to the new states, that are equivalent in fact.

NL66278:
@pedrobaeza I am not sure. In 12.0 the 'done' state is back. Apparently a need is still felt for this state.

pedrobaeza:
Not really, it was introduced randomly and we are willing to remove it:

https://github.com/OCA/bank-payment/issues/740#issuecomment-856960532

PR #826, opened 2021-06-21
[IMP] review sql request for migration fast treatment

jco-subteno-it:


pedrobaeza:
I don't think the change is faster. Subqueries are always slower.

jco-subteno-it:
@pedrobaeza I go from one hour of treatment to 5 minutes on my client migration

pedrobaeza:
The problem then should be a missing index in the involved fields, but subquey shouldn't be the solution.

jco-subteno-it:
i agree

pedrobaeza:
Then better to add the index where proceeding, or in the normal use of the module, the problem will appear when unlinking cascade or similar.

dreispt:
@jco-subteno-it can we close?

jco-subteno-it:
Yes sorry

pedrobaeza:
If the index is something that can be incorporated to the module, just tell it or do a PR adding such thing.

PR #827, opened 2021-06-21
[FIX] Add partner_bank_id validation in AccountMoveLine

danypr92:
We found a validation mistake when we are creating AccountMove to generate the payment order. Steps to reproduce the problem:

1. Create a AccountMove with one AccountMoveLine
2. Save the AccountMove
3. Access to the AccountMoveLine tree view with the smart button `Reconcilied Entries`
4. Open the AccountMoveLine
5. Edit the AccountMoveLine adding a Bank Account (this field only show the bank accounts related with the `partner_id`) and save
6. Edit the AccountMoveLine and change the Partner. Save.

This process allow us to create AccountMoveLines with a `partner_id` and a `partner_bank_id` related to another partner. Is a hack to pass the partner_bank_id domain.

With this validation, we maintain the logic of the domain and avoid this hack.

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -26,7 +26,8 @@ class AccountMoveLine(models.Model):
     )
 
     @api.onchange('partner_id')
-    def empty_partner_bank(self):
+    def _onchange_partner_id(self):
+        """Empty partner bank for avoiding inconsistencies."""
         self.partner_bank_id = None
 
     @api.multi
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -25,6 +25,10 @@ class AccountMoveLine(models.Model):
         string="Payment lines",
     )
 
+    @api.onchange('partner_id')
+    def empty_partner_bank(self):
+        self.partner_bank_id = None
+
     @api.multi
     def _prepare_payment_line_vals(self, payment_order):
         self.ensure_one()


pedrobaeza:
You must clear `partner_bank_id` on the onchange of the partner_id instead.

danypr92:
> You must clear `partner_bank_id` on the onchange of the partner_id instead.

Done! 
This `onchange` fixes the UI interaction, but we want to maintain the validation to force the constraint in the object creation and modification.
In case of we want to modify a AccountMoveLine from another method we want to execute this validation.

pedrobaeza:
> In case of we want to modify a AccountMoveLine from another method we want to execute this validation.

Then your code should be in charge of doing that.

If not, ugly patches will come for these things. I don't see correct these changes.

danypr92:
> > In case of we want to modify a AccountMoveLine from another method we want to execute this validation.
> 
> Then your code should be in charge of doing that.
> 
> If not, ugly patches will come for these things. I don't see correct these changes.

Ok! Done! I remove the first commit (validations) and move it to our custom module.

Keeping in this PR only the `onchange` method to clear the `partner_bank_id` if the `partner_id` is changed.

Thanks, @pedrobaeza, for the suggestions and for the help! 

How we can move this PR forward?

account_payment_order/models/account_move_line.py
```diff
@@ -25,6 +25,10 @@ class AccountMoveLine(models.Model):
         string="Payment lines",
     )
 
+    @api.onchange('partner_id')
+    def empty_partner_bank(self):
```
pedrobaeza: Follow guidelines for names and add docstring:

```suggestion
    def _onchange_partner_id(self):
        """Empty partner bank for avoiding inconsistencies."""
```

danypr92: Done! Thanks!

pedrobaeza (approved):

cesarlr (approved):

PR #828, opened 2021-07-05
[14.0][IMP] account_payment_order:- Improved Bill Reference field layout issue.

murtuzasaleh:
New-Odoo

### DIFF
--- a/account_payment_order/views/account_invoice_view.xml
+++ b/account_payment_order/views/account_invoice_view.xml
@@ -42,7 +42,7 @@
             <field name="payment_mode_id" position="after">
                 <field name="payment_order_ok" invisible="1" />
             </field>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@joao-p-marques didn't you do something similar?

joao-p-marques:
I did, in https://github.com/OCA/bank-payment/pull/817, that should've fixed it. Does this happen when integrating with another module @murtuzasaleh?

murtuzasaleh:
> I did, in #817, that should've fixed it. Does this happen when integrating with another module @murtuzasaleh?

I checked and found that the view is not updated in the database with the latest code, I updated the module and now the layout is working fine.

Thank you for the review @joao-p-marques @pedrobaeza 

I am closing this PR.

PR #832, opened 2021-07-07
[13.0][FIX] account_payment_partner: _compute_invoice_partner_bank isn't executed

carlosdauden:
when partner_id changes.

Odoo sets normal M2O invoice_partner_bank_id field
https://github.com/odoo/odoo/blob/a8d3f466dfffca08214acecf08ec298e3ace6272/addons/account/models/account_move.py#L218-L220
On this onchanges:
https://github.com/odoo/odoo/blob/a8d3f466dfffca08214acecf08ec298e3ace6272/addons/account/models/account_move.py#L364
https://github.com/odoo/odoo/blob/c1e8b5ac2ef4df5851147109a510bb317a747914/addons/purchase/models/account_invoice.py#L77

The account_payment_partner module changes the field
https://github.com/OCA/bank-payment/blob/32b6b88dfdb957e87635faf44f8e8634029382e2/account_payment_partner/models/account_move.py#L29-L34
And adds this method:
https://github.com/OCA/bank-payment/blob/32b6b88dfdb957e87635faf44f8e8634029382e2/account_payment_partner/models/account_move.py#L93-L104

It is computed fine when payment_mode_id changes, but it is not executed when partner_id changes, although it is correctly defined in the depends:
https://github.com/OCA/bank-payment/blob/32b6b88dfdb957e87635faf44f8e8634029382e2/account_payment_partner/models/account_move.py#L82

As the value is set by the onchange, the ORM does not recompute the field (which is optimal in most cases).
After studying several ways, the simplest and most self-explanatory thing is to call the compute method

@Tecnativa TT30875

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -79,6 +79,15 @@ class AccountMove(models.Model):
                             partner.supplier_payment_mode_id.refund_payment_mode_id
                         )
 
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
+        """Force compute because the onchange chain doesn't call
+        ``_compute_invoice_partner_bank``.
+        """
+        res = super()._onchange_partner_id()
+        self._compute_invoice_partner_bank()
+        return res
+
     @api.depends("partner_id", "payment_mode_id")
     def _compute_invoice_partner_bank(self):
         for move in self:
--- a/account_payment_partner/tests/test_account_payment_partner.py
+++ b/account_payment_partner/tests/test_account_payment_partner.py
@@ -172,7 +172,7 @@ class TestAccountPaymentPartner(SavepointCase):
         cls.journal_bank = cls.env["res.partner.bank"].create(
             {
                 "acc_number": "GB95LOYD87430237296288",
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Please check pre-commit (if you install it locally with `pre-commit install -f`, it's automatically run.

account_payment_partner/models/account_move.py
```diff
@@ -79,6 +79,13 @@ def _compute_payment_mode(self):
                             partner.supplier_payment_mode_id.refund_payment_mode_id
                         )
 
+    # Force compute because onachange partner_id don't execute _compute_invoice_partner_bank
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
```
pedrobaeza: This should be put as docstring instead:
```suggestion
    @api.onchange("partner_id")
    def _onchange_partner_id(self):
        """Force compute because the onchange chain doesn't call
        ``_compute_invoice_partner_bank``.
        """
```

pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
/ocabot merge patch

Please forwar-port it to v14

PR #833, opened 2021-07-08
[13.0][FIX] account_payment_partner: _compute_invoice_partner_bank isn't executed

carlosdauden:
when value is set from SO _prepare_invoice method.

This core method set invoice_partner_bank value, so the field is not computed:
https://github.com/odoo/odoo/blob/485c7453316d268029d8b8251e62b35ce74e37b4/addons/sale/models/sale.py#L566

@Tecnativa TT30875

Similar to https://github.com/OCA/bank-payment/pull/832

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -151,3 +151,11 @@ class AccountMove(models.Model):
             )
         # Return this as empty recordset
         return self.invoice_partner_bank_id
+
+    @api.model
+    def create(self, vals):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
This context method is a bit fragile and won't work if launching the sales order invoicing from other place than UI, but I don't have either a better solution without a glue module.

sergio-teruel:
For now it's the only way to do this...

sergio-teruel (approved):

PR #840, opened 2021-07-30
[13.0][FIX] account_payment_partner: pass invoice_partner_bank_id correctly on account move create

JordiMForgeFlow:
We need to pass the ID to the creation values to prevent the following error:

> psycopg2.ProgrammingError: can't adapt type 'res.partner.bank'

cc @pedrobaeza @carlosdauden


pedrobaeza (approved):

pedrobaeza:
Let's fast-track it. Thanks for spotting it:

/ocabot merge patch

Can you do the same to v14?

PR #842, opened 2021-08-03
[FIX] account_payment_partner: Not assign bank.

manuelcalerosolis:
When a bank is not found, it should not be assigned, so as not to change the value that the field brings.

### DIFF
--- a/account_payment_partner/__manifest__.py
+++ b/account_payment_partner/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     "name": "Account Payment Partner",
-    "version": "13.0.1.1.3",
+    "version": "13.0.1.1.4",
     "category": "Banking addons",
     "license": "AGPL-3",
     "summary": "Adds payment mode on partners and invoices",
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -110,7 +110,8 @@ class AccountMove(models.Model):
                         and move.commercial_partner_id.bank_ids
                     ):
                         bank_id = get_bank_id()
-            move.invoice_partner_bank_id = bank_id
+            if bank_id:
+                move.invoice_partner_bank_id = bank_id
 
     def _reverse_move_vals(self, default_values, cancel=True):
         move_vals = super()._reverse_move_vals(default_values, cancel=cancel)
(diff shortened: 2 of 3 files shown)


account_payment_partner/models/account_move.py
```diff
@@ -110,7 +110,8 @@ def get_bank_id():
                         and move.commercial_partner_id.bank_ids
                     ):
                         bank_id = get_bank_id()
-            move.invoice_partner_bank_id = bank_id
+            if bank_id:
+                move.invoice_partner_bank_id = bank_id
```
carlosdauden, line 114: Can you specify the steps of a real case where this condition is needed?

account_payment_partner/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     "name": "Account Payment Partner",
-    "version": "13.0.1.1.3",
+    "version": "13.0.1.1.4",
```
carlosdauden, line 9: Why do you change this manually?

carlosdauden (changes requested):

account_payment_partner/models/account_move.py
manuelcalerosolis: Creating an invoice of type "**out_invoice**" the field "**invoice_partner_bank_id**" have a default value res.partner.bank(47,) by example.

If we don't ask for the value before assigning, it replaces it with False.

Before this changes: 

image

After:

image

The company's default account has been set by default when creating the invoice.

account_payment_partner/__manifest__.py
manuelcalerosolis: I have not changed it, it must be an automatic process.

carlosdauden:
Are you try this options?
https://github.com/OCA/bank-payment/blob/7f6dcbab06c3e3641846efadc7f59ebf9e58b8ab/account_payment_partner/models/account_payment_mode.py#L13-L27
https://github.com/OCA/bank-payment/blob/7f6dcbab06c3e3641846efadc7f59ebf9e58b8ab/account_payment_partner/models/account_move.py#L134-L153

manuelcalerosolis:
> Are you try this options?
> https://github.com/OCA/bank-payment/blob/7f6dcbab06c3e3641846efadc7f59ebf9e58b8ab/account_payment_partner/models/account_payment_mode.py#L13-L27
> 
> 

I don't know what this has to do with not selecting the default account when creating an invoice, these values are for printing.

> https://github.com/OCA/bank-payment/blob/7f6dcbab06c3e3641846efadc7f59ebf9e58b8ab/account_payment_partner/models/account_move.py#L134-L153

This method is only called when the payment method contains a code for "Facturae", I also don't see that it has anything to do with the issue of assigning bank accounts on the invoice.

PR #843, opened 2021-08-05
[13.0][FIX] account_payment_partner: Error when salesperson tries to print invoices

CarlosRoca13:
This error is happening because the salesperson has not access to payment_method_id.

With this changes we avoid the error and the invoices are printed like normally.

cc @Tecnativa TT31296

ping @carlosdauden @pedrobaeza


carlosdauden (approved):

pedrobaeza (approved):

pedrobaeza:
Move it to v14:

/ocabot merge patch

PR #853, opened 2021-09-10
[14.0][FIX] account_payment_partner - Only update partner bank if not set

CasVissers-360ERP:
In some cases partner_bank_id might be set already, i.e. UBL or other imports.

In addition when no payment mode is set this method will remove the bank account, since payment mode is not mandatory this makes no sense to me.

@pedrobaeza

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -45,7 +45,7 @@ class AccountMove(models.Model):
 
     @api.depends("partner_id", "move_type")
     def _compute_partner_bank_filter_type_domain(self):
-        for move in self:
+        for move in self.filtered(lambda m: not m.partner_bank_id):
             if move.move_type in ("out_invoice", "in_refund"):
                 move.partner_bank_filter_type_domain = move.bank_partner_id
             elif move.move_type in ("in_invoice", "out_refund"):
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -45,7 +45,7 @@ class AccountMove(models.Model):
 
     @api.depends("partner_id", "move_type")
     def _compute_partner_bank_filter_type_domain(self):
-        for move in self.filtered(lambda m: not m.partner_bank_id):
+        for move in self:
             if move.move_type in ("out_invoice", "in_refund"):
                 move.partner_bank_filter_type_domain = move.bank_partner_id
             elif move.move_type in ("in_invoice", "out_refund"):
@@ -91,7 +91,7 @@ class AccountMove(models.Model):
 
     @api.depends("partner_id", "payment_mode_id")
     def _compute_partner_bank(self):
-        for move in self:
+        for move in self.filtered(lambda m: not m.partner_bank_id):
             # No bank account assignation is done for out_invoice as this is only
             # needed for printing purposes and it can conflict with
             # SEPA direct debit payments. Current report prints it.


account_payment_partner/models/account_move.py
```diff
@@ -45,7 +45,7 @@ def _compute_payment_mode_filter_type_domain(self):
 
     @api.depends("partner_id", "move_type")
     def _compute_partner_bank_filter_type_domain(self):
-        for move in self:
+        for move in self.filtered(lambda m: not m.partner_bank_id):
```
pedrobaeza: This is not correct, as all fields must have a result or there will be cache miss. You can do `self.partner_bank_filter_type_domain = self.partner_bank_filter_type_domain` for avoiding, but functionally I haven't checked what you want to achieve. Anyways, it sounds weird that you are acting on the domain.

CasVissers-360ERP:
What I want to achieve: 
Base Odoo fills the bank account on a supplier invoice. With this module installed and no payment mode set on a supplier the bank account is (incorrectly) removed:
https://drive.google.com/file/d/1Y_RxADwgOeEOTgOhjC91BNhzjOvXrSeB/view

pedrobaeza:
This is done on purpose: we don't want such bank account, as it's incorrect according bank-payment suite and the automatic retrieval of the bank account when filling the payment order.

CasVissers-360ERP:
>> we don't want such bank account, as it's incorrect according bank-payment suite

I don't understand why it's incorrect, if we don't set a payment mode on a supplier invoice we want to use Odoo's base method to pay the invoice? In that case we should let Odoo base decide what to do? If we want to use the OCA bank-payment suite we set a payment mode?

pedrobaeza:
Both in invoice report and in payment orders, the more suitable bank account is got when needed. If you pre-set it, you are losing flexibility, as it's fixed forever. There are cases where the bank account changes across times and invoices with a long due date points to the incorrect bank account, as it was fixed some time ago. So better to always leave it empty.

CasVissers-360ERP:
Just found out that I applied the fix to the wrong method in this commit. Fixed that now.

>> If you pre-set it, you are losing flexibility, as it's fixed forever. There are cases where the bank account changes across times and invoices with a long due date points to the incorrect bank account, as it was fixed some time ago. So better to always leave it empty.

If the user choses from i.e. a list of 5 bank accounts to use one, we never want to overwrite that? Simply selecting the first bank in the list (like this method does) might possible reset the bank account without even notifying the user.

To be honest, it just makes no sense to me to not set a bank account if available and overwriting it when a different payment mode is selected. Seems error-prone to me.

pedrobaeza:
The problem is that you don't know if it has been selected by you or by the system, so current code is correct. If you still want to select one, just do it after selecting the rest of the params, and no change will be done.

CasVissers-360ERP:
This still doesn't explain why we should not set it if no payment mode is selected?

pedrobaeza:
No setting payment mode doesn't mean that you are not going to use it in payment orders. And as said, in the invoice report, the bank account can be printed dynamically.

CasVissers-360ERP:
Well I will close the PR since we won't agree on this it seems.

For me the use-case is simple and logical. Account payment partner is a dependency oftenly needed for i.e. the UBL modules. It might be that the module is even installed without account_payment_order in the same system. Currently account_payment_partner breaks the base behaviour of Odoo since it removes the bank account from a supplier invoice if no payment mode is set. In my example UBL invoices only need a payment mode to be set on customer invoices (even if not set there it will just use a default one). Thus we are forced to use payment modes, else base Odoo will start acting differently.

PR #855, opened 2021-09-15
[13.0][FIX] account_payment_partner: Set the correct value of payment_mode_id instead of False

victoralmau:
Set the correct value of payment_mode_id instead of False.

@Tecnativa TT31904

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -56,7 +56,9 @@ class AccountMove(models.Model):
     @api.depends("partner_id", "company_id")
     def _compute_payment_mode(self):
         for move in self:
-            move.payment_mode_id = False
+            move.payment_mode_id = move.payment_mode_id
+            if move.company_id and move.payment_mode_id.company_id != move.company_id:
+                move.payment_mode_id = False
             if move.partner_id:
                 partner = move.with_context(force_company=move.company_id.id).partner_id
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

account_payment_partner/models/account_move.py
```diff
@@ -56,7 +56,7 @@ def _compute_partner_bank_filter_type_domain(self):
     @api.depends("partner_id", "company_id")
     def _compute_payment_mode(self):
         for move in self:
-            move.payment_mode_id = False
+            move.payment_mode_id = move.payment_mode_id
```
carlosdauden, line 59: What happens if the company changes?

victoralmau: I think that this change would not affect, but it is important that it is not set to False because if for example we want to define a default value we could not (with this change if it would be possible).

pedrobaeza: You can add following change for getting the desired result on company change:
```suggestion
            if move.payment_mode_id.company_id == move.company_id:
                move.payment_mode_id = move.payment_mode_id
            else:
                move.payment_mode_id = False
```

carlosdauden (approved):

HaraldPanten (approved):

pedrobaeza:
/ocabot merge patch

To v14, please

PR #856, opened 2021-09-17
[14.0][FIX] account_payment_partner: Set the correct value of payment_mode_id instead of False

victoralmau:
FWP from 14.0: https://github.com/OCA/bank-payment/pull/855

Set the correct value of `payment_mode_id` instead of `False`.

@Tecnativa TT31904

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -56,7 +56,9 @@ class AccountMove(models.Model):
     @api.depends("partner_id", "company_id")
     def _compute_payment_mode(self):
         for move in self:
-            move.payment_mode_id = False
+            move.payment_mode_id = move.payment_mode_id
+            if move.company_id and move.payment_mode_id.company_id != move.company_id:
+                move.payment_mode_id = False
             if move.partner_id:
                 partner = move.with_company(move.company_id.id).partner_id
                 if move.move_type == "in_invoice":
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

account_payment_partner/models/account_move.py
```diff
@@ -56,7 +56,8 @@ def _compute_partner_bank_filter_type_domain(self):
     @api.depends("partner_id", "company_id")
     def _compute_payment_mode(self):
         for move in self:
-            move.payment_mode_id = False
+            if move.company_id and move.payment_mode_id.company_id != move.company_id:
+                move.payment_mode_id = False
```
pedrobaeza, line 61: This is missing the other part that assigns `move.payment_mode_id = move.payment_mode_id`.

victoralmau: In 14.0 is not necessary.

pedrobaeza: Is there a test that proves it?

victoralmau: The test `test_account_move_payment_mode_id_default` that is added in this PR proves it.
However, the problem in 13.0 is that in a compute it is necessary to define all the options and always assign a value (or a CacheMiss error appears); on the other hand, in 14.0, if there is any case in which no value is defined, odoo already defines one without showing an error.

pedrobaeza: Are you talking about a default value or the previous one?

victoralmau: Everything is related to default value in computed fields.

pedrobaeza: But then it's needed, because we don't want to switch to the default value, but keep the previous one. There's something that is not fitting me about what you comment... Take into account that this method can be executed multiple times, not only on new record, but it's acting like an onchange.

pedrobaeza (approved):

PR #859, opened 2021-10-05
[12.0][IMP] account_payment_order: show total on account.move.line tree

robinkeunen:
Shows total to the fields balance, amount_residual, amount_residual_currency columns in account.move.line tree view.

Before:
account-move-line-before

After:
account-move-line-after

By the way, I guess there is a reason why this view is updated in the `fields_view_get` function rather than in an xml view but I don't understand it. Can someone explain ?

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -114,8 +114,9 @@ class AccountMoveLine(models.Model):
                     elem = etree.Element(
                         'field', {
                             'name': 'balance',
-                            'readonly': 'True'
-                        })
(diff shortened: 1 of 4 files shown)


robinkeunen:
I fixed the linting errors linked to this module, should I also fix the errors not linked to `account_payment_order` module ?

enricostano:
:+1:

remytms (commented):
LGTM. It can be good if this PR can be merged.

carmenbianca (commented):
LGTM.

Pinging @OCA/banking-maintainers to see if this can be reopened :)

PR #860, opened 2021-10-05
[IMP] account_payment_order: Set orders to done on payment reconciliation

hbrunn:
@luc-demeyer @pedrobaeza this is the close orders on reconciliation PR. But I don't really get what you mean in point 2 of https://github.com/OCA/bank-payment/pull/784#issuecomment-896785175 - isn't this in terms of reconciliation the same, just with a different account?

When all is ironed out here, this should be very simple to port to 13 and 14

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -148,3 +148,37 @@ class AccountMoveLine(models.Model):
             )
             result.update(arch=arch, fields=fields)
         return result
+
+    @api.multi
+    def reconcile(self, writeoff_acc_id=False, writeoff_journal_id=False):
+        """ Set payment orders with fully reconciled lines to done """
+        result = super().reconcile(
+            writeoff_acc_id=writeoff_acc_id,
+            writeoff_journal_id=writeoff_journal_id,
+        )
+        if not self.env.context.get('account_payment_order_defer_close'):
+            self.filtered('full_reconcile_id')._close_payment_orders()
+        return result
+
+    @api.multi
+    def _close_payment_orders(self):
+        """
+        Set payment orders linked to move lines in self to done if all
+        of them are reconciled
+        """
+        for order in self._find_payment_orders():
+            if order.state != 'done' and order._all_lines_reconciled():
+                order.action_done()
+
+    @api.multi
+    def _find_payment_orders(self):
+        """
+        Return all payment orders linked (directly by payment_line_ids
+        or indirectly by reconciliation with a transfer account) to self
+        """
+        return self.mapped(
+            'move_id.line_ids.bank_payment_line_id.order_id'
+        ) | self.mapped(
+            'full_reconcile_id.reconciled_line_ids.move_id.line_ids.'
+            'bank_payment_line_id.order_id'
+        )
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -114,7 +114,7 @@ class AccountPaymentOrder(models.Model):
     @api.multi
     def unlink(self):
         for order in self:
-            if order.state == 'uploaded':
+            if order.state in ('uploaded', 'done'):
                 raise UserError(_(
                     "You cannot delete an uploaded payment order. You can "
                     "cancel it in order to do so."))
@@ -359,13 +359,13 @@ class AccountPaymentOrder(models.Model):
 
     @api.multi
     def generated2uploaded(self):
-        for order in self:
-            if order.payment_mode_id.generate_move:
-                order.generate_move()
         self.write({
             'state': 'uploaded',
             'date_uploaded': fields.Date.context_today(self),
             })
+        for order in self:
+            if order.payment_mode_id.generate_move:
+                order.generate_move()
         return True
 
     @api.multi
@@ -514,3 +514,18 @@ class AccountPaymentOrder(models.Model):
         trfmoves = self._prepare_trf_moves()
         for hashcode, blines in trfmoves.items():
             self._create_reconcile_move(hashcode, blines)
+
+    @api.multi
+    def _all_lines_reconciled(self):
+        """
+        Return true if all payment lines' counterpart lines are reconciled too
+        Also return true if the counterpart account cannot be reconciled
+        """
+        self.ensure_one()
+        move_lines = self.env['account.move.line'].search([(
+            'bank_payment_line_id', 'in', self.mapped('bank_line_ids').ids
+        )]).mapped('move_id.line_ids')
+        return bool(move_lines) and all(
+            move_line.reconciled for move_line in move_lines
+            if move_line.account_id.reconcile
+        )
--- a/account_banking_sepa_credit_transfer/tests/test_sct.py
+++ b/account_banking_sepa_credit_transfer/tests/test_sct.py
@@ -199,7 +199,7 @@ class TestSCT(common.HttpCase):
             debtor_acc_xpath[0].text,
             self.payment_order.company_partner_bank_id.sanitized_acc_number)
         self.payment_order.generated2uploaded()
-        self.assertEqual(self.payment_order.state, 'uploaded')
+        self.assertEqual(self.payment_order.state, 'done')
         for inv in [invoice1, invoice2, invoice3, invoice4, invoice5]:
             self.assertEqual(inv.state, 'paid')
         return
@@ -276,7 +276,7 @@ class TestSCT(common.HttpCase):
             debtor_acc_xpath[0].text,
             self.payment_order.company_partner_bank_id.sanitized_acc_number)
         self.payment_order.generated2uploaded()
-        self.assertEqual(self.payment_order.state, 'uploaded')
+        self.assertEqual(self.payment_order.state, 'done')
         for inv in [invoice1, invoice2]:
             self.assertEqual(inv.state, 'paid')
         return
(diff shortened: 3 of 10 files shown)


pedrobaeza:
What I want to move is:

- Instead of generating `bank.payment.line`, to generate `account.payment`.
- On v14, it allows all the options of the transfer account (including specific account and journal).
- Use the standard flow, appearing as blue lines in reconciliation and being able to have them in "In process of payment" state in the invoice.

hbrunn:
your message is about 14, right?

pedrobaeza:
It can be done as well in v13, but yes, one missing thing in this version is to respect the define transfer account.

pedrobaeza:
But yes, for v12 I don't think it's possible.

pedrobaeza (approved):
Reviewing the PR, this can be complementary to what I'm talking about, so OK for me.

account_payment_order/models/account_payment_order.py
```diff
@@ -514,3 +514,20 @@ def generate_move(self):
...
         for hashcode, blines in trfmoves.items():
             self._create_reconcile_move(hashcode, blines)
+
+    @api.multi
+    def _all_lines_reconciled(self):
+        """
+        Return true if all payment lines' counterpart lines are reconciled too
+        Also return true if the counterpart account cannot be reconciled
+        """
+        result = True
```
StefanRijnhart: Slight preference for `self.ensure_one()` here. It will likely be a mistake when this is called on multiple orders ever.

account_payment_order/models/account_payment_order.py
```diff
@@ -514,3 +514,20 @@ def generate_move(self):
...
+        for this in self:
+            move_lines = self.env['account.move.line'].search([(
+                'bank_payment_line_id', 'in', this.mapped('bank_line_ids').ids
+            )]).mapped('move_id.line_ids')
+            result &= all(
```
StefanRijnhart: I know this is currently only called on orders with move lines, but in case this gets a life of its own it might be useful to check ifthere are any move lines to begin with.

luc-demeyer:
I see the code is already in process of being merged, hence my feedback comes a bit late:
imho we should not put the state back to uploaded once the order is done.
If the accountant removes a reconciliation than he usually knows which actions he needs to perform to redo the reconciles but he will not expect a status change on a done payment order.
I suggest to drop the def remove_move_reconcile(self) logic.

pedrobaeza:
It's still time for comments, Luc, as no merge command has been launched. I think the reconcile cancel logic is OK, as long as after reconciling again, it turns again to "Done" status, which is what seems to happen.

hbrunn:
yes, when you reconcile those move lines again the state goes back to done. @luc-demeyer would you be satisfied by a test that proves this?

luc-demeyer:
@hbrunn @pedrobaeza 
Even though the order comes back to 'done' after redoing the reconcile I think it will not be appreciated by the accountant to reopen a 'done' payment order.
Consider it as a TODO list for the accountant for his payments. Once it's done, it means to him that either the payment has been successfully executed or alternatively that he may even never perform the payment (in which case the open item will usually be reconciled via a Refund or a Journal Entry in a Miscellanuous Journal).
For V13 we have added a button to set a payment order to 'done', cf. https://github.com/OCA/bank-payment/commit/752da9a23f866a2ce9358a1daf4fc0c7096ff06a, which I think is a good idea to force a payment order to 'done'. I mentionned the partial reconciles in https://github.com/OCA/bank-payment/pull/784#issuecomment-896785175. Imagine following use case: a payment order is uploaded by the accountant assistant to the web banking application, the accountant connects to the web banking application and changes the amount of one of the payments because of e.g. there is a dispute on a delivery before signing and hence executing the payment. Once the bank statement is downloaded with this reduced amount we'll have a partial reconcile for that particular supplier invoice and this one will remain in that state until the dispute is resolved (could be easily half year later). For the accountant the payment order is 'done'. After dispute resolution the open balance will be paid via a new Payment Order (and that one will be automatically set to done with the logic in this PR).
Hence I think it would be useful to also cherry-pick  https://github.com/OCA/bank-payment/commit/752da9a23f866a2ce9358a1daf4fc0c7096ff06a for V12.

hbrunn:
thanks for the point of view of the accountant. Is everyone on board with removing the unreconcile functionality and adding https://github.com/OCA/bank-payment/commit/752da9a23f866a2ce9358a1daf4fc0c7096ff06a here?

pedrobaeza:
In my country, it's not possible to edit a sent SEPA order, and if it's the case, as you have said: you have the manual done button to set it again as done. I'm against removing such part, as it's the way to control things. Invoices work this way as well, removing its paid state when unreconciling.

account_payment_order/models/account_move_line.py
```diff
@@ -148,3 +148,47 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
...
         return result
+
+    @api.multi
+    def reconcile(self, writeoff_acc_id=False, writeoff_journal_id=False):
+        """ Set payment orders with fully reconciled lines to done """
+        result = super().reconcile(
+            writeoff_acc_id=writeoff_acc_id,
+            writeoff_journal_id=writeoff_journal_id,
+        )
+        self._close_payment_orders()
```
StefanRijnhart: Filter on items with a full_reconcile_id here?

account_payment_order/models/account_move_line.py
```diff
@@ -148,3 +148,47 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
...
+                order.action_done()
+
+    @api.multi
+    def _find_payment_orders(self):
+        """
+        Return all payment orders linked (directly by payment_line_ids
+        or indirectly by reconciliation with a transfer account) to self
+        """
+        return self.mapped(
+            'move_id.line_ids.bank_payment_line_id.order_id'
```
StefanRijnhart, line 180: ~~Are other lines from the reconciled moves relevant or would self.mapped("bank_payment_line_id.order_id") be equivalent and a little more efficient?~~ Ah no

account_payment_order/models/account_move_line.py
```diff
@@ -148,3 +148,47 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
...
+        ) | self.mapped(
+            'full_reconcile_id.reconciled_line_ids.move_id.line_ids.'
+            'bank_payment_line_id.order_id'
+        )
```
StefanRijnhart, line 184: It bothers me a little bit that this code generates a false positive when the partner move line is reconciled.
For orders that are transferred to a non-reconcilable account, it triggers the order's `action_done` and makes the order go to `done` state temporarily before state is set to 'uploaded' here: https://github.com/OCA/bank-payment/blob/14.0/account_payment_order/models/account_payment_order.py#L416

  by replacing the above with something like this:

```
reconciled_lines = self.mapped('full_reconcile_id.reconciled_line_ids')
return (reconciled_lines.mapped('move_id.line_ids') - reconciled_lines).mapped(
    'bank_payment_line_id.order_id')
```

but it requires that you call this method in `remove_move_reconcile` before calling super. This would also cover my comment in the `reconcile` method.

However, the same problem still persists in case the transfer setting is 'bank_account', because then the monitored reconciliation is taking place in this early stage. Maybe it also needs to be avoided at that point using a context value or something.

hbrunn: I think the filtering above is a good one. Why not here simply move the `write` above the call to `generate_move`?

StefanRijnhart: Ah, you mean the state change to 'done' is intentional for these types, and the problem is that the state is then reset to 'uploaded' afterwards. Yes, that makes sense.

hbrunn: great, did so in  https://github.com/OCA/bank-payment/pull/860/commits/1c2472f473156fb6f6dab6e0af65ff7a598d75b7. Also added a context key allowing to suppress this behavior, which we'll probably need for ie. bank statements where we only want to run the check once for all reconciled lines and not per reconciliation. Also cherry picked https://github.com/OCA/bank-payment/commit/752da9a23f866a2ce9358a1daf4fc0c7096ff06a

StefanRijnhart (approved):

EvvFoxy:
> I see the code is already in process of being merged, hence my feedback comes a bit late: imho we should not put the state back to uploaded once the order is done. If the accountant removes a reconciliation than he usually knows which actions he needs to perform to redo the reconciles but he will not expect a status change on a done payment order. I suggest to drop the def remove_move_reconcile(self) logic.

@luc-demeyer @pedrobaeza @hbrunn From a functional point of view, I agree with Luc Demeyer. Undoing a reconciliation should never reopen a payment order. A reconciliation can be undone for different reasons, but IMHO should never lead to reopening a payment order.

account_payment_order/models/account_move_line.py
hbrunn: I had to add another 35215233 to make the other tests work then

hbrunn:
now two people for don't mess with the state on unreconcile and two against. @StefanRijnhart would you care to break the tie?

StefanRijnhart:
@hbrunn I believe that in the case that the transfer move of a payment order is unreconciled after it was first reconciled it does not mean that the payment order is not done. I can only see this happen as a temporary, administrative situation so I tend to go with Luc and Els. Pedro's argument about the analogy with reopened invoices rings true on the technical level yet at the same time it will be much more common to have multiple invoices with the same amount, even for the same customer, than it is to have payment orders of the same amount so it makes much more sense to have this behaviour for invoices, but maybe not for payment orders.

pedrobaeza:
Ok, you have one vote more for the no state modification on unreconciling after Stefan's argumentation.

hbrunn:
now without reopening on unreconciliation. Thanks everyone involved!

luc-demeyer (approved):

PR #870, opened 2021-11-24
[15.0] [MIG] account_payment_mode

misern2:
Standrad Migration


ajaniszewska-dev (approved):

account_payment_mode/models/account_journal.py
```diff
@@ -0,0 +1,60 @@
...
+        all_out = self.env["account.payment.method"].search(
+            [("payment_type", "=", "outbound")]
+        )
+        return all_out
+
+    def _default_inbound_payment_methods(self):
+        all_in = self.env["account.payment.method"].search(
+            [("payment_type", "=", "inbound")]
+        )
+        return all_in
```
i-vyshnevska, line 22: fields to this method were removed 
do we need to migrate them or remove them as well?

i-vyshnevska (commented):
translation commits can be squashed

account_payment_mode/models/account_journal.py
gurneyalex: The methods are still present and used in account.journal:

https://github.com/odoo/odoo/blob/87e913b892c2518a716fceb247b57ab8c955a07f/addons/account/models/account_journal.py#L32-L36

https://github.com/odoo/odoo/blob/87e913b892c2518a716fceb247b57ab8c955a07f/addons/account/models/account_journal.py#L280-L302

so it totally makes sense to keep them in the module. 

(actually these were never method to compute a default value for a field on account.journal, but used to compute a default value for a payment being recorded on that journal)

account_payment_mode/tests/test_account_payment_mode.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2016-2020 ForgeFlow S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo.exceptions import UserError, ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestAccountPaymentMode(TransactionCase):
+    def setUp(self):
```
simahawk: pls move this setup code to `setUpClass` and start like this:

```py
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
```

account_payment_mode/views/account_payment_method.xml
```diff
@@ -0,0 +1,82 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+    <record id="account_payment_method_tree" model="ir.ui.view">
+        <field name="name">account_payment_method.tree</field>
+        <field name="model">account.payment.method</field>
+        <field name="arch" type="xml">
+            <tree>
+               <!--<field name="sequence" widget="handle" /> -->
```
simahawk: needed?

account_payment_mode/models/account_journal.py
simahawk: @i-vyshnevska fine for you?

simahawk (changes requested):

account_payment_mode/tests/test_account_payment_mode.py
misern2: :+1:

account_payment_mode/views/account_payment_method.xml
misern2: Not needed. Removed. :+1:

i-vyshnevska (approved):

simahawk (approved):

pedrobaeza:
@simahawk FYI we have that new command that I have put in the previous comment for auto-filling migration issue and milestone.

simahawk:
> @simahawk FYI we have that new command that I have put in the previous comment for auto-filling migration issue and milestone.

waaaat? I missed it! Cool! :smile: 
Thanks for the heads up :+1:

PR #873, opened 2021-12-02
[15.0] [MIG] account_payment_partner

misern2:
Standard Migration
Depends on
- [x] #870


account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,155 @@
...
+                move.partner_bank_filter_type_domain = move.bank_partner_id
+            elif move.move_type in ("in_invoice", "out_refund"):
+                move.partner_bank_filter_type_domain = move.commercial_partner_id
+            else:
+                move.partner_bank_filter_type_domain = False
+
+    @api.depends("partner_id", "company_id")
+    def _compute_payment_mode(self):
+        for move in self:
+            move.payment_mode_id = move.payment_mode_id
```
i-vyshnevska:
```suggestion
```

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,155 @@
...
+    @api.depends("partner_id", "payment_mode_id")
+    def _compute_partner_bank(self):
+        for move in self:
+            # No bank account assignation is done for out_invoice as this is only
+            # needed for printing purposes and it can conflict with
+            # SEPA direct debit payments. Current report prints it.
+            def get_bank_id():
+                return move.commercial_partner_id.bank_ids.filtered(
+                    lambda b: b.company_id == move.company_id or not b.company_id
+                )[:1]
```
i-vyshnevska:
```suggestion
                return fields.first(move.commercial_partner_id.bank_ids.filtered(
                    lambda b: b.company_id == move.company_id or not b.company_id
                ))
```

simahawk:
@misern2 to depend on #870 for the build, you can add a `test-requirements.txt` file (if not present yet) and add a line like this
```
odoo${10|11|12|13|14}-addon-${addon_name} @ git+https://github.com/OCA/$repo_name@refs/pull/$PR_NR/head#subdirectory=setup/$addon_name
```
In this case it would be:
```
odoo-addon-account_payment_mode @ git+https://github.com/OCA/bank-payment@refs/pull/870/head#subdirectory=setup/account_payment_mode
```
Note that for v15 we don't need the Odoo version in the pkg namespace `odoo-addon`.

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,155 @@
...
+
+    payment_mode_filter_type_domain = fields.Char(
+        compute="_compute_payment_mode_filter_type_domain"
+    )
+    partner_bank_filter_type_domain = fields.Many2one(
+        comodel_name="res.partner", compute="_compute_partner_bank_filter_type_domain"
+    )
+    payment_mode_id = fields.Many2one(
+        comodel_name="account.payment.mode",
+        compute="_compute_payment_mode",
```
simahawk:
```suggestion
        compute="_compute_payment_mode_id",
```

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,155 @@
...
+                or self.partner_id.valid_mandate_id.partner_bank_id
+            )
+        # Return this as empty recordset
+        return self.partner_bank_id
+
+    @api.model
+    def create(self, vals):
+        """Force compute partner_bank_id when invoice is created from SO
+        to avoid that odoo _prepare_invoice method value will be set.
+        """
```
simahawk: this is not the docstring of the create method, use a comment pls

account_payment_partner/tests/test_account_payment_partner.py
```diff
@@ -0,0 +1,580 @@
...
+from odoo import _, fields
+from odoo.exceptions import UserError, ValidationError
+from odoo.fields import Date
+from odoo.tests.common import Form, TransactionCase
+
+
+class TestAccountPaymentPartner(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 14:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

account_payment_partner/models/account_move.py
```diff
@@ -0,0 +1,155 @@
...
+        store=True,
+        ondelete="restrict",
+        readonly=False,
+        check_company=True,
+    )
+    bank_account_required = fields.Boolean(
+        related="payment_mode_id.payment_method_id.bank_account_required", readonly=True
+    )
+    partner_bank_id = fields.Many2one(
+        compute="_compute_partner_bank",
```
simahawk:
```suggestion
        compute="_compute_partner_bank_id",
```

simahawk (changes requested):
LG overall

account_payment_partner/models/account_move.py
misern2: @i-vyshnevska Thanks for pointing out. I didn't know this method! :+1:

alexis-via:
No need to add the field commercial_partner_id as invisible to the "view_move_form" view because it is already present in the original view, cf:
https://github.com/odoo/odoo/blob/15.0/addons/account/views/account_move_views.xml#L733

misern2:
> No need to add the field commercial_partner_id as invisible to the "view_move_form" view because it is already present in the original view, cf: https://github.com/odoo/odoo/blob/15.0/addons/account/views/account_move_views.xml#L733

@alexis-via  Done!. :+1:

eduardgm (approved):
Functional review, LGTM +1

HaraldPanten (approved):
Functional review 👍

@simahawk Could you update your review, please?

@alexis-via @pedrobaeza WDYT about this PR? Could it be merged?

TY all.

PR #874, opened 2021-12-09
[14.0][FIX] account_banking_sepa_direct_debit: Post mandate expiration message by mandate register

ao-landoo:
When multiple mandates are set to `expired` state, it's currently calling `message_post` function, which is supposed to be called from a single register (`ensure_one`).

I created a loop so that `message_post` is called for each mandate that has been expired.

@CarlosRoca13 @alexis-via what do you think?

### DIFF
--- a/account_banking_sepa_direct_debit/models/account_banking_mandate.py
+++ b/account_banking_sepa_direct_debit/models/account_banking_mandate.py
@@ -102,16 +102,14 @@ class AccountBankingMandate(models.Model):
         )
         if expired_mandates:
             expired_mandates.write({"state": "expired"})
-            expired_mandates.message_post(
-                body=_(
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
Trivial. Thanks !

pedrobaeza (approved):

alexis-via:
Thanks for the fix !

pedrobaeza:
@rousseldenis the error may be to the change in Odoo that requires tests to be post-installed?

rousseldenis:
> @rousseldenis the error may be to the change in Odoo that requires tests to be post-installed?

I include the fix here https://github.com/OCA/bank-payment/pull/918

If you want to merge that one before :smiley:

pedrobaeza:
Please fw-port it to v15

PR #875, opened 2021-12-14
[13.0][FIX] account_banking_sepa_direct_debit: Bank Account should not be required

etobella:
The bank account is taken from the mandate. `invoice_partner_bank_id` is never used in this module, so it should not be taken as required.

### DIFF
--- a/account_banking_sepa_credit_transfer/data/account_payment_method.xml
+++ b/account_banking_sepa_credit_transfer/data/account_payment_method.xml
@@ -4,7 +4,7 @@
         <field name="name">SEPA Credit Transfer to suppliers</field>
         <field name="code">sepa_credit_transfer</field>
         <field name="payment_type">outbound</field>
-        <field name="bank_account_required" eval="True" />
+        <field name="bank_account_required" eval="False" />
         <field name="pain_version">pain.001.001.03</field>
     </record>
 </odoo>
--- a/account_banking_sepa_direct_debit/data/account_payment_method.xml
+++ b/account_banking_sepa_direct_debit/data/account_payment_method.xml
@@ -4,8 +4,8 @@
         <field name="name">SEPA Direct Debit for customers</field>
         <field name="code">sepa_direct_debit</field>
         <field name="payment_type">inbound</field>
-        <field name="bank_account_required" eval="True" />
-        <field name="mandate_required" eval="True" />
+        <field name="bank_account_required" eval="False" />
+        <field name="mandate_required" eval="False" />
         <field name="pain_version">pain.008.001.02</field>
     </record>
 </odoo>


account_banking_sepa_direct_debit/data/account_payment_method.xml
```diff
@@ -4,7 +4,7 @@
         <field name="name">SEPA Direct Debit for customers</field>
         <field name="code">sepa_direct_debit</field>
         <field name="payment_type">inbound</field>
-        <field name="bank_account_required" eval="True" />
+        <field name="bank_account_required" eval="False" />
         <field name="mandate_required" eval="True" />
```
pedrobaeza: This one as well, and also for credit transfer, please

JordiBForgeFlow (approved):

account_banking_sepa_direct_debit/data/account_payment_method.xml
pedrobaeza: Mandate shouldn't be required as well.

pedrobaeza (approved):

account_banking_sepa_credit_transfer/data/account_payment_method.xml
```diff
@@ -4,7 +4,7 @@
         <field name="name">SEPA Credit Transfer to suppliers</field>
         <field name="code">sepa_credit_transfer</field>
         <field name="payment_type">outbound</field>
-        <field name="bank_account_required" eval="True" />
+        <field name="bank_account_required" eval="False" />
         <field name="pain_version">pain.001.001.03</field>
```
alexis-via, line 8: Why do you change this?? You are modifying the file for credit transfers !

pedrobaeza: Such bank account is as well not needed, as it's auto-populated on payment order filling.

alexis-via: Then change the code, don't change the value of this field.

alexis-via: Or remove the field. It would be ridiculous to change the value of those field to False.

pedrobaeza: Why is ridiculous? We just don't want that option enabled for this payment method by default.

etobella: I changed the value in order to make it clear that it is not required

pedrobaeza:
@alexis-via we want to continue with the merging.

HaraldPanten (approved):
Small changes. LGTM

pedrobaeza:
Cherry-picked for v14 at 24d7a061c6229ad9c5c612b8ccf8fadd8cab66d5 and f8b0245226b686bacd7c2b885b00bdb70929d32b

account_banking_sepa_credit_transfer/data/account_payment_method.xml
alexis-via: This is completely ridiculous!!! In the definition of the account.payment.method for credit transfer, we say that bank account is not required !!! If it's not required for credit transfer, for which payment method will it be required? You should have changed the code to make it required starting from the payment line instead of starting from the invoice. That could make sense. But the change here doesn't make sense at all. You're screwing up the architecture.

pedrobaeza: No, it isn't. Requiring it at invoice level is what it doesn't make sense, and more populate it when selecting the partner. Maybe the invoice was done 8 months ago, and the partner has changed 3 times of bank account. The bank account is populated when you do the payment/debit order, and shouldn't be before if you want to avoid surprises. If you tell me, I would totally remove such checks.

alexis-via:
BRAVO BRAVO BRAVO
You just broke the OCA implementation of SEPA direct debit. According to my test in v14 and an audit of the code:
1) you can't select a mandate on an invoice any more (field hidden, cf https://github.com/OCA/bank-payment/blob/14.0/account_banking_mandate/views/account_move_view.xml#L16)
2) you can't select a mandate on a payment line any more (field hidden, cf https://github.com/OCA/bank-payment/blob/14.0/account_banking_mandate/views/account_payment_line.xml#L20)
3) with this change, the code that checks that a mandate is present when you confirm the payment order is disabled (cf https://github.com/OCA/bank-payment/blob/14.0/account_banking_mandate/models/account_payment_line.py#L60)

The change introduced by this PR don't reflect the title of the PR nor the message of @etobella at the begining of the PR. I'm ok to set bank_account_required to False on sepa_direct_debit, but:
- I'm not ok to set mandate required to False on sepa_direct_debit
- I'm not ok to set bank account required to False on sepa_credit_transfer
I am the author of this code and I had the feeling that we had a problem with this PR, but my messages were not taken seriously. I'll propose to partially revert it and then, it we want to make additional changes, take the time to discuss them and measure the impact of the changes.

account_banking_sepa_credit_transfer/data/account_payment_method.xml
alexis-via: OK, thanks for giving the real reason for the change... at last ! I can agree with this reason, but it's not written in the title of the PR nor in the first comment of @etobella You should have told your motivations clearly, I would have advised on the proper solution. I'm the author of this code, so I have a good idea of the consequences, and it would have avoided the breakage.

PR #879, opened 2021-12-17
[14.0][FW] [IMP] account_payment_sale: set account_id on payment order…

hailangvn:
… if fixed journal is set on payment mode

* This is to forward port #751.

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -31,9 +31,12 @@ class SaleOrder(models.Model):
                 self.payment_mode_id.bank_account_link == "fixed"
                 and self.payment_mode_id.payment_method_id.code == "manual"
             ):
-                vals[
-                    "partner_bank_id"
-                ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
+                journal = self.payment_mode_id.fixed_journal_id
+                vals["partner_bank_id"] = journal.bank_account_id.id
+                if journal.payment_debit_account_id:
+                    for line in vals["invoice_line_ids"]:
+                        line[2]["account_id"] = journal.payment_debit_account_id.id
+        return vals
 
     def _prepare_invoice(self):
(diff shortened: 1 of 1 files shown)


leemannd (approved):
Technically looks good
Was done in 12.0 IDK if it is required in 14.0. Are there unit tests about this code?

hailangvn:
> Technically looks good Was done in 12.0 IDK if it is required in 14.0. Are there unit tests about this code?

Thanks @leemannd for your review. Yes, the code is covered as https://app.codecov.io/gh/OCA/bank-payment/compare/879/diff.

account_payment_sale/models/sale_order.py
```diff
@@ -31,9 +31,12 @@ def _get_payment_mode_vals(self, vals):
...
                 and self.payment_mode_id.payment_method_id.code == "manual"
             ):
-                vals[
-                    "partner_bank_id"
-                ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
+                journal = self.payment_mode_id.fixed_journal_id
+                vals["partner_bank_id"] = journal.bank_account_id.id
+                if journal.payment_debit_account_id:
+                    for line in vals["invoice_line_ids"]:
+                        line[2]["account_id"] = journal.payment_debit_account_id.id
```
grindtildeath, line 38: This is really wrong as this code is supposed to change the account for every lines that would have been prepared when it's executed :warning: 
Fortunately, the invoice lines are prepared after the call to `_prepare_invoice`, therefore this code doesn't mess with the data but it's not doing what it's supposed to do.

grindtildeath (changes requested):
I'll work on an alternative solution, I guess we'll close this PR.

account_payment_sale/models/sale_order.py
grindtildeath: Issue is solved through https://github.com/OCA/bank-payment/pull/937

PR #881, opened 2021-12-18
[14.0]account_payment_order - hide reference_type from invoice view

luc-demeyer:
I think we should not add the reference_type field for outgoing invoices in the account_payment_order module.
Adding this field for outgoing invoices adds confusion for the end user.
It could be added by modules that introduce other payment reference types, but even in that case I am not sure we need to visualise this field on outgoing invoices since the structured communication will probably be autocalculated and readonly for outgoing invoices.


pedrobaeza:
I don't use such field, but not sure if for Belgium with BBA is needed. @sbidoul do you use it?

pedrobaeza:
@luc-demeyer remember to put the involved module in the commit title as well on the PR title.

rousseldenis:
@luc-demeyer

ghost (approved):

PR #887, opened 2022-01-07
[15.0][MIG] account_banking_sepa_direct_debit

alexis-via:


account_banking_sepa_direct_debit/data/account_payment_method.xml
```diff
@@ -5,7 +5,7 @@
         <field name="code">sepa_direct_debit</field>
         <field name="payment_type">inbound</field>
         <field name="bank_account_required" eval="False" />
-        <field name="mandate_required" eval="False" />
+        <field name="mandate_required" eval="True" />
```
pedrobaeza, line 8: Please don't change this again.

pedrobaeza:
You are missing the procedure at https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0 where a previous commit is done for pre-commit stuff like `setup` folder generation.

account_banking_sepa_direct_debit/data/account_payment_method.xml
alexis-via: If mandate_required=False, it doesn't work any more !!! So I change it back... to allow users to use the module!
On the payment order line, you the mandate field is invisible when mandate_required is False, cf https://github.com/akretion/banking/blob/15-mig-account_banking_mandate/account_banking_mandate/views/account_payment_line.xml#L20
I guess it's also broken on v14...

pedrobaeza: Then remove such invisible part, which is the nonsense, as you one thing is to be required, and the other to be hidden. I can fix it in v14 that way.

account_banking_sepa_direct_debit/models/account_banking_mandate.py
```diff
@@ -0,0 +1,121 @@
...
+        tracking=80,
+    )
+    unique_mandate_reference = fields.Char(size=35)  # cf ISO 20022
+    display_name = fields.Char(compute="_compute_display_name2", store=True)
+
+    @api.constrains("type", "recurrent_sequence_type")
+    def _check_recurring_type(self):
+        for mandate in self:
+            if mandate.type == "recurrent" and not mandate.recurrent_sequence_type:
+                raise exceptions.Warning(
```
rlizana: Warning exceptions is deprecated, change to UserError please

rlizana (changes requested):

MiquelRForgeFlow:
please cherry-pick https://github.com/OCA/bank-payment/commit/a5d23b17917d0de4b2c62b8403cd56c370d6c2ec

HaraldPanten:
superseded by another PR

PR #888, opened 2022-01-08
[14.0][FIX] mandate_required back to True on SDD, bank_account_required back to True on SCT

alexis-via:
Partial revert of the breakage of PR #875

### DIFF
--- a/account_banking_sepa_credit_transfer/data/account_payment_method.xml
+++ b/account_banking_sepa_credit_transfer/data/account_payment_method.xml
@@ -4,7 +4,7 @@
         <field name="name">SEPA Credit Transfer to suppliers</field>
         <field name="code">sepa_credit_transfer</field>
         <field name="payment_type">outbound</field>
-        <field name="bank_account_required" eval="False" />
+        <field name="bank_account_required" eval="True" />
         <field name="pain_version">pain.001.001.03</field>
     </record>
 </odoo>
--- a/account_banking_sepa_direct_debit/data/account_payment_method.xml
+++ b/account_banking_sepa_direct_debit/data/account_payment_method.xml
@@ -5,7 +5,7 @@
         <field name="code">sepa_direct_debit</field>
         <field name="payment_type">inbound</field>
         <field name="bank_account_required" eval="False" />
-        <field name="mandate_required" eval="False" />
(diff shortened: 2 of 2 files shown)


etobella:
@alexis-via I checked your comments and I understand your problem with the missing field. It is true that with the change we cannot select anymore the mandate, so that's a problem, but when marked, it is incorrect to make it required, as the code will fill it properly if not found. So, we can make it true again but make it not required in `account.move`. WDYT?

https://github.com/OCA/bank-payment/blob/14.0/account_banking_mandate/views/account_move_view.xml#L16

Maybe it will be right to make the same here with the bank_account_required

https://github.com/OCA/bank-payment/blob/14.0/account_payment_partner/views/account_move_view.xml#L44

alexis-via:
I'm open to a discussion about whether the mandate (SDD) or bank account (SCT) should start to be required on invoice or only start to be required on payment line. I don't have a strong opinion on this at the moment. My initial idea when I designed this during the Sorrento code sprint was : "this info (mandate for SDD, bank account for SCT) is required. The sooner it is set by the user, the better".
Maybe, with the experience, we think that this info should only be required "at the last moment" i.e. only on payment line. I would be interested to hear the real-life experience of those who think that making it required "at the last moment" is the best option. Pedro mentioned that a supplier can change his bank account between the time it send it's invoice and the time we have to make the payment. It's probably not very common, but I can hear that argument. There are probably other more convincing arguments.
We can take some time to decide this, there is no emergency. The emergency is to repair what has been broken, so I propose to merge this today and make another PR to propose to move the requirement from invoice to payment line.

pedrobaeza:
OK, let's merge it:

/ocabot merge minor

pedrobaeza:
FYI, the 100% of my instances have these marks unchecked, and they don't feel it's broken. In fact, there's no problem with not requiring them even at payment line level. Later when confirming the payment order, the implemented check is avoiding to have incorrect data.

And the changing bank accounts problem is not only for suppliers. It's for customer as well, and in fact, it's a real case of one of my customers: they debit to an incorrect account and get a return with the corresponding fees.

pedrobaeza:
@alexis-via I reverted this with the promise that you we'll study this for avoiding the lock-ins, but letting the check unmarked, and no change for now.

PR #890, opened 2022-01-14
[14.0][IMP] account_payment_partner: setting to force blank partner bank

hailangvn:
The l10n-switzerland/l10n_ch_isr_payment_grouping module uses partner
bank to work. This is quick fix while waiting module owners, who has
more business knowledge to resolve conflict.

### DIFF
--- a/account_payment_partner/__manifest__.py
+++ b/account_payment_partner/__manifest__.py
@@ -15,6 +15,7 @@
     "depends": ["account_payment_mode"],
     "data": [
         "views/res_partner_view.xml",
+        "views/res_config_settings.xml",
         "views/account_move_view.xml",
         "views/account_move_line.xml",
         "views/account_payment_mode.xml",
--- a/account_payment_partner/models/__init__.py
+++ b/account_payment_partner/models/__init__.py
@@ -1,1 +1,2 @@
 from . import account_move, account_move_line, account_payment_mode, res_partner
+from . import res_config_settings, res_company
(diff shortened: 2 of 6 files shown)


hailangvn:
Good day @pedrobaeza, could you please advise who is the owner of the `account_payment_partner` module? And who can review this.

pedrobaeza:
Why don't you simply remove the "bank required" check on the payment method?

hailangvn:
Thanks @pedrobaeza for your suggestion. The method has [comment](https://github.com/OCA/bank-payment/blob/14.0/account_payment_partner/models/account_move.py#L97) `No bank account assignation is done` so current logic is to set `bank_id = False` [by default](https://github.com/OCA/bank-payment/blob/14.0/account_payment_partner/models/account_move.py#L105) and assign value to it when both conditions below are true.
    a) [payment mode is set](https://github.com/OCA/bank-payment/blob/14.0/account_payment_partner/models/account_move.py#L110), and
    b) [its payment method](https://github.com/OCA/bank-payment/blob/14.0/account_payment_partner/models/account_move.py#L112) has `bank_account_required = True`
Meanwhile, all impacted tests are running without payment mode thus (a) does not meet. [Trying to add payment mode](https://github.com/OCA/l10n-switzerland/commit/98e2bf9d2896058f832119f371e5ed4d18ce0e16) still does not resolve the test error due to combined reference `120000000000234478943216899 703192500010549027000209403`. Will try to allocate time to investigate why.

hailangvn:
Thanks to @pedrobaeza's suggestion, the test of modules in other repo has been updated to work properly with this module. So this PR is not needed anymore.

PR #893, opened 2022-01-23
[13.0][IMP] account_payment_order - Split method create_account_payment_line for better inheritability

ghost:
In the current logic, if we want to change the applicable lines to add in a payment order trough the actions available on a move, we have to inherit the full method. We suggest to split the method create_account_payment_line in two for better inheritability of that logic.

In our customer case, we repurpose the Purchase Receipts and we want the purchase receipt lines to be added in the payment orders. The move lines to be added are not using a receivable or payable account.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -48,6 +48,19 @@ class AccountMove(models.Model):
         # in account_payment_order.py
         return vals
 
+    def _get_applicable_lines(self, move):
+        return move.line_ids.filtered(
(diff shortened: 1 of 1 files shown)


luc-demeyer (approved):

rousseldenis (approved):
Great to see you on OCA !

MiquelRForgeFlow (approved):

AaronHForgeFlow:
/ocabot merge patch

Please check if FW port is needed

PR #898, opened 2022-02-23
[14.0][FIX] account_banking_sepa_*_transfer: add migration script

MiquelRForgeFlow:
Followup of https://github.com/OCA/bank-payment/commit/144d2d98b14acd5ec38ee8b1b6c9a8273f1dbbcf (migration)


pedrobaeza:
This shouldn't be put. I want to be False in my installations, and in fact, @alexis-via promised to study another solution for harmonizing both things that we want removing such flag.

MiquelRForgeFlow:
Ok

PR #901, opened 2022-02-25
[FIX] [14.0] Removed external dependencies already defined in Odoo base

Shide:
Removed external dependencies already defined in Odoo base

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L48

MT-282 @moduon

### DIFF
--- a/account_banking_sepa_direct_debit/__manifest__.py
+++ b/account_banking_sepa_direct_debit/__manifest__.py
@@ -11,7 +11,6 @@
     "website": "https://github.com/OCA/bank-payment",
     "category": "Banking addons",
     "depends": ["account_banking_pain_base", "account_banking_mandate"],
-    "external_dependencies": {"python": ["stdnum"]},
     "data": [
         "views/account_banking_mandate_view.xml",
         "views/res_config_settings.xml",
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 # generated from manifests external_dependencies
 lxml
-python-stdnum
 unidecode


pedrobaeza:
I think the proper patch is to simply remove that external dependency, as it's included in Odoo core:

https://github.com/odoo/odoo/blob/3ff444ecc0213b41c5e456e724c8ef4aa4f94ed0/requirements.txt#L48

yajo (approved):
Can be improved later. For now, it fixes the problem. 😃

pedrobaeza (changes requested):
Remove the external dependency instead

yajo:
Is that really a better solution?

Imagine that Odoo removes that dependency in Odoo 16.0. Keeping it in the manifest would make that incompatibility evident.

I mean, I'm not going to make a rule out of this, but since it's another approach to fix the problem and it has its own benefits, what's the real problem in current solution? 🤔

pedrobaeza:
If this would be an optional library included in Odoo by chance, I would agree with the principle of "explicit better than implicit", but being so basic library that is there for years and years, I prefer to have this clean, because in the same manner, tomorrow you have another fork of `python-stdnum` being used by Odoo, and you will have the same theoretical problem here declaring a explicit library incompatible with core. And being in the same mood of explicit better than implicit, then you should need to declare lxml external dependency due to this: https://github.com/OCA/bank-payment/blob/14.0/account_banking_sepa_direct_debit/models/account_payment_order.py#L5, so I think it's better to keep the manifest minimal in this case.

yajo:
Yes, I understand that it's not a rule to include the external dependencies if Odoo depends on them.

My question is: why is it now a rule to not include them, if it harms nobody?



> tomorrow you have another fork of `python-stdnum` being used by Odoo

You mean if they do their own fork of that library and name it differently?

pedrobaeza:
> Yes, I understand that it's not a rule to include the external dependencies if Odoo depends on them.
> 
> My question is: why is it now a rule to not include them, if it harms nobody?

In this case, the DRY principle that you also like a lot is applied in a non-written rule.

> 
> > tomorrow you have another fork of `python-stdnum` being used by Odoo
> 
> You mean if they do their own fork of that library and name it differently?

Yeah, for example, or `python-stdnum` gets unmaintained through time and other people starts a fork and Odoo decides to use such fork instead. This a similar case to the one you have talked about Odoo removing the dependency over. All of them are "What if?" that are very theoretical and solvable when they happen.

pedrobaeza (approved):

PR #907, opened 2022-03-08
[15.0][FIX] account_payment_partner: only use bank_account_required in bills

MiquelRForgeFlow:
Forward port of https://github.com/OCA/bank-payment/pull/905.

### DIFF
--- a/account_payment_partner/views/account_move_view.xml
+++ b/account_payment_partner/views/account_move_view.xml
@@ -40,8 +40,9 @@
                     '|',('company_id', '=', company_id),('company_id', '=', False)]
                 </attribute>
                 <attribute name="attrs">
-                    {'required': [('bank_account_required', '=', True)], 'readonly':
-                    [('state', '!=', 'draft')]}
+                    {'required': [('bank_account_required', '=', True),('move_type', 'in', ('in_invoice', 'in_refund'))],
+                    'readonly': [('state', '!=', 'draft')],
+                    'invisible': [('move_type', '=', 'entry')]}
                 </attribute>
                 <attribute name="context">
                     {'default_partner_id':commercial_partner_id}


pedrobaeza (approved):

account_payment_partner/views/account_move_view.xml
```diff
@@ -40,8 +40,8 @@
                     '|',('company_id', '=', company_id),('company_id', '=', False)]
                 </attribute>
                 <attribute name="attrs">
-                    {'required': [('bank_account_required', '=', True)], 'readonly':
-                    [('state', '!=', 'draft')]}
+                    {'required': [('bank_account_required', '=', True),('move_type', 'in', ('in_invoice', 'in_refund'))],
+                    'readonly': [('state', '!=', 'draft')]}
```
pedrobaeza: What about invisible?

MiquelRForgeFlow: Yeah, but as it wasn't put, I also don't put it.

pedrobaeza: I think that attribute needs some love too, keeping the standard domain on it

LoisRForgeFlow: I agree with Pedro, you should also put the `'invisible': [('move_type', 'not in', ('in_invoice', 'in_refund', 'in_receipt'))]` which was the standard.

pedrobaeza: I think @alexis-via wanted to use this also for customer invoices. Isn't it?

MiquelRForgeFlow: @LoisRForgeFlow  not true. see https://github.com/OCA/bank-payment/blob/11.0/account_payment_partner/views/account_invoice_view.xml#L34

I put invisible only for entries

LoisRForgeFlow: Fair enough, it does not hurt

LoisRForgeFlow (approved):

PR #918, opened 2022-04-14
[14.0] account_payment_order: Better communication field management

rousseldenis:
This improves a little bit the communication field content.

- [x] Adds a hook
- [x] Take into account refunds in normal communication
- [x] Use of 'payment_reference' field if filled in.

Other change:
- [x] Set tests as post-install since https://github.com/odoo/odoo/commit/81aac30dd2278e43a43cf8fd6cef31e1e8c60f3f

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -67,15 +67,17 @@ class AccountMoveLine(models.Model):
                     self.move_id.move_type in ("in_invoice", "in_refund")
                     and self.move_id.ref
                 ):
-                    communication = self.move_id.ref
+                    communication = self.move_id.payment_reference or self.move_id.ref
                 elif "out" in self.move_id.move_type:
                     # Force to only put invoice number here
-                    communication = self.move_id.name
+                    communication = self.move_id.payment_reference or self.move_id.name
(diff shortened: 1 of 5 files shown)


rousseldenis:
@sbidoul @luc-demeyer @alexis-via @sanrav

rousseldenis:
@luc-demeyer I saw the payment_reference field has been added in v14 there https://github.com/odoo/odoo/commit/caeb782841fc5a7ad71a196e2c9ee67644ef9074

I think some flows can be impacted, no ?

driehuis:
@rousseldenis Desk checked, LGTM. I think the rule is that commits should be squashed whenever possible, you may want to use `git rebase -i HEAD~5` and `git push -f`.

rousseldenis:
> @rousseldenis Desk checked, LGTM. I think the rule is that commits should be squashed whenever possible, you may want to use `git rebase -i HEAD~5` and `git push -f`.

This assertion is true if each commit has no signification by its own. That is not true here. I want to keep them.

pedrobaeza (approved):
Thanks for the extra checks

pedrobaeza:
Let's fast-track this for solving the CI problem:

/ocabot merge minor

Can you please fw-port it to v15?

PR #925, opened 2022-04-22
[13.0][FIX] account_payment_order: Don't fail due to tz

pedrobaeza:
When the test hour is around midnight, and demo data using CET/CEST timezones, current code fails due to some asserts not having this into consideration. This fix avoids such problem, and irons the execution no matter the hour.

@Tecnativa


pedrobaeza:
Let's see if it fails :crossed_fingers:

pedrobaeza:
And got it!

/ocabot merge patch

PR #927, opened 2022-05-19
[14][account_payment_order] Fix French translation : fr_FR is not used

florian-dacosta:
There are 2 french .po files : fr.po and fr_FR.po
From what I see, the fr_FR.po is more up to date, must be updated by weblate.
Indeed, some translation in fr.po file don't make any sense.
But Odoo won't use fr_FR file for french lang, it will only use the base fr.po file.
So I replace the fr.po file by the fr_FR.po file.

@alexis-via


pedrobaeza:
As this cleaning can't be done through Weblate, accepting it:

/ocabot merge patch

PR #929, opened 2022-05-20
[FIX][14.0] account_payment_order_vendor_email

Vandan-OSI:
Remove the Extra Double Quote from the Mail Template.


flotho (approved):
easy LGTM

PR #930, opened 2022-05-27
[14.0][FIX] Fire trigger only when purchase_vendor_bill field has been changed

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/account_payment_purchase/models/account_invoice.py
+++ b/account_payment_purchase/models/account_invoice.py
@@ -19,6 +19,7 @@ class AccountMove(models.Model):
             self.purchase_vendor_bill_id.purchase_order_id.supplier_partner_bank_id.id
             or self.purchase_id.supplier_partner_bank_id.id
         )
+        has_purchase_vendor_bill = bool(self.purchase_vendor_bill_id)
 
         res = super()._onchange_purchase_auto_complete() or {}
         if self.payment_mode_id and new_mode and self.payment_mode_id.id != new_mode:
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
Makes sense

MiquelRForgeFlow (approved):

pedrobaeza (commented):
I think you have to change instead `new_mode` and `new_bank` variables for not assigning current one if no new is set.

dreispt:
@ChrisOForgeFlow Can you comment on Pedro's question please?

ChrisOForgeFlow:
> I think you have to change instead `new_mode` and `new_bank` variables for not assigning current one if no new is set.

I'm not sure if payment_mode_id should be changed with empty value(Empty values of new_move, new_bank), because that I tried to fire change  only when field purchase_vendor_bill_id is assigned, otherwise the field will be empty even when the onchange is fired by an empty field, like purchase_id

Because that I think that change only be triggered when field is not empty

ChrisOForgeFlow:
> @ChrisOForgeFlow Can you comment on Pedro's question please?

Done, thanks

pedrobaeza:
That's not what I was saying. I want to avoid such convoluted code for not changing what is not needed.

pedrobaeza:
The proper fix IMO is #955, that is already in 13.0. Can we close this one in favor of the other?

pedrobaeza:
Closing in favor of #955 as no more comments.

PR #933, opened 2022-06-13
[10.0][IMP] account_banking_pain_base: Allow to add bank account currency

rousseldenis:
Following this : 

image


rousseldenis:
@alexis-via @OCA/accounting-maintainers

PR #936, opened 2022-06-29
[FIX]account_payment_order

Vandan-OSI:
**Step to Reproduce:**
1.	Invoicing > Vendors > Bills
2.     Create Vendor Bill with Partially Payment
3.	Click on the payment order

**Issue Log:**
Traceback (most recent call last):
  File "/home/workspace/U14/U14.0/odoo/http.py", line 639, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/workspace/U14/U14.0/odoo/http.py", line 315, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: unsupported operand type(s) for +=: 'bool' and 'str'

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -89,7 +89,7 @@ class AccountMoveLine(models.Model):
         aplo = self.env["account.payment.line"]
         # default values for communication_type and communication
         communication_type = "normal"
-        communication = self.ref or self.name
+        communication = self.ref or self.name or ""
         # change these default values if move line is linked to an invoice
         if self.move_id.is_invoice():
             if (self.move_id.reference_type or "none") != "none":


pedrobaeza:
This is not possible, as https://github.com/OCA/bank-payment/pull/936/files#diff-c44811bc7189570a631438d2dca538756d8ad24b26ced88bfd5df3ab44b74f60R92 should give a value. Anyways, if you want to make sure, add there an ` or ""` part.

pedrobaeza (changes requested):
See my previous comment

pedrobaeza:
Now we need steps to reproduce the problem from the scratch. How a journal item doesn't have neither `name` nor `ref` set?

Vandan-OSI:
> Now we need steps to reproduce the problem from the scratch. How a journal item doesn't have neither `name` nor `ref` set?

**Step to Reproduce:**
1.  Invoicing  > Vendors > Bills
2.  Create Vendor Bill with Partially Payment
3.  Click on the payment order

I tested in Bank-Payment Run Bot here URL: http://oca-bank-payment-14-0-31d375aa9353.runboat.odoo-community.org/web#id=22&action=209&model=account.move&view_type=form&cids=&menu_id=107

image

pedrobaeza:
Why the partial payment influences this?

Vandan-OSI:
> Why the partial payment influences this?

It only Affected partial payment because there is One Method  **_get_linked_move_communication** (https://github.com/OCA/bank-payment/blob/4f86dad002675627cb8fa5340e64e5fe5a352095/account_payment_order/models/account_move_line.py#L51)  which is Only Collects Reversal moves and Partial payments So When We Do Full Payment Order of Bill at the time this Condition will be False So It Won't Send Traceback / Error.

pedrobaeza:
Thanks for the extra information. I don't see where this method is being called in current chain. Can you point me? Another interesting thing is to add a comment in the code about why doing something for later references, and also add a regression test.

Vandan-OSI:
A method called  in a current chain from this Line method https://github.com/OCA/bank-payment/blob/4f86dad002675627cb8fa5340e64e5fe5a352095/account_payment_order/models/account_move_line.py#L108

pedrobaeza:
OK, but this is still not answering the original question: why both `ref` and `name` are empty.

Vandan-OSI:
> OK, but this is still not answering the original question: why both `ref` and `name` are empty.

In Core, there is One Method https://github.com/odoo/odoo/blob/ecff2d13dc7739dc089e5851e5b424acf93f8852/addons/account/models/account_move.py#L944 in that Method another method https://github.com/odoo/odoo/blob/ecff2d13dc7739dc089e5851e5b424acf93f8852/addons/account/models/account_move.py#L1005 in that else Condition Name Field set based on payment_reference if is there otherwise it set '' https://github.com/odoo/odoo/blob/ecff2d13dc7739dc089e5851e5b424acf93f8852/addons/account/models/account_move.py#L1037

pedrobaeza:
So that confirms that is not False, but "". Anyways, putting this safeguard is not too much, so I stop being nitpicking.

pedrobaeza (approved):

murtuzasaleh (approved):
LGTM

rvalyi:
ping @dreispt @pedrobaeza @Vandan-OSI did you see the ocabot didn't picked its merge command? I didn't check this PR myslef so I won't risk the merge myself, but I let you know...

pedrobaeza:
The command was not correct (a missing `a`)

/ocabot merge patch

PR #937, opened 2022-06-30
[14.0][Add] Create module account_payment_mode_default_account

grindtildeath:
This module allows to define default receivable and payable accounts
on payment mode to override the account selected on the customer
when computing payment terms lines on invoices.


alexis-via (approved):
Useful for projects with a high-volume of transaction, where you need to separate receivable accounts per payment mode. Not useful for small-scale projects.

yankinmax (approved):

ivantodorovich (approved):

PR #939, opened 2022-07-13
[IMP][14.0]account_payment_purchase

Vandan-OSI:
**Steps to Reproduce:**
1) Create Purchase Order with that Partner who has Parent.
2) Check Banking Information (Banking information not populating for a partner who has Parent)

### DIFF
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -25,6 +25,10 @@ class PurchaseOrder(models.Model):
     @api.model
     def _get_default_supplier_partner_bank(self, partner):
         """This function is designed to be inherited"""
+        if partner.parent_id:
+            return (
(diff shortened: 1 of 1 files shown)


account_payment_purchase/models/purchase_order.py
```diff
@@ -25,6 +25,10 @@ class PurchaseOrder(models.Model):
     @api.model
     def _get_default_supplier_partner_bank(self, partner):
         """This function is designed to be inherited"""
+        if partner.parent_id:
+            return (
+                partner.parent_id.bank_ids and partner.parent_id.bank_ids[0].id or False
+            )
         return partner.bank_ids and partner.bank_ids[0].id or False
```
pedrobaeza, line 32:
```suggestion
        partner = partner.commercial_partner_id
        return partner.bank_ids[:1]
```

pedrobaeza (changes requested):

PR #943, opened 2022-07-20
Cp 81 cannot remove line of a debit order / payment order

duvenhage:
cannot remove line of a debit order / payment order


pedrobaeza:
Please do the pull request when things are in order in your side (no conflicts, no mix of modules, etc).

PR #948, opened 2022-08-10
[14.0][FIX] account_payment_order: small usability fix

alexis-via:
create_account_payment_line() is supposed to returned a form view for a single payment order and a tree view for multiple payment orders. Before this fix, it would return a tree view if you were paying more than 1 invoices while generating a single payment order, because the list result_payorder_ids would contain several time the same ID.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -53,7 +53,7 @@ class AccountMove(models.Model):
 
     def create_account_payment_line(self):
         apoo = self.env["account.payment.order"]
-        result_payorder_ids = []
+        result_payorder_ids = set()
         action_payment_type = "debit"
         for move in self:
             if move.state != "posted":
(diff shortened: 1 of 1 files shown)


account_payment_order/models/account_move.py
```diff
@@ -53,7 +53,7 @@ def get_account_payment_domain(self, payment_mode):
 
     def create_account_payment_line(self):
         apoo = self.env["account.payment.order"]
-        result_payorder_ids = []
+        result_payorder_ids = {}
```
pedrobaeza:
```suggestion
        result_payorder_ids = set()
```

account_payment_order/models/account_move.py
```diff
@@ -91,7 +91,7 @@ def create_account_payment_line(self):
                         move._prepare_new_payment_order(payment_mode)
                     )
                     new_payorder = True
-                result_payorder_ids.append(payorder.id)
+                result_payorder_ids[payorder.id] = True
```
pedrobaeza:
```suggestion
                result_payorder_ids.add(payorder.id)
```

alexis-via:
@pedrobaeza I took into account your suggestion

pedrobaeza (approved):

flotho (approved):
LGTM,

thanks

rvalyi (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to v15

PR #951, opened 2022-08-12
[15.0][FIX] account_payment_order: Fixing newId issue

### LINKED ISSUE #950: [FIX] account_payment_order
Issue in module --> Account Payment Order

Bank journal in the payment order is not populated correctly.

Affected versions: 15.0.1.0.0

Create payment mode, set Link to Bank Account to Fixed and select any journal.
Start creating new payment order and select above created payment mode.
Bank Journal is not filled in, nor it should be applied from the payment mode.
Bank journal should be populated whenever payment mode is changed.

tbanevicius:
This fix will make bank journal populate after payment mode on change.

fixes #950

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -246,7 +246,7 @@ class AccountPaymentOrder(models.Model):
     @api.onchange("payment_mode_id")
     def payment_mode_id_change(self):
         if len(self.allowed_journal_ids) == 1:
-            self.journal_id = self.allowed_journal_ids
+            self.journal_id = self.allowed_journal_ids._origin
         if self.payment_mode_id.default_date_prefered:
             self.date_prefered = self.payment_mode_id.default_date_prefered
 
--- a/account_payment_order/tests/test_payment_mode.py
+++ b/account_payment_order/tests/test_payment_mode.py
@@ -3,7 +3,7 @@
 
 from unittest.mock import patch
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import Form, TransactionCase
 
 from odoo.addons.account.models.account_payment_method import AccountPaymentMethod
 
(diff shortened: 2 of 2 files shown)


aisopuro:
The diff for this PR seems quite large. Are all of these changes necessary for this fix? It seems you might not have branched from `15.0`?

I'm not a maintainer, but it feels like it would be better if you limited this PR to only the changes necessary to fix #950 . Makes it easier for reviewers to go through the code that is relevant to the fix.

tbanevicius:
> The diff for this PR seems quite large. Are all of these changes necessary for this fix? It seems you might not have branched from `15.0`?
> 
> I'm not a maintainer, but it feels like it would be better if you limited this PR to only the changes necessary to fix #950 . Makes it easier for reviewers to go through the code that is relevant to the fix.

@aisopuro  Thanks for noticing, I fixed commits.

account_payment_order/models/account_payment_order.py
```diff
@@ -246,7 +246,7 @@ def create(self, vals):
     @api.onchange("payment_mode_id")
     def payment_mode_id_change(self):
         if len(self.allowed_journal_ids) == 1:
-            self.journal_id = self.allowed_journal_ids
+            self.journal_id = self.allowed_journal_ids._origin
```
pedrobaeza, line 249: Is this not working?

```suggestion
            self.journal_id = self.allowed_journal_ids.id
```

My doubt about your solution is what happens if you change the payment mode to another.

rafaelbn (commented):
@tbanevicius thank you very much for taking care of this! 😄 

In ny review I don't see this PR fix it.

1. If payment mode is `Link to Bank Account | Variable`: How should be populated?
2. As you can check in runboat in this PR it is sill not populated

GIF 08-10-2022 11-56-07

pedrobaeza:
That's because there are 2 journals linked to that payment mode, isn't it?

rafaelbn:
> That's because there are 2 journals linked to that payment mode, isn't it?

It is. With only 1 journal linked it is populated.

That's why I said:

> If payment mode is Link to Bank Account | Variable: How should be populated?

Just informing in order to decide which is the most common and desired functionality 😄

pedrobaeza:
I don't think you should popular any bank account.

PR #955, opened 2022-08-29
[14.0][FIX] account_payment_purchase: Don't propagate empty payment mode

pedrobaeza:

- Have a partner without payment mode.
- Create a PO with such partner.
- No payment mode is filled.
- Now fill the payment mode in the partner.
- Create the invoice for the PO.

Current behavior:

The invoice has empty payment mode.

Expected behavior:

The invoice has the partner payment mode. Someone may think that having no payment mode in the PO may prevail over the partner's payment mode, or even their flows may consist in empyting the payment mode in the PO for not binding it with anything yet, but that strategy has more holes than the one implemented here, as the flow presented proves.

You can then use another one like having an extra payment mode "Undetermined" or similar for doing such classification as a more resilient strategy.

It includes a regression test that fails before the change and now is correct.

Forward-port of #953 

@Tecnativa TT38608

### DIFF
--- a/account_payment_purchase/models/__init__.py
+++ b/account_payment_purchase/models/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from . import account_invoice, purchase_order
+from . import account_move
+from . import purchase_order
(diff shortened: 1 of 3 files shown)


victoralmau (approved):

LudLaf (approved):
Tested on runboat. Working fine.

PR #967, opened 2022-10-07
[13.0][IMP] account_payment_partner: Make payment mode editable

pedrobaeza:
This field was editable in previous version before invoice > move refactoring, and it's logic to allow to change payment mode once the invoice has been posted without the need of resetting it to draft.

Thus, the field has been changed to computed writable field, taking care of the consequences at model level (compute) and view level (add the field in views + proper attrs).

This commits adds tracking=True to the payment mode field as well to be aware when and who the payment mode is changed in the invoice, no matter if directly changed in draft, or through the change on the journal item.

@Tecnativa TT39850

### DIFF
--- a/account_payment_partner/models/account_move_line.py
+++ b/account_payment_partner/models/account_move_line.py
@@ -13,9 +13,10 @@ class AccountMoveLine(models.Model):
         store=True,
         ondelete="restrict",
         index=True,
+        readonly=False,
     )
 
-    @api.depends("move_id.payment_mode_id")
+    @api.depends("move_id", "move_id.payment_mode_id")
     def _compute_payment_mode(self):
         for line in self:
             if line.move_id.is_invoice() and line.account_internal_type in (
@@ -24,4 +25,19 @@ class AccountMoveLine(models.Model):
             ):
                 line.payment_mode_id = line.move_id.payment_mode_id
             else:
-                line.payment_mode_id = False
+                line.payment_mode_id = line.payment_mode_id  # HACK not needed for v14+
+
+    def write(self, vals):
+        """Propagate up to the move the payment mode if applies."""
+        if "payment_mode_id" in vals:
+            for record in self:
+                move = (
+                    self.env["account.move"].browse(vals.get("move_id", 0))
+                    or record.move_id
+                )
+                if (
+                    move.payment_mode_id.id != vals["payment_mode_id"]
+                    and move.is_invoice()
+                ):
+                    move.payment_mode_id = vals["payment_mode_id"]
+        return super().write(vals)
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -22,6 +22,7 @@ class AccountMove(models.Model):
         store=True,
         ondelete="restrict",
         readonly=False,
+        tracking=True,
     )
     bank_account_required = fields.Boolean(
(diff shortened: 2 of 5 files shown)


mileo (approved):
LGTM

image

marcelsavegnago:
cc @douglascstd

marcelsavegnago (approved):
LGTM

rafaelbn (commented):
> and it's logic to allow to change payment mode once the invoice has been posted without the need of resetting it to draft

Depends.

I've tested and runboat make what your PR wants but take care of:

- It allow you to change payment mode even you have a confirmed payment order

Doubt, if you want to do this:

- Why allow from journal item and not from invoice directly?

This change is not tracked anywhere, so:

- You have an invoice with SEPA Direct debit
- Anyone change it in a journal item
- The payment mode has change in the invoice and there is not any clue about who or when change that
- The payment order was confirmed and done 1 week ago after this change

imagen


Thank you!

pedrobaeza:
> It allow you to change payment mode even you have a confirmed payment order

As we are doing the change in this module, we don't know yet about payment orders. I have changed it for a good compromise right now in this module: it's readonly if the item is reconciled. This covers all the done payment orders. For non done payment orders, it can be debatable if it should be restricted, as you can select items to be included in a payment order even if the payment mode doesn't match (through the option on the wizard), so it's not incoherent at all to leave it this way.

> Why allow from journal item and not from invoice directly?

Indeed it can be put to be modified on the header, but then you will need to handle the non propagation of the value if there are several AR/AP lines to the ones that are reconciled. I won't handle it in this PR, as the aim is to have it editable at journal item view level, and also on the tree view of `account_due_list` (later PR coming).

> This change is not tracked anywhere

I have added `tracking=True` to the field at `account.move` level to tackle this.

carlosdauden (approved):
I am not sure what the correct behavior would be if we had an invoice with several due dates and we wanted to change the payment method in only one of them...

account_payment_partner/models/account_move_line.py
```diff
@@ -24,4 +25,19 @@ def _compute_payment_mode(self):
             ):
                 line.payment_mode_id = line.move_id.payment_mode_id
             else:
-                line.payment_mode_id = False
+                line.payment_mode_id = line.payment_mode_id  # HACK not needed for v14+
+
+    def write(self, vals):
```
adrienpeiffer, line 30: Wouldn't it be better to put this in a inverse method?

beneditomonteiro:
Hello Guys,

I made the test, and it was updated the payment mode after you edit the field and save it.
Make sure that this changing doesn´t affect the journals entries, because this changing at this level, is not allowed in brazilian accounting system. (for a certified system)
As far as I have checked this is only affecting the invoice details, which is ok for me, even seems weird to me to make it through journal items, but, if this a system limitation, I'm ok with that.

Good work guys!

Benedito Monteiro 
Brazilian Certified Accountant 
5243-O/4 SP

etobella (approved):
Makes sense to me

rafaelbn:
Any blocking point here? This looks like to me merged

pedrobaeza:
I will try these days to make the payment mode in header also editable. If not done this week, we proceed to merge it.

pedrobaeza:
I have added the editability on the invoice header taking as criteria if there's no reconciliation made. The final result is satisfactory with no so much overhead, so proceeding finally to merge:

/ocabot merge minor

Issue #966, opened 2022-10-06
[14.0] account_payment_order: Performance problem when posting + reconciling transit moves


pedrobaeza:
Steps to reproduce:

- Have a big bunch of invoices (200+).
- Include all of them in a payment order.
- Configure the payment mode to post transit moves.
- Confirm the order, generate file, and click on "File successfully uploaded".

The operation will take a lot of time, or even not being able to finish depending on the timeouts.

Tracing down the problem, it's a proportional performance degradation on each loop of:

https://github.com/OCA/bank-payment/blob/485abeb889d344f2dda52adc71928967629ac71e/account_payment_order/models/bank_payment_line.py#L141

![Selección_043](https://user-images.githubusercontent.com/7165771/194406219-ee9a761b-71ab-4288-82a1-5dbc8f658454.png)

And it's due to a call over _compute_amount:

https://github.com/odoo/odoo/blob/062fb69f346873fe05ebb5d429ab3535fd1df5cc/addons/account/models/account_move.py#L1406

coming from:

https://github.com/odoo/odoo/blob/062fb69f346873fe05ebb5d429ab3535fd1df5cc/addons/account/models/account_move.py#L4997

that contains each time current invoices plus the old ones.

```
_compute_amount account.move(204346, 204864)
_compute_amount account.move(204320, 204346, 204864, 204319)
_compute_amount account.move(204318, 204864, 204320, 204346, 204319)
_compute_amount account.move(204317, 204864, 204320, 204346, 204318, 204319)
_compute_amount account.move(204315, 204864, 204320, 204346, 204317, 204318, 204319)
_compute_amount account.move(204313, 204864, 204320, 204346, 204315, 204317, 204318, 204319)
_compute_amount account.move(204312, 204320, 204864, 204313, 204346, 204315, 204317, 204318, 204319)
_compute_amount account.move(203154, 204864, 204320, 204312, 204313, 204346, 204315, 204317, 204318, 204319)
_compute_amount account.move(203153, 204864, 204320, 203154, 204312, 204313, 204346, 204315, 204317, 204318, 204319)
_compute_amount account.move(203152, 204320, 204864, 203153, 203154, 204312, 204313, 204346, 204315, 204317, 204318, 204319)
_compute_amount account.move(203151, 204864, 204320, 203152, 203153, 203154, 204312, 204313, 204346, 204315, 204317, 204318, 204319)
[...]

PR #979, opened 2022-11-06
[14.0][REF] account_payment_*: Use native payment object

pedrobaeza:
The previous approach creates manually the journal entries and does all the hard work, plus not being 100% compatible with the bank statement reconciliation widget (requiring a patch on OCB to see blue lines).

That decision made sense on the moment it was done (v9), where the native payment model (account.payment) was very limited, and wasn't able to store all the needed information for the bank transaction.
    
Now that the limitations are gone, we can get rid off this extra model, and generate instead `account.payment` records, using both the native model + methods to perform the same operations.
    
This serves also to workaround the problem found in #966.
    
All the code, views and tests of main module have been adapted to this new approach

@Tecnativa TT39832


pedrobaeza:
@alexis-via @sbidoul @HaraldPanten @ValentinVinagre @carlosdauden here it's the long awaited refactoring to make use of the native `account.payment` object for handling the bank lines and the move generation. The stats are good: +205 lines added against −810 lines removed.

See individual commits in order to have better comprehension of the process.

Working on the migration scripts now.

carlosdauden (changes requested):
At some point we should extend the account.move method `def _get_invoice_in_payment_state(self):` to return 'in_payment' to activate the payment status "In process" at the invoices.
https://github.com/odoo/odoo/blob/e337a7b4b0da970041ea26d6fdc6e2d39eb3e5be/addons/account/models/account_move.py#L1389

pedrobaeza:
Indeed, but that's the next step not included in this PR.

carlosdauden (commented):
I have reviewed and tested and everything is correct, I am waiting for the migration scripts to approve

HaraldPanten:
First tests seem to be working fine. But I'll need more time to check it properly.

Have you tested it in "semi-real" use cases?

carlosdauden (commented):
We should synchronize PR merge with changes in these modules so as not to cause malfunctions:

https://github.com/OCA/account-payment/blob/9f5e77607690da8f275ccec6270a2e8aad652145/account_payment_return_import_iso20022/models/payment_return.py#L16

https://github.com/OCA/account-reconcile/blob/2b77e6078785a08890205fa94dd7682eec102466/account_reconcile_payment_order/models/account_reconciliation_widget.py#L25-L33

pedrobaeza:
Migration scripts are completed and tested on a real environment with all the possible cases, as well as the features of the module. We have an slower payment order confirmation due to the full creation of the account.payment + account.move, but faster times later on "File successfully uploaded" phase.

pedrobaeza:
> We should synchronize PR merge with changes in these modules so as not to cause malfunctions:

Done on:

https://github.com/OCA/account-payment/pull/555
https://github.com/OCA/account-reconcile/pull/505

gdgellatly:
@dannyadair Can we review this one, full functional with current work.

carlosdauden:
Can you rebase?

pedrobaeza:
Rebased

carlosdauden (approved):

pedrobaeza:
People, please review this for switching definitively to a more standard system and go to the next step to have in-payment states in invoices, which will be a very nice thing for our customers.

legalsylvain:
Thanks a lot for this refactoring pedro.

sbidoul:
Thanks for doing this work! I generally agree with the approach of replacing bank.payment.line with the native Odoo model.

I think there should be a major version bump on all modules.

Also it would be nice if modules that need changing following to this major breaking change to `account_payment_order` get a `setup.py` like this:

```
import setuptools

setuptools.setup(
    setup_requires=['setuptools-odoo'],
    odoo_addon={
        'depends_override': {
            'account_payment_order': 'odoo14-addon-account_payment_order>=14.0.2.0.0',
        },
    },
)
```

pedrobaeza:
Proceeding then with the merge with a major bump:

/ocabot merge major

HaraldPanten:
Thanks for this work!

LoisRForgeFlow:
Isn't this a bit of a risky refactor to push in a stable version?

It is affecting other extensions, like the ones at https://github.com/OCA/l10n-usa (an likely other localizations)

pedrobaeza:
Yes, it's not the ideal thing, but we are very late on this, and problems like #966 makes this even more needed. What are the problems? Check other adapted modules or tell me and I help you.

qgroulard:
Great work, thank you !

Looks like `move_option` got lost in the process.
The field is still on `account.payment.mode` but isn't used anywhere anymore.
Any plan on restoring this feature ?

pedrobaeza:
I'm removing that leftovers in #1015, but there can't be equivalent options, as now each `account.payment` generates one move, and you can't group several payments from different partners in one record.

hhgabelgaard:
Great work, thank you !
Has  this change been forward ported to 15.0?
Or is there any plans for this?
The upgrade path is kind of broken now it seems ?

pedrobaeza:
Yes, indeed. I want to do it for 15.0, but no time for now. Do you want to do it? I will review it. It should be pretty straight-forward.

account_banking_sepa_direct_debit/models/account_payment_order.py
```diff
@@ -281,18 +278,19 @@ def generated2uploaded(self):
             to_expire_mandates = abmo.browse([])
             first_mandates = abmo.browse([])
             all_mandates = abmo.browse([])
-            for bline in order.bank_line_ids:
-                if bline.mandate_id in all_mandates:
+            for payment in order.payment_ids:
```
acsonefho, line 281: @pedrobaeza This field `payment_ids` doesn't exist on `account.payment.order` model.
Or maybe there is a missing dependency?

pedrobaeza: This PR adds such field: https://github.com/OCA/bank-payment/pull/979/files#diff-e1bae20141deca080d6db095e9874359ec1ed9d92f0b4d46aedc9f3ecea71cf0R80

PR #984, opened 2022-11-17
[13.0][FIX] account_payment_order_notification: Create notification with the correct partner (invoice partner).

victoralmau:
Create notification with the correct partner (invoice partner).

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40516

### DIFF
--- a/account_payment_order_notification/wizard/wizard_account_payment_order_notification.py
+++ b/account_payment_order_notification/wizard/wizard_account_payment_order_notification.py
@@ -30,7 +30,11 @@ class WizardAccountPaymentOrderNotification(models.TransientModel):
         )
         if po:
             line_ids = []
-            for partner in po.payment_line_ids.mapped("partner_id"):
+            # We need to transfer invoice partner so that the email is sent to the
+            # correct partner (not parent)
+            for partner in po.payment_line_ids.mapped(
+                lambda x: x.move_line_id.move_id.partner_id or x.partner_id
+            ):
                 line_ids += [
                     (
                         0,
@@ -59,7 +63,7 @@ class WizardAccountPaymentOrderNotification(models.TransientModel):
         notifications = []
         for item in self.line_ids.filtered("to_send"):
             payment_line_ids = self.order_id.payment_line_ids.filtered(
-                lambda x: x.partner_id == item.partner_id
+                lambda x: x.partner_id == item.partner_id.commercial_partner_id
             )
             data = {
                 "partner_id": item.partner_id.id,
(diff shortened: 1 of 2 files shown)


account_payment_order_notification/wizard/wizard_account_payment_order_notification.py
```diff
@@ -30,7 +30,11 @@ def default_get(self, fields):
         )
         if po:
             line_ids = []
-            for partner in po.payment_line_ids.mapped("partner_id"):
+            # We need to transfer invoice partner so that the email is sent to the
+            # correct partner (not parent)
+            for partner in po.payment_line_ids.mapped(
+                "move_line_id.move_id.partner_id"
```
pedrobaeza: This is not correct: The move line may not be an invoice and don't have `partner_id`. You have to select the filled one (through a lambda in the mapped).

victoralmau: Changes done (I believe both use cases are taken into account).

account_payment_order_notification/wizard/wizard_account_payment_order_notification.py
```diff
@@ -59,7 +63,8 @@ def action_process(self):
         notifications = []
         for item in self.line_ids.filtered("to_send"):
             payment_line_ids = self.order_id.payment_line_ids.filtered(
-                lambda x: x.partner_id == item.partner_id
+                lambda x: (x.move_line_id.move_id.partner_id or x.partner_id)
+                == item.partner_id
             )
             data = {
                 "partner_id": item.partner_id.id,
```
pedrobaeza, line 69: This one should change as well, isn't it? Can't you avoid the `or` each time?

victoralmau: IMO previously is correct, although a bit confusing, it is changed.

pedrobaeza: But you are not getting here the invoice partner...

victoralmau: `line_ids` records already have the correct `partner_id` (previously set in `default_get()`)

pedrobaeza: OK, got it.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking the fix:

/ocabot merge patch

PR #993, opened 2022-12-06
[13.0][FIX] [URGENT] generate pain address block that is supported by german banks

luc-demeyer:


luc-demeyer:
In the last two weeks I encountered the same issue with the OCA sepa payment order with three different German banks:
Sparkasse Neuss, Kreissparkasse Göppingen and ING.
All these banks have stopped supporting pain.001.003.03 in Germany in favour of pain.001.001.03.

But in all three cases they do not support the full pain.001.001.03.xsd scheme : 

In the <PstlAdr> block you can specify zip and city via <PstCd> and <TwnNm> or alternatively via a second <AdrLine>.

The <PstCd> and <TwnNm> variant results in failed payment orders by all of these three banks hence German customers running with the OCA module have a serious problem !

I generated the same pain file via the Odoo Enterprise account_sepa module and that one uses a second <AdrLine> to represent the exact same info. Uploading this file to the concerned banks fixes the issue.

This fix now generates the second  <AdrLine> in stead of <PstCd> and <TwnNm>.
Imho this fix needs a fast track approach for review, merge followed cherry-pick to Odoo 14 and Odoo 15.

fkantelberg:
@luc-demeyer Specifications I find state that the you can't mixed AdrLine and the usage of StrtNm/PostCd/TwnNm. Currently the versions are using AdrLine for the street and I think the problem is there. Hence I would favor the change in the AdrLine -> StrNm and let the PostCd/TwnNm be like currently.

I also find that the AdrLine element is no longer supported after Nov 2022 and TwnNm and Ctry should be used until 2025 anyways.

fkantelberg:
@luc-demeyer Small update from our side: We talked to our bank and they only support `pain.001.001.09` which means we have to add support for that format too in the next days.

In the [specification](https://www.ebics.de/de/datenformate) you can read that AdrLine usage is not recommended hence I would suggest to revert it and delete the handling of the street because proper handling of the street is impossible hard but street is currently optional. I also recommend that because I found different banks (some swiss ones) who claim, that TwnNm will be made required which means that the move towards AdrLine is the opposite direction and the format may break again soon.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -480,7 +459,31 @@ def generate_address_block(self, parent_node, partner, gen_args):
...
+                    val = ""
+                if partner.city:
+                    val += " " + self._prepare_field(
+                        "city",
+                        "partner.city",
+                        {"partner": partner},
+                        70,
+                        gen_args=gen_args,
+                    )
+                adrline2.text = val
```
fkantelberg: You might want to use a list here to prevent a leading space:

```
val = []
if partner.zip:
    val.append(...)
if partner.city:
    val.append(...)
adrline2.text = " ".join(val)
```

fkantelberg (approved):
Just a small thing but it already produces valid pain.001.001.03 and pain.008.001.*

pedrobaeza (approved):
OK, let's merge it: /ocabot merge patch

PR #999, opened 2022-12-13
[14.0][FIX] account_payment_order_notification: Fix template body.

victoralmau:
Avoid re-formatting the content causing an error when rendering the template.

@Tecnativa TT40795


account_payment_order_notification/data/mail_template_data.xml
```diff
@@ -30,14 +30,18 @@
...
-        % for payment_line in object.payment_line_ids:
-            <tr>
-                <td>${payment_line.communication}</td>
-                <td>${payment_line.date}</td>
-                <td>${format_amount(payment_line.amount_currency, payment_line.currency_id)}</td>
-            </tr>
-        % endfor
     </table>
+    % for payment_line in object.payment_line_ids:
+        <table>
```
pedrobaeza: Why a new table? The previous problem seems to be the absence of the `<tbody>` element?

victoralmau: I would have to check specifically why (html formatting "breaks" it).

I have found a "quick" example of odoo core in which there is also a similar error: https://github.com/odoo/odoo/blob/14.0/addons/gamification/data/goal_base.xml#L26-L226, however, in `sale` there is this example https://github.com/odoo/odoo/blob/14.0/addons/sale/data/mail_data.xml#L119-L294 in which it uses several tables.

pedrobaeza: But putting `tbody` should solve it.

victoralmau: I think it is best to add several examples to better explain what happens when you install the module (and therefore, when you try to send an email with that template).

KO > **Tbody option**:
ejemplo-tbody
ejemplo-tbody-template1
ejemplo-tbody-template2


OK > **Table option**:
ejemplo-table
ejemplo-table-template1
ejemplo-table-template2

pedrobaeza: The second one is not OK, as the width of the header is not going to be adapted to the content of the rest. Anyway, ignore the editor. The important thing is that the content gets to the mail correctly.

account_payment_order_notification/data/mail_template_data.xml
```diff
@@ -30,13 +30,15 @@
...
         </thead>
-        % for payment_line in object.payment_line_ids:
-            <tr>
-                <td>${payment_line.communication}</td>
-                <td>${payment_line.date}</td>
-                <td>${format_amount(payment_line.amount_currency, payment_line.currency_id)}</td>
-            </tr>
-        % endfor
+        <tbody>
+            % for payment_line in object.payment_line_ids:
```
pedrobaeza: No need of extra indentation that makes bigger the diff

pedrobaeza (approved):

ernestotejeda (approved):

Issue #964, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] account_banking_mandate - By @Reyes4711-S73 - #1028
- [x] account_banking_pain_base - By @pedrobaeza - #1087
- [x] account_banking_sepa_credit_transfer - By @dzungtran89 - #990
- [x] account_banking_sepa_direct_debit - By @Reyes4711-S73 - #1029
- [x] account_payment_mode - By @ThomasBinsfeld - #968
- [x] account_payment_order - By @wpichler / @pedrobaeza - #1056
- [x] account_payment_order_grouped_output - By @ferran-S73 - #1044
- [x] account_payment_order_notification - By @Yadier-Tecnativa - #1095
- [x] account_payment_order_return - By @almumu - #1359
- [ ] account_payment_order_transfer_journal - By @luc-demeyer - #1108
- [x] account_payment_partner - By @ThomasBinsfeld - #969
- [x] account_payment_purchase - By @ramiadavid - #1004
- [x] account_payment_purchase_stock - By @SodexisTeam - #1063
- [x] account_payment_sale - By @atchuthan - #982

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

dzungtran89:
Hello, I will migrate `account_banking_sepa_credit_transfer`

ramiadavid:
Hi @OCA/banking-maintainers
Im working on `account_payment_purchase` #1004

JavierIniesta:
Hi @sbidoul 

I have seen that the `account_payment_sale` addon is already migrated but it is not checked in the list.

fgarcia-humanoide:
[16.0][MIG] account_banking_sepa_direct_debit: Migration to 16.0 #1029

sepa direct debit 16 is being migrated but do not appear in the list as being in process of migration

ferran-S73:
account_banking_mandate_sale https://github.com/OCA/bank-payment/pull/1008

vincent-hatakeyama:
account_payment_order_tier_validation #1161

This was not migrated to 15.0 so it does not appear in the list.

Tisho99:
I'm migrating account_banking_mandate_contact #1233

PR #1004, opened 2022-12-20
[16.0][MIG] account_payment_purchase: Migration to 16.0

ramiadavid:
#964


flotho (approved):
code review,
tested on runboat,
LGTM
Thanks @alexis-via

account_payment_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    @api.model
+    def _get_default_supplier_partner_bank(self, partner):
+        """This function is designed to be inherited"""
+        return partner.bank_ids and partner.bank_ids[0].id or False
+
+    @api.onchange("partner_id", "company_id")
+    def onchange_partner_id(self):
+        ret = super(PurchaseOrder, self).onchange_partner_id()
```
RodrigoBM: Apply PEP 3135 (28-Apr-2007) to "super()."

account_payment_purchase/tests/test_account_payment_purchase.py
```diff
@@ -0,0 +1,136 @@
...
+
+from odoo import fields
+from odoo.tests import TransactionCase, tagged
+
+
+@tagged("-at_install", "post_install")
+class TestAccountPaymentPurchase(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestAccountPaymentPurchase, cls).setUpClass()
```
RodrigoBM: Apply PEP 3135 (28-Apr-2007) to "super()."

account_payment_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,41 @@
...
+    payment_mode_id = fields.Many2one(
+        comodel_name="account.payment.mode",
+        string="Payment Mode",
+        domain="[('payment_type', '=', 'outbound')]",
+    )
+
+    @api.model
+    def _get_default_supplier_partner_bank(self, partner):
+        """This function is designed to be inherited"""
+        return partner.bank_ids and partner.bank_ids[0].id or False
```
RodrigoBM: Reviewing the changes in odoo v16. I have seen that in the bank accounts, you can mark if the account is to send money (allow_out_payment), but I have not seen any logic restrictions with this field.

But I think it is a good idea to sort the bank accounts by field allow_out_payment.

`return partner.bank_ids and partner.bank_ids.sorted(lambda bank: not bank.allow_out_payment)[0].id or False`

And the first bank account you have is allowed to send money.

what do you think? @ramiadavid @flotho

account_payment_purchase/tests/test_account_payment_purchase.py
ramiadavid: Done

account_payment_purchase/models/purchase_order.py
ramiadavid: Done

account_payment_purchase/models/purchase_order.py
ramiadavid: Done

account_payment_purchase/models/purchase_order.py
ramiadavid: I've already done it, I think it's a good idea

RodrigoBM (approved):
LGTM.

RodrigoBM:
@etobella can you merge it?

account_payment_purchase/tests/test_account_payment_purchase.py
```diff
@@ -84,8 +84,8 @@ def test_purchase_order_invoicing(self):
         invoice = self.env["account.move"].create(
             {"partner_id": self.partner.id, "move_type": "in_invoice"}
         )
-        with Form(invoice) as inv:
-            inv.purchase_id = self.purchase
+        invoice.purchase_id = self.purchase
```
etobella, line 88: Why are you removing the form? It was done this way in order to keep the same logic done by odoo interface

ramiadavid: @etobella AssertionError: can't write on invisible field purchase_id

StephaneMangin (approved):

account_payment_purchase/tests/test_account_payment_purchase.py
pedrobaeza: But with this, you are not testing the real thing, as onchanges are not triggered.

simahawk: Having a quick look at how is implemented in core: it sucks. It seems to go against the principle of not having business logic in onchanges. Unfortunately, if you don't set `force_save="1"` on the field it does not work and - BTW - the field is emptied at the end of the onchange.
Also, if you look at `_set_purchase_orders` they call this onchange explicitly :facepalm: 
Finally, there's no test on their onchange on purchase_id, likely because they cannot test it :smile: 

However, if you want to keep it like this you should call `_onchange_partner_id`  too IMO.

account_payment_purchase/tests/test_account_payment_purchase.py
```diff
@@ -0,0 +1,136 @@
...
+from odoo import fields
+from odoo.tests import TransactionCase, tagged
+
+
+@tagged("-at_install", "post_install")
+class TestAccountPaymentPurchase(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.journal = cls.env["account.journal"].create(
```
simahawk, line 15:
```suggestion
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
        cls.journal = cls.env["account.journal"].create(
```

account_payment_purchase/tests/test_account_payment_purchase.py
ramiadavid: @simahawk

If I'm not mistaken, the function that is called when the purchase_id field is modified is _onchange_purchase_auto_complete

https://github.com/odoo/odoo/blob/08cf96d9ba169287a7e8d7dece4ac0110fa01742/addons/purchase/models/account_invoice.py#L33:L34

And this is already being called in the test, exactly the same as in the rest of the tests that I have not modified.

https://github.com/OCA/bank-payment/blob/1102ba6ed053e3cbe708512d8e71d0535aad0bc0/account_payment_purchase/tests/test_account_payment_purchase.py#L88

Isn't that correct?

astirpe: I just faced the same issue while porting module `account_move_line_purchase_info` to V16. I think that this part of test may be written the same way as in standard Odoo:
https://github.com/odoo/odoo/blob/16.0/addons/purchase/tests/test_purchase_order_report.py#L76-L79

Something like this:

```python
        with Form(invoice) as inv:
            inv.purchase_vendor_bill_id = self.env['purchase.bill.union'].browse(-self.purchase.id)
```

StephaneMangin:
cc @ramiadavid  Any update about the reviews ?

account_payment_purchase/tests/test_account_payment_purchase.py
ramiadavid: this done

HaraldPanten:
@ramiadavid is this PR ready to review? THX

ramiadavid:
@HaraldPanten I think yes

HaraldPanten:
@ramiadavid Could you rebase, please? I'll test it in runboat. THX!

ramiadavid:
@HaraldPanten done

HaraldPanten (approved):
THX David,

Functional review LGTM.

@pedrobaeza GO?

PR #1005, opened 2022-12-20
[13.0][FIX] account_banking_pain_base: Proper condition for AdrLine

pedrobaeza:

- Have a partner with no zip code and no city.
- Include it in a payment order.
- Generate the SEPA XML file.

Current behavior: The file couldn't be generated because there's an empty <AdrLine> element.

We put properly parenthesis in the conditions for this to no happen.

@Tecnativa TT40885

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -462,8 +462,7 @@ class AccountPaymentOrder(models.Model):
             if (
                 gen_args.get("pain_flavor").startswith("pain.001.001.")
                 or gen_args.get("pain_flavor").startswith("pain.008.001.")
-                and (partner.zip or partner.city)
-            ):
+            ) and (partner.zip or partner.city):
                 adrline2 = etree.SubElement(postal_address, "AdrLine")
                 val = []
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@luc-demeyer this one is caused by your last commit. Fast-tracking this:

/ocabot merge patch

Another thing: is it correct to send 2 `<AdrLine>` blocks?

luc-demeyer:
@pedrobaeza 
thanks a lot.
Can you also cherry-pick to 14.0 and 15.0 ?

pedrobaeza:
Yes, I will. Can you answer my other question?

luc-demeyer:
@pedrobaeza 
>>> Another thing: is it correct to send 2 <AdrLine> blocks?
Yes.

pain.001.001.003.xsd:
<xs:element maxOccurs="7" minOccurs="0" name="AdrLine" type="Max70Text"/>

German subset pain.001.001.03_GBIC_3.xsd:
xs:element name="AdrLine" type="Max70Text" minOccurs="0" maxOccurs="2"/>

pedrobaeza:
OK, thanks for the confirmation.

PR #1008, opened 2022-12-29
[MIG] account_banking_mandate_sale: 16.0

ferran-S73:


CasVissers-360ERP:
@ferran-S73 I get two warnings:
The model odoo.addons.account_banking_mandate.models.account_banking_mandate is not overriding the create method in batch 
The model odoo.addons.account_banking_mandate.models.account_move is not overriding the create method in batch

ferran-S73:
> @ferran-S73 I get two warnings: The model odoo.addons.account_banking_mandate.models.account_banking_mandate is not overriding the create method in batch The model odoo.addons.account_banking_mandate.models.account_move is not overriding the create method in batch

Hi @CasVissers-360ERP that seems weird to me since this module isn't inheriting any of those create methods, where are you getting the warnings?

account_banking_mandate_sale/oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+sale-workflow
```
astirpe: This file `oca_dependencies.txt` should be not placed inside the module. Use the one already existing in the repo https://github.com/OCA/bank-payment/tree/16.0

ferran-S73: You're right, I don't know who put it there it comes from a previous version. I'll remove it

astirpe (approved):

alexis-via:
I improved this migration in this other PR https://github.com/OCA/bank-payment/pull/1100

PR #1009, opened 2023-01-02
[16.0][FIX] account_payment_partner: fix error in _reverse_moves

ramiadavid:


account_payment_partner/models/account_move.py
```diff
@@ -94,6 +94,9 @@ def _compute_partner_bank_id(self):
         return res
 
     def _reverse_moves(self, default_values_list=None, cancel=False):
+        if not default_values_list:
+            default_values_list = [{} for _ in self]
+
```
pedrobaeza: Don't add empty lines.

pedrobaeza (commented):
Can you provide steps to reproduce the problem? Why it would get this method an empty `default_values_list` content?

ramiadavid:
Hi @pedrobaeza 

An error occurs when running tests on this file: https://github.com/OCA/l10n-spain/blob/483e5c7adb3e34948067df6aab7b9d779d888eb2/l10n_es_aeat/tests/test_l10n_es_aeat_taxinfo.py#L26

All the official addons that include this function have that code that I have added.
For example:
https://github.com/odoo/enterprise/blob/1d4432f6394d68f2e2d37d98babece555ae1c9ff/account_asset/models/account_move.py#L122
https://github.com/odoo/enterprise/blob/1d4432f6394d68f2e2d37d98babece555ae1c9ff/partner_commission/models/account_move.py#L150

account_payment_partner/models/account_move.py
ramiadavid: done

pedrobaeza (approved):
I don't have access to the enterprise repo, but I trust you, so let's merge this:

/ocabot merge patch

Thanks for the patch

PR #1020, opened 2023-01-18
[14.0][IMP] account_payment_partner: Make payment mode editable

pedrobaeza:
Forward-port of #967 to 14.0

This field was editable in previous version before invoice > move refactoring, and it's logic to allow to change payment mode once the invoice has been posted without the need of resetting it to draft.

Thus, the field has been changed to computed writable field, taking care of the consequences at model level (compute) and view level (add the field in views + proper attrs).

This commit adds tracking=True to the payment mode field as well to be aware when and who the payment mode is changed in the invoice, no matter if directly changed in draft, or through the change on the journal item.

@Tecnativa TT39850

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -22,6 +22,7 @@ class AccountMove(models.Model):
         ondelete="restrict",
         readonly=False,
         check_company=True,
+        tracking=True,
     )
     bank_account_required = fields.Boolean(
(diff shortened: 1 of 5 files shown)


carlosdauden (approved):

pedrobaeza:
Merging as already merged in 13.0, for continuing with the fw-port chain:

/ocabot merge minor

PR #1021, opened 2023-01-19
[14.0] account_payment_partner: Add pre_init_hook to add computed columns

dannyadair:
Add columns to avoid Memory error on an existing Odoo instance with lots of data.

The payment_mode_id fields are introduced by this module and computed only from each other or the also newly introduced supplier_payment_mode_id and customer_payment_mode_id on res.partner, so they can stay NULL, nothing to compute.

### DIFF
--- a/account_payment_partner/__init__.py
+++ b/account_payment_partner/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from .hooks import pre_init_hook
 from . import models
 from . import reports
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

dannyadair:
@pedrobaeza Thank you very much for the quick review!

dalonsod (approved):
👍

pedrobaeza:
This should be fw-port it to upper versions.

dannyadair:
> This should be fw-port it to upper versions.

Submitted #1046

PR #1022, opened 2023-01-20
[14.0] account_payment_order: Add pre_init_hook to add computed columns

dannyadair:
Add columns to avoid MemoryError on an existing Odoo instance with lots of data.

payment_order_ok on account.move is computed from payment_order_ok on account.payment.mode which are both introduced by this module, so nothing to compute.
(see AccountMove._compute_payment_order_ok() in models/account_move.py)

partner_bank_id on account.move.line requires payment_order_ok to be True which it won't be as it's newly introduced - again nothing to compute. (see AccountMoveLine._compute_partner_bank_id() in models/account_move_line.py)

### DIFF
--- a/account_payment_order/__init__.py
+++ b/account_payment_order/__init__.py
@@ -1,3 +1,4 @@
+from .hooks import pre_init_hook
 from . import models
 from . import report
 from . import wizard
--- a/account_payment_order/__manifest__.py
+++ b/account_payment_order/__manifest__.py
@@ -38,4 +38,5 @@
     ],
     "demo": ["demo/payment_demo.xml"],
     "installable": True,
+    "pre_init_hook": "pre_init_hook",
 }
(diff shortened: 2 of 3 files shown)


pedrobaeza:
`payment_order_ok` is a computed **non-stored** field, so you don't have to create the column. 

`partner_bank_id` is OK to be pre-created.

dannyadair:
@pedrobaeza Thank you very much for the quick review! I've adjusted the hook accordingly.

pedrobaeza (approved):

dalonsod (approved):
LGTM

pedrobaeza:
/ocabot merge patch

Can you fw-port it to 15?

dannyadair:
> /ocabot merge patch
> 
> Can you fw-port it to 15?

Submitted #1047

PR #1031, opened 2023-02-09
[15.0][FIX] account_banking_sepa_*: Remove post-installs

### LINKED ISSUE #977: [15.0] Error when installing account chart - account_payment_mode
## Module
account_payment_mode

## Describe the bug
The order of installation matters for `account_payment_mode`, a payment provider and an account chart causing crashes during automatic installations and manual installations. While for manual installations there is a workaround a workaround for automatic installations can be more tricky.

## To Reproduce
**Affected versions**: 15.0

1. Fresh Odoo installation
2. Install the modules `account_payment_mode` and a payment module (e.g. `payment_stripe`)
3. Install an account chart (e.g. `l10n_de_skr04`)

**Expected behavior**
If the installation happens in this order it produces an error and the installation of the account chart fails. I expect that the order of installation doesn't matter for the modules `account_payment_mode`, `payment_stripe` and `l10n_de_skr04`.

**Additional context**
```
2022-10-28 06:42:44,147 11 INFO odoo odoo.modules.loading: loading l10n_de_skr03/data/account_chart_template_data.xml
2022-10-28 06:42:49,714 11 WARNING odoo odoo.modules.loading: Transient module states were reset
2022-10-28 06:42:49,723 11 ERROR odoo odoo.modules.registry: Failed to load registry
2022-10-28 06:42:49,723 11 ERROR odoo odoo.http: Exception during JSON request handling.
...
  File "/tmp/addons/account/models/account_journal.py", line 437, in _check_payment_method_line_ids_multiplicity
    raise ValidationError(_("Some payment methods supposed to be unique already exists somewhere else.\n"
    odoo.exceptions.ValidationError: Some payment methods supposed to be unique already exists somewhere else.
    ([stripe] Stripe (inbound))

    View error context:
    '-no context-'
```

As far as I can see the bug is caused by the overwriting of `_default_outbound_payment_methods` and `_default_inbound_payment_methods` of `account.journal` of the mentioned module. If I remove those lines it works again. I'm not experienced enough with the in-depth accounting here to decide if this would be the right approach or it should be handled differently. Please support me here.

pedrobaeza:
From this version, Odoo already includes code on the account.payment.method creation for adding such method to each journal:

https://github.com/odoo/odoo/blob/a93ed75cb80f16f/addons/account/models/account_payment_method.py#L24-L38

so we don't need to explicitly create them for the new payment methods we are adding in these modules.

Fixes #977

@Tecnativa TT41146


victoralmau (approved):

dalonsod (approved):
👍

PR #1034, opened 2023-02-10
[14.0][IMP] account_payment_order: added bill date to Create Transactions wizard

dalonsod:
Some customers find useful bill date to be shown in Select Move Lines to Create Transactions wizard form, this PR simply adds it.

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -9,6 +9,7 @@ from odoo.fields import first
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    invoice_date = fields.Date(related="move_id.invoice_date")
     partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
         string="Partner Bank Account",
--- a/account_payment_order/wizard/account_payment_line_create_view.xml
+++ b/account_payment_order/wizard/account_payment_line_create_view.xml
@@ -53,6 +53,11 @@
                     >
                         <tree>
                             <field name="date" />
+                            <field
+                                name="invoice_date"
+                                optional="hide"
+                                string="Bill Date"
+                            />
                             <field name="move_id" required="0" />
                             <field name="journal_id" />
                             <field name="partner_id" />


dalonsod:
@mtbochaca could you review? Thanks!

dalonsod:
@dannyadair could you review? Thanks!

dalonsod:
@dannyadair @pedrobaeza could you review? Thanks!

dalonsod:
@ChristianSantamaria @lmiguens-solvos could you review? Thanks!

ChristianSantamaria (approved):
LGTM 👍

lmiguens-solvos (approved):
LGTM

dalonsod:
@pedrobaeza is this PR ready to be merged? When done, I'll make the fw-port to v15

pedrobaeza:
This needs a PSC review. I'm not able to do it right now, sorry.

dalonsod:
@HaraldPanten could you review? Thanks!

pedrobaeza (commented):
I have made a quick review, and adding such related field drains performance on the tree, as it needs to be extracted for each line, so I'm not sure this is convenient.

dalonsod:
Yes, I've simply used a related non-stored field because the expected list of moves to be shown in that tree view shouldn't be a large one. And that field is not intended to be used in other views.

Anyway, if this field could be dangerous, it could be solved making it `store=True` + pre-migration script, is it right?

pedrobaeza:
Not really, that would be worse, as on installation/update, existing lines will need to store the value.

It will also be confused for move lines not coming from invoices, or redundant on customer invoices. It seems a very specific need that shouldn't go on standard IMO.

gdgellatly:
Agree with @pedrobaeza , the practice of having a different invoice and move date is very very rare.

This from a midsized database, subject to IFRS, say 3 years old doing vanilla Odoo.

```
devel=# SELECT count(*) FROM account_move WHERE invoice_date != date AND invoice_date IS NOT NULL;
 count 
-------
   158
(1 row)

devel=# SELECT count(*) FROM account_move WHERE invoice_date IS NOT NULL;
 count  
--------
 786057
(1 row)
```

gdgellatly (changes requested):
See general comments in PR, if anything in a standard installation it will do more to confuse users than help them. and is mostly redundant anyway because by default, the accounting date and invoice date are the same. I think if you face a situation where this isn't the usual case, probably best addressed in a separate module, possibly more completely rather than show 2 dates.

dalonsod:
Thanks a lot for the comments @gdgellatly !

I have not problem creating a custom addon for this. I've simply make this improvement here because I've got some customers with such requirement (one of them has 99% of vendor bills in this situation) and I thought that it could be useful for other people.

Anyway, preventing that `invoice_date` wanted to be used by a few users, I've made the field `optional="hide"` in the view.

As @pedrobaeza said, if performance could be affected for the current functionality, no need to continue this way. Then, closing this.

PR #972, opened 2022-10-25
[16.0][MIG] account_payment_order: Migration to 16.0

wpichler:
https://github.com/OCA/bank-payment/issues/964

Needs https://github.com/OCA/bank-payment/pull/969

Standard Migration - just one thing - it was necessary to add groups="base.group_multi_company" to the order_id field in account_payment_line_form. Without this i do get this error:

Field 'company_id' used in domain of field 'order_id' ([('company_id', 'in', [company_id, False])]) is restricted to the group(s) base.group_multi_company.

order_id is check_company - so company_id domain gets added indirectly and should be ok without this group. Maybe because of the related company_id this problem occours ?

PR #1036, opened 2023-02-14
[16.0][MIG] account_payment_order: Migration to 16.0 

AnizR:
This pr is based on pr: #972 .
This is a proposition of correction of every open comment.
I created a new PR because I can't create a PR on your repo @wpichler

correction of comment https://github.com/OCA/bank-payment/pull/972#discussion_r1100214007:
[IMP] create with @api.model_create_multi tag

correction of comment https://github.com/OCA/bank-payment/pull/972#discussion_r1093122726:
[MIG] migration scripts

correction of comment https://github.com/OCA/bank-payment/pull/972#discussion_r1093130642:
[IMP] report account_payment_order use t-field

correction of comment https://github.com/OCA/bank-payment/pull/972#discussion_r1099896627:
[IMP] field company_id invisible on view account_payment_line


Reyes4711-S73 (approved):

hildickethan:
@pedrobaeza comments here https://github.com/OCA/bank-payment/pull/972#discussion_r1093143945 on the topic of migration files that they should be removed entirely. We only need the steps for their respective versions, a 15 -> 16 migration shouldn't need these scripts so they shouldn't be necessary.

Aside from that, the changes LGTM, but I would still encourage the squashing of some commits like I commented here https://github.com/OCA/bank-payment/pull/972#issuecomment-1422332367, there should only really be the "pre-commit" commit and the "migration" commit in the end, with the post-migration commits either being squashed or moved before the pre-commit since they're cherry-picks for previous versions. I would also squash your changes into @wpichler's as co-author

AnizR:
> @pedrobaeza comments here [#972 (comment)](https://github.com/OCA/bank-payment/pull/972#discussion_r1093143945) on the topic of migration files that they should be removed entirely. We only need the steps for their respective versions, a 15 -> 16 migration shouldn't need these scripts so they shouldn't be necessary.
> 
> Aside from that, the changes LGTM, but I would still encourage the squashing of some commits like I commented here [#972 (comment)](https://github.com/OCA/bank-payment/pull/972#issuecomment-1422332367), there should only really be the "pre-commit" commit and the "migration" commit in the end, with the post-migration commits either being squashed or moved before the pre-commit since they're cherry-picks for previous versions. I would also squash your changes into @wpichler's as co-author

Indeed but I will let @wpichler merge our pr's

AnizR:
Changes added in #972

PR #990, opened 2022-12-01
[16.0][MIG] account_banking_sepa_credit_transfer: Migration to 16.0

dzungtran89:
#964 

Depends:
- #974

PR #974, opened 2022-10-25
[16.0][MIG] account_banking_pain_base

wpichler:
https://github.com/OCA/bank-payment/issues/964

PR #1037, opened 2023-02-16
[MIG] account_banking_sepa_credit_transfer: Migration to 16.0

AnizR:
Migration of module account_banking_sepa_credit_transfer

Depends #974 #1036 

I know that there is pr #990 but imo I couldn't just suggest changes because there were too important.

e.g. https://github.com/OCA/bank-payment/blob/6edf8370fe695a1bad87ea711efd319560191e2a/account_banking_sepa_credit_transfer/models/account_payment_order.py#L81: it assumed that a payment order has bank_line_ids (it was the case in 15.0 but not anymore)


pedrobaeza:
This should include the part of #1031 that corresponds to this.

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -107,6 +113,11 @@ def setUpClass(cls):
         # Trigger the recompute of account type on res.partner.bank
         cls.partner_bank_model.search([])._compute_acc_type()
 
+        # cls.main_company.write({
+        #     'account_journal_payment_debit_account_id': cls.payment_debit_account_id.id,
+        #     'account_journal_payment_credit_account_id': cls.payment_credit_account_id.id
+        # })
+
```
hildickethan, line 120: to remove if not necessary

hildickethan:
> This should include the part of #1031 that corresponds to this.

From what I can see this is included in the migration commit, but this being a migration from 14 -> 16 when 15 is merged seems a bit off when it comes to commit history, as if it was done from 15 that commit would be there

We have 2 PRs now for the same module, I think this one is unnecessary with the current state of https://github.com/OCA/bank-payment/pull/990 :thinking:

PR #1038, opened 2023-02-16
[14.0][IMP] account_payment_order: Allow fixed date in the past

hildickethan:
- Add a checkbox to allow a past accounting date when the type is "Fixed"
- I made it checkbox reliant so it's opt-in, but IMO the accounting date should always be the scheduled date when using "Fixed" type, since that's logically what a fixed/scheduled date would be

Use case:

Payments are scheduled banking-side to be done on a specific day, you receive the info from the bank but for whatever reason, no one confirms the Payment Order in Odoo on the same day.
Next working day you confirm the payment order but the accounting date fast-forwards to the current day, which no longer reflects the schedule that was planned

### DIFF
--- a/account_payment_order/i18n/account_payment_order.pot
+++ b/account_payment_order/i18n/account_payment_order.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 14.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2023-02-16 12:27+0000\n"
+"PO-Revision-Date: 2023-02-16 12:27+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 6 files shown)


hildickethan:
@OCA/banking-maintainers if you could have a look please, I'm still of the opinion that it doesn't really need to be opt-in since "fixed" date should just force it to that date, the config was mostly there to be easier to accept the change

pedrobaeza (approved):

carlosdauden (approved):

giarve:
Let's consider the following use case:

The user confirms an `account.payment.order`, generates a SEPA file and presses the button to confirm that the file has been uploaded. Then the account moves for the payments are confirmed, but they are confirmed with accounting dates in the future, and not when the move to the pending payments account has been actually made (imagine it is a pending payment account).

I would allow the user to set these move dates to date_uploaded date instead (maybe with another dropdown, but different dates other than `date_prefered`).

I can elaborate more if it is hard to understand or even create a new issue, but I found this is kind of a recent PR.

alexis-via:
If payments are scheduled banking-side, why do you need to create a payment order in Odoo ?? I think this change was not a good idea. And if you are in a corner-case scenario with payments scheduled banking side and you really want, for some reason, to make a payment order in Odoo, you can make the change in the code in a specific module, and not change the account_payment_order module. If we start to support all the weired scenario, we will end-up with a super-complex code base with tons of options that are used by 0,5% of the users... and it will kill OCA/bank-payment.
That's my opinion ! :)

pedrobaeza:
There's a flow that this change is favoring: you need to cancel a payment order already done by any reason (for example, wrong outstanding account), but correctly sent to the bank. Finishing the order again will change the date to today instead of the real date where it was done. That's why I think this can be useful.

PR #1039, opened 2023-02-16
[14.0] Show all errors on validation at once

simgonzalez:
Pull request to avoid testing in the loop fields that are available before the loop and to add the feature of showing all the validation user errors at once and for the user not to have to launch 10 times the process to get errors one by one.


pedrobaeza:
This seems to be doing more than the aggregation of the messages. Please, each change at once.

simgonzalez:
I've closed this one to open one on a clean branch
https://github.com/OCA/bank-payment/pull/1040

PR #1040, opened 2023-02-16
[14.0] [IMP] Gather all the UserError as one when validating payments in payment order

simgonzalez:
PR top add the feature of showing all the validation user errors at once and for the user not to have to launch 10 times the process to get errors one by one.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -289,9 +289,13 @@ class AccountPaymentOrder(models.Model):
             order.payment_ids.action_cancel()
             order.payment_ids.unlink()
             # Prepare account payments from the payment lines
+            payline_err_text = []
             group_paylines = {}  # key = hashcode
             for payline in order.payment_line_ids:
-                payline.draft2open_payment_line_check()
+                try:
+                    payline.draft2open_payment_line_check()
+                except UserError as e:
+                    payline_err_text.append(e.args[0])
                 # Compute requested payment date
                 if order.date_prefered == "due":
                     requested_date = payline.ml_maturity_date or payline.date or today
@@ -300,8 +304,7 @@ class AccountPaymentOrder(models.Model):
                 else:
                     requested_date = today
                 # No payment date in the past
-                if requested_date < today:
-                    requested_date = today
+                requested_date = max(today, requested_date)
                 # inbound: check option no_debit_before_maturity
                 if (
                     order.payment_type == "inbound"
@@ -309,7 +312,7 @@ class AccountPaymentOrder(models.Model):
                     and payline.ml_maturity_date
                     and requested_date < payline.ml_maturity_date
                 ):
-                    raise UserError(
+                    payline_err_text.append(
                         _(
                             "The payment mode '%s' has the option "
                             "'Disallow Debit Before Maturity Date'. The "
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please follow commit message guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

- No version on the commit message, but on the PR title.
- `[FIX]` tag
- Name of the module
- Short explanation of what in the first line
- Long explanation, whys and so on on following lines

account_payment_order/models/account_payment_order.py
```diff
@@ -344,6 +348,13 @@ def draft2open(self):
                         "paylines": payline,
                         "total": payline.amount_currency,
                     }
+            # Raise errors that happened on the validation process
+            if payline_err_text:
+                raise UserError(
+                    _("A bunch of validation didn't passed\n")
```
pedrobaeza:
```suggestion
                    _("There's at least one validation error:\n")
```

pedrobaeza (commented):
And check pre-commit stuff

account_payment_order/models/account_payment_order.py
simgonzalez: Is there a way to check pre-commit stuff before actually the PR check ?

pedrobaeza: You can install pre-commit locally. Having pipx installed (`pip install pipx`), do `pipx install pre-commit`. Then, being on your repo directory, do `pre-commit install -f` for installing a hook. Now, each `git commit` on that repo, will launch first the pre-commit stuff.

account_payment_order/models/account_payment_order.py
```diff
@@ -333,17 +336,26 @@ def draft2open(self):
...
-                    hashcode = payline.id
-                if hashcode in group_paylines:
-                    group_paylines[hashcode]["paylines"] += payline
-                    group_paylines[hashcode]["total"] += payline.amount_currency
-                else:
-                    group_paylines[hashcode] = {
+                    group_paylines[payline.id] = {
                         "paylines": payline,
                         "total": payline.amount_currency,
                     }
```
pedrobaeza: Don't repeat code:

```suggestion
                hashcode = payline.payment_line_hashcode() if order.payment_mode_id.group_lines else payline.id
                if hashcode in group_paylines:
                    group_paylines[hashcode]["paylines"] += payline
                    group_paylines[hashcode]["total"] += payline.amount_currency
                else:
                    group_paylines[hashcode] = {
                        "paylines": payline,
                        "total": payline.amount_currency,
                    }
```

account_payment_order/models/account_payment_order.py
```diff
@@ -331,11 +334,11 @@ def draft2open(self):
                 with self.env.norecompute():
                     payline.date = requested_date
                 # Group options
-                if order.payment_mode_id.group_lines:
-                    hashcode = payline.payment_line_hashcode()
-                else:
-                    # Use line ID as hascode, which actually means no grouping
-                    hashcode = payline.id
+                hashcode = (
```
pedrobaeza, line 337: Now that the diff is clearer, I'm seeing that previous code was OK. Why did you initially change it? I think it's better to leave it as it was for reducing diff. Sorry for changing my mind.

simgonzalez: I know but the pre-commit didn't let me commit as it was because of complexity factor that was too high. This is the reason why i modified this part

pedrobaeza: OK then.

pedrobaeza (approved):

simgonzalez:
@pedrobaeza Do you know to who i should ask a second review ? Before this PR automatically closes :D 
Thanks ^^

ecino (approved):

pedrobaeza:
You can fw-port it to upper versions:

/ocabot merge patch

PR #919, opened 2022-04-19
[14.0][IMP] account_payment_order: Add existing payment references to communication

rousseldenis:
If some movements have been reconciled with the original invoice,
their references should be added in communication too.

e.g.: Manual credit notes

PR #1041, opened 2023-02-16
[15.0][FW] [14.0][REF] account_payment_*: Use native payment object

hhgabelgaard:
Port of #979 from 14.0 to 15.0.

Also includes:

- #918
- #919
- #936


pedrobaeza:
Hi @hhgabelgaard, thanks for the work. I know it's very difficult to equal branches, because contributors are not forward-porting things at the same time they incorporate it on a lower branch (which I consider a bit disrespectful, but also counterproductive at long term), joined to the changes over versions.

Anyway, for not charging you more, I'm doing such work in this some kind of spare time, starting with #1053 for things before the refactoring, and then I will port the refactoring.

Closing this thus. Thanks again.

pedrobaeza:
Continuation at #1054

PR #1045, opened 2023-02-22
[14.0][FIX] account_payment_order_notification: Show supplier invoice reference if applicable.

victoralmau:
Show supplier invoice reference if applicable.

Please @pedrobaeza and @ernestotejeda can you review it?

@Tecnativa TT40795


account_payment_order_notification/migrations/14.0.1.4.0/noupdate_changes.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<odoo>
+    <record id="email_account_payment_order_notification" model="ir.rule">
```
ernestotejeda:
```suggestion
    <record id="email_account_payment_order_notification" model="mail.template">
```

ernestotejeda (changes requested):

ernestotejeda (approved):

pedrobaeza (commented):
If the record was noupdate=0, there's no need of migration scripts. And I'm not sure to need to switch to noupdate=1. You can do copies of the template...

victoralmau:
Sorry for the confusion, I thought it was already in noupdate=1 and when I saw it, I changed it.

IMO all templates (or almost) should be set to noupdate=1 to allow the user to change what they want. Almost all odoo core and OCA module mail templates are set like this, preventing the user from having to create a copy unnecessarily.

pedrobaeza:
`noupdate=1` is needed when a process (like the button "Send quotation" in sales orders) uses as default, and you can't change this. But in this case, you decide which mail template to send on the credit control policies, so the flow should be:

- Copy the standard template.
- Use that other one.

I think it's correct to keep it noupdate=0, and remove migration scripts.

victoralmau:
Ok, changes done.

pedrobaeza:
You can remove this migration script: https://github.com/OCA/bank-payment/blob/2b0d65105113b6930701e0521cc5616dc983dc58/account_payment_order_notification/migrations/14.0.1.2.0/noupdate_changes.xml

pedrobaeza (approved):
/ocabot merge minor

PR #1046, opened 2023-02-22
[15.0] account_payment_partner: Add pre_init_hook to add computed columns

dannyadair:
Add columns to avoid Memory error on an existing Odoo instance with lots of data.

The payment_mode_id fields are introduced by this module and computed only from each other or the also newly introduced supplier_payment_mode_id and customer_payment_mode_id on res.partner, so they can stay NULL, nothing to compute.


pedrobaeza (approved):
Fast-tracking the fw-port:

/ocabot merge patch

Please move this one also to 16.0

dannyadair:
> Please move this one also to 16.0

Submitted #1048

PR #1047, opened 2023-02-22
[15.0] account_payment_order: Add pre_init_hook to add computed column

dannyadair:
Add partner_bank_id column to avoid MemoryError on an existing Odoo instance with lots of data.

partner_bank_id on account.move.line requires payment_order_ok to be True which it won't be as it's newly introduced - again nothing to compute. (see AccountMoveLine._compute_partner_bank_id() in models/account_move_line.py)

### DIFF
--- a/account_payment_order/__init__.py
+++ b/account_payment_order/__init__.py
@@ -1,3 +1,4 @@
+from .hooks import pre_init_hook
 from . import models
 from . import report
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Fast-tracking the fw-port:

/ocabot merge patch

PR #1050, opened 2023-02-27
[14.0] Add warning banner when payment order is not SEPA

alexis-via:
The field 'sepa' on account.payment.order is only display for SEPA payment methods.
If the option "show warning if not SEPA" is enabled on the payment method, a warning banner is now displayed on payment orders with a SEPA payment method which are not SEPA:

sepa_warn_banner


PaulGoubert (approved):

alexis-via:
Tests are now green!

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -105,6 +111,8 @@ def _sepa_iban_prefix_list(self):
         ]
 
     @api.depends(
+        "payment_mode_id.payment_method_id.pain_version",
```
pedrobaeza: I don't think you need this convoluted depends. Only `payment_mode_id`. If not, changing these fields in payment methods will trigger the recomputation of all payment orders, which is not performant and even not convenient.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -24,7 +24,13 @@
 class AccountPaymentOrder(models.Model):
     _inherit = "account.payment.order"
 
-    sepa = fields.Boolean(compute="_compute_sepa", readonly=True, string="SEPA Payment")
+    sepa = fields.Boolean(compute="_compute_sepa", string="SEPA Payment", store=True)
+    sepa_payment_method = fields.Boolean(
+        compute="_compute_sepa",
+        string="SEPA Payment Method",
+        store=True,
```
pedrobaeza: Please don't store these fields that are only for showing a banner, and that will be triggered on module update, making it slower.

pedrobaeza (changes requested):

account_banking_pain_base/models/account_payment_order.py
alexis-via: done

account_banking_pain_base/models/account_payment_order.py
alexis-via: OK. So I removed the "sepa" field from the account.payment.order tree view, because it is a bad practise to have a non-stored field in a tree view.

alexis-via:
@pedrobaeza I took into account all your remarks.

alexis-via:
I'll forward-port it to v16 once all modules are merged (this PR impacts several different modules, some of which are not merged yet)

pedrobaeza:
Please fw-port it to v15 too. I'll promise to review and fast-track it. Deal?

alexis-via:
I also rewrote the code for allow_blocked, which was a mess: there was no solution to block when push payment lines from an invoice. The allow_blocked configuration is now on the payment mode and applies both to push and pull mode.

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -77,8 +76,8 @@ def _prepare_move_line_domain(self):
             domain += [("partner_id", "in", self.partner_ids.ids)]
         if self.target_move == "posted":
             domain += [("move_id.state", "=", "posted")]
-        if not self.allow_blocked:
-            domain += [("blocked", "!=", True)]
+        if not self.order_id.allow_blocked:
```
pedrobaeza: Shouldn't this be `if not self.order_id.payment_mode_id.allow_blocked:`?

account_banking_pain_base/views/account_payment_method.xml
```diff
@@ -10,10 +10,15 @@
         <field name="arch" type="xml">
             <field name="payment_type" position="after">
                 <field name="pain_version" />
+
```
pedrobaeza: Please avoid this introduced empty line

account_banking_pain_base/models/account_payment_method.py
```diff
@@ -16,6 +16,7 @@ class AccountPaymentMethod(models.Model):
         "the corresponding unaccented character, so that only ASCII "
         "characters are used in the generated PAIN file.",
     )
+    warn_not_sepa = fields.Boolean(string="Warn If Not SEPA")
```
pedrobaeza, line 19: I think the warning shouldn't be selectable. If the payment method is SEPA, then show the warning.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -115,30 +121,41 @@ def _compute_sepa(self):
...
-                and order.company_partner_bank_id.sanitized_acc_number[:2]
-                not in sepa_list
-            ):
-                sepa = False
-            for pline in order.payment_line_ids:
-                if pline.currency_id != eur:
-                    sepa = False
-                    break
-                if pline.partner_bank_id.acc_type != "iban":
+            sepa_payment_method = False
```
pedrobaeza, line 124: For making the code more readable, since v13 you can assign directly to the field the variable without performance penalty, and also assign for the whole recordset.

account_banking_pain_base/models/account_payment_method.py
alexis-via: I made it an option because I thought that this module may be used in non-sepa countries to produce ISO2022 files. And those users maybe upset to have the warning banner all the time.

pedrobaeza: But the banner should only be shown when the payment method is SEPA. I think you have already checked that on your compute. No problem thus on being "mandatory". It's not going to appear for non SEPA orders.

account_banking_pain_base/views/account_payment_method.xml
alexis-via: fixed

account_banking_pain_base/models/account_payment_method.py
pedrobaeza: This is not attended.

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -26,7 +26,6 @@ class AccountPaymentLineCreate(models.TransientModel):
         selection=[("posted", "All Posted Entries"), ("all", "All Entries")],
         string="Target Moves",
     )
-    allow_blocked = fields.Boolean(string="Allow Litigation Move Lines")
```
pedrobaeza, line 29: Why removing this? It's useful.

account_payment_order/wizard/account_payment_line_create.py
alexis-via: You're right. Fixed.

pedrobaeza (changes requested):
Apart for not being all the comments attended, this is introducing other undesired changes like the move of the allow_blocked to the payment mode. Please only one change per PR for being discussed separately.

account_banking_pain_base/models/account_payment_method.py
alexis-via: Let me re-explain my point of view in more details. The modules account_banking_sepa_credit_transfer and account_banking_sepa_direct_debit produce ISO20022 files that can be SEPA or not SEPA. We could have called those modules account_banking_iso20022_credit_transfer and account_banking_iso20022_direct_debit, but we chose to call them "sepa"... because, from our point of view of EU users, we use it mostly for SEPA payments. Same for the "code" of the payment method.
But, maybe some users outside of EU use those module to generate non-SEPA ISO20022 files. And they may not want to always have the warning banner. I must say I have no idea if the modules account_banking_sepa_xxx are used by companies outside EU. But I know that ISO20022 is a world-wide standard ; it is not just designed for EU/SEPA.

account_payment_order/wizard/account_payment_line_create.py
alexis-via: Because it is now a property of the payment mode, not a question you ask to the use in this wizard. The user can use the button on the invoice to push it on the payment order and skip the wizard.

alexis-via:
@pedrobaeza I made separate PRs in v16. We can focus on v16 now.

PR #1052, opened 2023-03-02
[16.0] account_payment_mode: add "sequence" field on account.payment.mode

alexis-via:
Currently, payment modes are ordered by name. With the introduction of a sequence field on account.payment.mode, the user can now decide the order.

### DIFF
--- a/account_payment_mode/models/account_payment_mode.py
+++ b/account_payment_mode/models/account_payment_mode.py
@@ -13,7 +13,7 @@ class AccountPaymentMode(models.Model):
 
     _name = "account.payment.mode"
     _description = "Payment Modes"
-    _order = "name"
+    _order = "sequence, name"
     _check_company_auto = True
 
     name = fields.Char(required=True, translate=True)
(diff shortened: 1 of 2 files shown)


rvalyi (approved):
LGTM

account_payment_mode/models/account_payment_mode.py
```diff
@@ -13,7 +13,7 @@ class AccountPaymentMode(models.Model):
 
     _name = "account.payment.mode"
     _description = "Payment Modes"
-    _order = "name"
+    _order = "sequence, id"
```
pedrobaeza: Please keep name as second order factor

account_payment_mode/models/account_payment_mode.py
```diff
@@ -68,6 +68,7 @@ class AccountPaymentMode(models.Model):
     )
     active = fields.Boolean(default=True)
     note = fields.Text(translate=True)
+    sequence = fields.Integer()
```
pedrobaeza: It's the usual default. Not sure if 0 can cause any trouble.

```suggestion
    sequence = fields.Integer(default=10)
```

flotho (approved):
LGTM

pedrobaeza (approved):
/ocabot merge minor

PR #1054, opened 2023-03-04
[15.0][REF] account_payment_*: Use native payment object

pedrobaeza:
Forward-port of #979 

The previous approach creates manually the journal entries and does all the hard work, plus not being 100% compatible with the bank statement reconciliation widget (requiring a patch on OCB to see blue lines).

That decision made sense on the moment it was done (v9), where the native payment model (account.payment) was very limited, and wasn't able to store all the needed information for the bank transaction.
    
Now that the limitations are gone, we can get rid off this extra model, and generate instead `account.payment` records, using both the native model + methods to perform the same operations.
    
This serves also to workaround the problem found in #966.
    
All the code, views and tests of main module have been adapted to this new approach.

@Tecnativa


pedrobaeza:
Working on the migration scripts meanwhile I see the CI to not execute them if coming from v14.

pedrobaeza:
Everything done and green, so merging this one and now proceeding with the extra module for grouped output + the migration to 16.0. Almost there!

/ocabot merge major

PR #1066, opened 2023-04-04
[15.0][FIX] account_banking_sepa_direct_debit: CSS styling not applied in reports.

stefan-tecnativa:
cc @Tecnativa TT42403

@carlosdauden please review!

The css styling for mandate reports was not being applied.

### DIFF
--- a/account_banking_sepa_direct_debit/__manifest__.py
+++ b/account_banking_sepa_direct_debit/__manifest__.py
@@ -18,7 +18,7 @@
         ],
     },
     "assets": {
-        "web.assets_backend": [
+        "web.report_assets_common": [
             "/account_banking_sepa_direct_debit/static/src/css/report.css"
         ],
     },


carlosdauden (approved):

pedrobaeza:
Please check pre-commit

account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <template id="sepa_direct_debit_mandate_document">
+        <link
```
sergio-teruel: Yo do not need load the style here.
The style sheet is not loaded because in manifest file the declaration of css is not right.

Change in manifest file assets_backend by report_assets_common

sergio-teruel (changes requested):

account_banking_sepa_direct_debit/views/report_sepa_direct_debit_mandate.xml
stefan-tecnativa: Thanks for the correction.

sergio-teruel (approved):

PR #1067, opened 2023-04-04
[14.0][IMP] actually link the payment_order to the move

simgonzalez:
When using the button "add to debit order" the account_move should actually be linked with payment_order.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -118,6 +118,7 @@ class AccountMove(models.Model):
                         )
                         % (count, payorder.id, payorder.display_name)
                     )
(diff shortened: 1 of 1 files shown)


pedrobaeza:
IIRC, that many2one is not for linking the imported move, but for the generated move from the payment order.

simgonzalez:
@pedrobaeza I agree that certainly works like this but why would the link not work in the other way around ? Maybe my solution isn't the best but currently we have a button that add the move to payment order without setting correctly some properties that we would except to be set for a move in a payment order. 
If we don't want that line of code, in my opinion, the button shouldn't exist either or I might be missing something in the process.

PR #1068, opened 2023-04-05
[16.0][FIX] account_banking_mandate: move's create should be overrided properly to avoid flush_recordset

dzungtran89:
## Context

- Install two modules: `account_banking_mandate` and `account_banking_sepa_credit_transfer`
- The test cases will be run in order: `*mandate` first, then `*transfer`
- Since [cb76725b](https://github.com/odoo/odoo/commit/cb76725b), odoo calls [flush_recordset](https://github.com/odoo/odoo/commit/cb76725b#diff-1e3bd6be3bfb83a37ec9fb800ce8b1c95afe0be90ff792874ae7299c320a2f6eR2257)() on moves before posting them
- The tests of `*transfer` will fail. Somehow the move's names are incorrectly computed and they both have the same, so flush_recordset() fails:

```
...
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "account_move_unique_name"
DETAIL:  Key (name, journal_id)=(PBNKFB/2023/00001, 10) already exists.
...
```

### DIFF
--- a/account_banking_mandate/models/account_move.py
+++ b/account_banking_mandate/models/account_move.py
@@ -12,48 +12,20 @@ class AccountMove(models.Model):
         "account.banking.mandate",
         string="Direct Debit Mandate",
         ondelete="restrict",
-        readonly=True,
+        readonly=False,
         check_company=True,
         states={"draft": [("readonly", False)]},
+        compute="_compute_mandate_id",
+        store="True",
     )
     mandate_required = fields.Boolean(
         related="payment_mode_id.payment_method_id.mandate_required", readonly=True
     )
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        """Fill the mandate_id from the partner if none is provided on
-        creation, using same method as upstream."""
-        onchanges = {
-            "_onchange_partner_id": ["mandate_id"],
-            "_onchange_payment_mode_id": ["mandate_id"],
-        }
-        for onchange_method, changed_fields in list(onchanges.items()):
-            for vals in vals_list:
-                if any(f not in vals for f in changed_fields):
-                    move = self.new(vals)
-                    move = move.with_company(move.company_id.id)
-                    getattr(move, onchange_method)()
-                    for field in changed_fields:
-                        if field not in vals and move[field]:
-                            vals[field] = move._fields[field].convert_to_write(
-                                move[field], move
-                            )
-        return super().create(vals_list)
-
-    def set_mandate(self):
-        if self.payment_mode_id.payment_method_id.mandate_required:
-            self.mandate_id = self.partner_id.valid_mandate_id
-        else:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why not converting the fields to computed writable (computed stored with readonly=False) to avoid these hacks?

dzungtran89:
> Why not converting the fields to computed writable (computed stored with readonly=False) to avoid these hacks?

Hi @pedrobaeza , thanks for the review. You meant the `name` field? This field has already been [computed writable field](https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L79)

pedrobaeza:
No, I meant `mandate_id`, which seems to be the field the code in the create is being triggered.

hildickethan:
We saw this recently in some tests too of another module, not sure on the details or the part of this module causing it but we found a workaround with creating the move with name `/` so Odoo would do its regular sequencing.
- https://github.com/OCA/bank-payment/pull/1070

Though I think changing to `@api.model_create_multi` is a "must do" since every create in base 16.0 is now multi, that is if the create is preserved but I think it is a good idea to make the fields computed to avoid the create with onchanges altogether

dzungtran89:
Thanks @pedrobaeza for your suggestion, the PR has been updated

About the attr `readonly`, it should be `False` to avoid the recomputation on the field which causes the empty mandate, see [here](https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L3910)

account_banking_mandate/views/account_move_view.xml
```diff
@@ -12,6 +12,7 @@
             <field name="partner_bank_id" position="after">
                 <field
                     name="mandate_id"
+                    readonly="1"
```
pedrobaeza: Why this without force_save=1?

dzungtran89: The field is defined on the form field, that attr will be required if the field is put on the tree (i.e: invoice lines)

pedrobaeza: This is the invoice form, so I don't get what are you saying about a tree, but this should definitively be editable.

pedrobaeza:
Please don't do several commits. Squash them. We can see the partial changes here in GitHub thanks to the messages that are automatically logged.

hildickethan:
Could you resolve the conflicts that have emerged since https://github.com/OCA/bank-payment/pull/1078 was merged?

dzungtran89:
> Could you resolve the conflicts that have emerged since #1078 was merged?

ok, then my PR is an improve on computing the `mandate_id` instead of relying on `onchange`

pedrobaeza (approved):

hildickethan (approved):

PR #1070, opened 2023-04-05
[16.0][FIX] account_payment_order: create lines with same sequence name

Reyes4711-S73:
When you are creating a debit order with various transactions you get the same sequence name in payment_lines. When you are confirming the debit order, you couldn't do it due to the fact that the payment lines have the same name.
Before
image
image
After
image
debit order done
image

### DIFF
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -192,6 +192,9 @@ class AccountPaymentLine(models.Model):
             "date": self[:1].date,
             "currency_id": self.currency_id.id,
             "ref": self.order_id.name,
+            # Put the name as the wildcard for forcing a unique name. If not, Odoo gets
+            # the sequence for all the payment at the same time
+            "name": "/",
             "payment_reference": "-".join([line.communication for line in self]),
             "journal_id": journal.id,
             "partner_bank_id": self.partner_bank_id.id,


pedrobaeza:
This seems a bit weird... Why tests aren't triggering this problem? There are payment orders with several lines. Can you point to the code where such name is put?

Another thing: in the commit message (on contrary that the PR title), you shouldn't include version number, so please remove `[16.0]` if this still applies.

Reyes4711-S73:
@pedrobaeza https://github.com/odoo/odoo/blob/629d9cb15a6a6eb68756bad9da83cf0a81a3b5bd/addons/account/models/account_move.py#L2419
In this code line, the domain gets the same name. This is due to the fact that in odoo can't group moves. I think it's easier to change in this module adding the "/" name in draft state (like if you create manually one payment) than change this case in odoo or creating one per one all payments.

pedrobaeza:
OK, thanks for the explanation. The title of the PR was correct and should contain the Odoo version. It's the commit message which shouldn't.

account_payment_order/models/account_payment_line.py
```diff
@@ -192,6 +192,7 @@ def _prepare_account_payment_vals(self):
             "date": self[:1].date,
             "currency_id": self.currency_id.id,
             "ref": self.order_id.name,
+            "name": "/",
```
pedrobaeza, line 197: Always put comments for things that are not self-explanatory. If not, you'll have the risk that someone (or yourself) in the future removes them considering it not needed.

```suggestion
            # Put the name as the wildcard for forcing a unique name
            # If not, Odoo gets the sequence for all the payment at 
            # the same time
            "name": "/",
```

Reyes4711-S73:
@pedrobaeza Done

pedrobaeza (approved):

hildickethan (approved):

PR #1071, opened 2023-04-09
[14.0][FIX] account.payment - use same field for payment communicatio…

luc-demeyer:
The replacement of bank.payment.line by the native account.payment (cf. https://github.com/OCA/bank-payment/commit/b8d818967bc221f1663bf8be99ffa4165f321414) has caused a couple of issues for which this PR provides the first of a couple of corrections that we'll need to make.

The Odoo standard account.payment uses the 'ref' field for the Payment Communication ('Memo' field on the UI) whereas the current OCA module uses the 'ref' field for the payment order number and the payment_reference field for the payment communication.

The refactored code works +/- fine since the generated xml file also uses the payment_reference field, but it's not consistent with native Odoo and the User Interface is clumpsy since clicking on a Payment Transaction displays the wrong  Payment Communication. Especially in case of a grouped transaction the end user may want to see (and modify) the payment communication.

This PR now uses the 'Memo (ref) field like native Odoo.

A few areas to address after merging this PR:
- Imho we should add the 'ref' field to the 'Payment Tr'ansactios' tree view
- We may consider to align also the grouping logic to native Odoo (" ".join in stead of "-".join)
- After merging this PR we can change the payment communication but it requires too many clicks (we need to modify the  account.payment object via Vendor -> Payments, imho we should provide an easy but safe way to do this from the payment order)


pedrobaeza:
Isn't this changing the fields used in `account_payment_order` before the refactoring?

luc-demeyer:
@pedrobaeza 
>>> Isn't this changing the fields used in account_payment_order before the refactoring?

The ref field is used by native Odoo, cf. https://github.com/odoo/odoo/blob/2902ca109248fbe3aba3ff1c328fbc10c86a8b4c/addons/account/wizard/account_payment_register.py#L516
If we refactor to use native Odoo code than imho we should go all the way, currently we don't see the payment communication on the UI since we use another field than native Odoo.

pedrobaeza:
Yeah, but this doesn't try to imitate the enterprise behavior, but to continue with what is expected using payment orders before and after the refactoring, and I think you are changing such behavior.

luc-demeyer:
@pedrobaeza 
A couple of screen shots will make it clear that the code (since the refactoring) doesn't work from a usability standpoint.

We create a payment order with two transactions (payment mode is set to group payments)

transactions

When we confirm we get a single payment transaction:

payment_transactions

On the tree view we have no clue what payment communication the vendor will receive in his bank statement.

When we open the payment transaction  form view on a db without this PR you get:

payment_transactions_form

Hence the user gets the impression that the payment communication sent to the bank will be the payment order number.

After applying this PR you get:

payment_transactions_form_with_fix

Hence now you see what the payment communication will be.
If your vendor requires more details in the communication than you can go to Vendor -> Payments and adjust it.

I made this PR for a customer who needs to change on a regular basis this communication.
That's way I have stated in this PR that followup PR's would be appropriate. 
Changing the communication now requires too many clicks.

pedrobaeza:
Thanks for the screenshots. OK to make more accessible the field to show and modify the communication, but not OK to change the field to use. Please add `payment_ref` to the views instead. If not, you are breaking the compatibility and half-baked payment orders won't have any reference. Maybe a migration script can be done to transfer the information, but I think I still prefer to keep the same field, as it's also more meaningful in the name than "memo".

luc-demeyer:
I have now completed this PR based upon feedback from Pedro and user feedback on the payment order usability.

Changes in the second commit:
- migration script to ensure that the legacy payment order payment_line_ids are consistent with the newly created ones
- add payment_order_id to account.payment form view
- account.payment,ref field : label 'Memo' -> renamed to 'Payment Reference' (same label as on the Invoices at the origin of the payment transactions)
- account.payment.line,communication field : label 'Communication' -> renamed to 'Payment Reference' (same label as on the Invoices at the origin of the payment transactions)
- account.payment.line,name field : technical field, not relevant for end user. label 'Payment Reference' -> renamed to 'Payment Line ID' and hidden from UI via optional='hide'
- added 'Payment Reference' field also to account.payment tree view.
- Payment Reference (communication) in case of grouped invoices into a single payment: " - ".join in stead of "-".join since we see many vendors using "-" in their invoice numbering which makes the "-".join confusing.

Once this PR is merged I'll submit 2 add-on PRs:
- allow to update payment reference in payment order,payment_line_ids
- allow to see payment order from invoice form view
I have code available for these two extra usability features but since there are several ways to implement this it's imho better to work with two seperate pull requests and get a consensus on how we add these elements to the UI.

victoralmau:
In general I think it is a good approach, but it is important to differentiate between sales and purchases.

- **Sales**: The `payment_reference` field of the invoice should be used.
- **Purchases**: The `ref` field of the invoice should be used.

I think this should take into account what was said before and in my opinion, set the value of the `communication` field from `account.payment.line` (https://github.com/OCA/bank-payment/blob/14.0/account_payment_order/models/account_move_line.py#L98 in this case it would be the opposite, `ref` field first).

What do you think @pedrobaeza ?

TT45153

pedrobaeza:
Please go with it, Víctor.

victoralmau:
Superseed by https://github.com/OCA/bank-payment/pull/1142

pedrobaeza:
Wrong reference, Víctor.

PR #1072, opened 2023-04-13
[16.0][FIX] account_payment_partner: bad application of attributes on the field partner_bank_id

RodrigoBM:
This commit fixes showing the same field twice in customer invoices.

Without change:

- Customer Invoice:
Selección_320

With change:

- Customer Invoice:
Selección_321

- Vendor Bills:
Selección_323

### DIFF
--- a/account_payment_partner/views/account_move_view.xml
+++ b/account_payment_partner/views/account_move_view.xml
@@ -35,7 +35,13 @@
                 <field name="payment_mode_filter_type_domain" invisible="1" />
                 <field name="partner_bank_filter_type_domain" invisible="1" />
             </field>
-            <field name="partner_bank_id" position="attributes">
+            <xpath
+                expr="//group[@id='header_right_group']//field[@name='partner_bank_id']"
+                position="attributes"
+            >
+                <attribute name="context">
+                    {'default_partner_id':commercial_partner_id}
+                </attribute>
                 <attribute name="domain">
                     [('partner_id', '=', partner_bank_filter_type_domain),
                     '|',('company_id', '=', company_id),('company_id', '=', False)]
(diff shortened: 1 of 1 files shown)


account_payment_partner/views/account_move_view.xml
```diff
@@ -35,20 +35,29 @@
...
+                </attribute>
                 <attribute name="domain">
                     [('partner_id', '=', partner_bank_filter_type_domain),
                     '|',('company_id', '=', company_id),('company_id', '=', False)]
                 </attribute>
                 <attribute name="attrs">
                     {'required': [('bank_account_required', '=', True),('move_type', 'in', ('in_invoice', 'in_refund'))],
                     'readonly': [('state', '!=', 'draft')],
-                    'invisible': [('move_type', '=', 'entry')]}
+                    'invisible': [('move_type', 'not in', ('in_invoice', 'in_refund', 'in_receipt'))]}
```
pedrobaeza, line 52: It should be shown on customer invoices as well

RodrigoBM: The problem is that without the change, the partner_bank_id field that is being modified in the xml is the one shown in supplier invoices, causing it to be shown in both supplier invoices and customer invoices and this is incorrect.

For this reason it is necessary to apply the attribute changes to the two partner_bank_id fields in the base odoo xml. 

- https://github.com/OCA/bank-payment/pull/1072/files#diff-eb13f3e50e164744b9ed2075e137dc0d7dc5ea097a6fa128d7c2366a0241f432R38
- https://github.com/OCA/bank-payment/pull/1072/files#diff-eb13f3e50e164744b9ed2075e137dc0d7dc5ea097a6fa128d7c2366a0241f432R55

With the change, now only the correct field is shown on customer invoices.

- Customer Invoice:

Selección_321

- Bills:

Selección_323

Regards

pedrobaeza (approved):

RodrigoBM:
FYI @ThomasBinsfeld @atchuthan @pere-aquarian

JavierIniesta (approved):

RodrigoBM:
ocabot is on holiday today

pedrobaeza:
All the nights there are some processes that are executed and queued, and the queue is still being handled at these hours. About 10:00 or so (it depends the day), it will be empty and will start to process current commands.

PR #1078, opened 2023-04-26
[FIX] account_banking_mandate is not overriding the create method in batch

ramiadavid:
Fix Odoo warning:

```
The model odoo.addons.account_banking_mandate.models.account_banking_mandate is not overriding the create method in batch 
The model odoo.addons.account_banking_mandate.models.account_move is not overriding the create method in batch 
```


pedrobaeza (approved):
/ocabot merge patch

Issue #896, opened 2022-02-03
Error when add new payment method


rr10:
## Module
account.payment.mode odoo 15

https://user-images.githubusercontent.com/26082390/152379331-605c08fb-2fe9-4f37-9015-6874b5982909.mp4

RPC_ERROR
Odoo Server Error
Traceback (most recent call last):
  File "/opt/odoo15/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
    result = request.dispatch()
  File "/opt/odoo15/odoo/odoo/http.py", line 687, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo15/odoo/odoo/http.py", line 359, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo15/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo15/odoo/odoo/http.py", line 348, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo15/odoo/odoo/http.py", line 916, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo15/odoo/odoo/http.py", line 535, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo15/odoo/addons/web/controllers/main.py", line 1335, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo15/odoo/addons/web/controllers/main.py", line 1327, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo15/odoo/odoo/api.py", line 462, in call_kw
    result = _call_kw_model_create(method, model, args, kwargs)
  File "/opt/odoo15/odoo/odoo/api.py", line 442, in _call_kw_model_create
    result = method(recs, *args, **kwargs)
  File "<decorator-gen-189>", line 2, in create
  File "/opt/odoo15/odoo/odoo/api.py", line 412, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo15/odoo/addons/account/models/account_payment_method.py", line 28, in create
    if information.get('mode') == 'multi':
Exception

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/odoo15/odoo/odoo/http.py", line 643, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
[...]

PR #1079, opened 2023-05-10
[FIX] Fixed Error when create new payment method

Nikul-OSI:
Fixed https://github.com/OCA/bank-payment/issues/896

**Steps to reproduce:**
1. Go to Accounting/Invoicing -> Configuration -> Payment Methods
2. Create new method

image

3. Save Record

ERROR:

image

### DIFF
--- a/account_payment_mode/models/account_payment_method.py
+++ b/account_payment_mode/models/account_payment_method.py
@@ -2,7 +2,8 @@
 # @author: Alexis de Lattre <alexis.delattre@akretion.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class AccountPaymentMethod(models.Model):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The video in the issue is broken. Can you please add the steps to reproduce the problem by text? Also a regression test would be welcome.

pedrobaeza:
And another, please follow the commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Nikul-OSI:
@pedrobaeza Already here https://github.com/OCA/bank-payment/issues/896, But different version.

When we try to create payment method that time got Error.

pedrobaeza:
The video in such issue is broken, so I can't see the steps to reproduce the problem.

Nikul-OSI:
Yes @pedrobaeza , In this video when create simple payment method that time broken. I added steps with error.

pedrobaeza:
The problem is not in this module, but in Odoo code:

https://github.com/odoo/odoo/blob/a7be194f731370db0d8b0a2075e67676708fbd94/addons/account/models/account_payment_method.py#L28

The previous line should be:

```
information = methods_info.get(method.code, {})
```

instead of just

```
information = methods_info.get(method.code)
```

so don't put here code that is not related.

Closing this.

Please open instead a PR in Odoo solving the problem. It's also happening in 15.0.

Nikul-OSI:
Yes, I knew, But in this code, I suggested to the user, they only set the **self._get_payment_method_information().keys()** code.

https://github.com/OCA/bank-payment/pull/1079/files#diff-1957a14450d19e93d0b1f1eb50d968293c0c998896604c36d5113a317028abc2R55

pedrobaeza:
This is not the place to do that. Better fix the source problem in odoo/odoo.

Nikul-OSI:
@pedrobaeza I created https://github.com/odoo/odoo/issues/121239

PR #1081, opened 2023-05-11
[15.0][FIX] account_banking_pain_base: Proper condition for AdrLine

pedrobaeza:
Forward-port of #1005 

- Have a partner with no zip code and no city.
- Include it in a payment order.
- Generate the SEPA XML file.

Current behavior: The file couldn't be generated because there's an empty <AdrLine> element.

We put properly parenthesis in the conditions for this to no happen.

@Tecnativa TT43260


pedrobaeza:
Fast-tracking the patch already merged in previous version:

/ocabot merge patch

PR #1083, opened 2023-05-12
[16.0][FIX] account_payment_order: Set payment method line for account.payment

pedrobaeza:
Forward-port of #1082 

On v15+, a new model account.payment.method.line is introduced for fine-graining the outstanding account, being `payment_method_id` a related field referred to this new model, so we need to change the previous approach to select the proper method line only if found. If not found, it will be auto-selected by standard code.

@Tecnativa


pedrobaeza:
Equaling branches with the patches:

/ocabot merge patch

PR #1084, opened 2023-05-15
[15.0][FIX] account_banking_sepa_credit_transfer: fix inactive currency error in the tests

pilarvargas-tecnativa:
In case the currency is not active, the test fails because an invoice cannot be validated with an inactive currency
image

cc @Tecnativa TT43341

@victoralmau @pedrobaeza please review


pedrobaeza (approved):
Fast-tracking the patch

/ocabot merge patch

PR #1093, opened 2023-06-06
[16.0][FIX] account_payment_mode: filter out unique payment methods

astirpe:
Fixes https://github.com/OCA/bank-payment/issues/977

Not sure if this fix is correct...


pedrobaeza (approved):
Seems OK

fkantelberg (approved):
Tested and works flawless.

Issue #1091, opened 2023-06-05
[16.0] account_payment_partner: partner_bank_id empty for out_invoice with l10n_ch


grindtildeath:
## Module
account_payment_partner

## Describe the bug
With the swiss localization, field account.move.partner_bank_id is used to generate QR reference and QR code for customer invoices: https://github.com/odoo/odoo/blob/16.0/addons/l10n_ch/models/account_invoice.py#L40-L44
This field is computed in standard to the company's partner bank account (again for customer invoices): https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L778-L783
Now, with the module account_payment_partner installed, the computation of this field is redefined to set a `False` value in case the move is not a vendor bill: https://github.com/OCA/bank-payment/blob/16.0/account_payment_partner/models/account_move.py#L89-L99

Looking at this piece of code, the assumption "No bank account assignation is done for out_invoice as this is only needed for printing purposes" is wrong in the context of swiss localization. I'm curious though to know how this "can conflict with SEPA direct debit payments" and if that's still the case since the original message was about avoiding to assign a value from the payment mode and not to remove a value computed by the super function: https://github.com/OCA/bank-payment/commit/fbb15b4d587cd3461c91ebbc62bf5b2870513dd5#diff-e896d047b041c58135104731ffd6608aa8f8cd47b83ca6f4822051b8be32818eL36-R39

FTR, this bug only appears in v16.0 because in previous versions, the field partner_bank_id was passed explicitly when creating a invoice from a sale order: https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order.py#L633

## To Reproduce
**Affected versions**: 16.0

1. Install l10n_ch and account_payment_partner
2. Define a bank account on the company partner
3. Create a customer invoice

**Expected behavior**
The customer invoice should use the bank account defined on the company partner to be able to be valid with swiss QR requirements.

**Additional context**

grindtildeath:
@carlosdauden @pedrobaeza @alexis-via 
Your help would be appreciated here, as I'm not sure what's the "good" solution to this.

Should we remove the override of the compute function as the conflict with SEPA might not exist anymore?

Could we define other conditions to avoid overwriting the value computed by Odoo in the super function?

yvaucher:
[...]

PR #1094, opened 2023-06-06
[16.0] account_payment_partner: Fix computation of partner_bank_id

grindtildeath:
The computation of partner_bank_id has not been migrated properly since the move from account.invoice to account.move.

Before, we had an onchange on partner_id that did the following:
  - If no partner or no company, remove payment mode - If type = in_invoice, remove partner_bank_id as well
  - If partner and company,
    - If type = in_invoice, set payment mode from partner - if payment mode's payment method has bank account required, set the partner_bank_id to the commercial partner's bank - else set partner_bank_id to False
    - If type = out_invoice, set payment mode from partner and do not assign partner_bank_id

We also had an onchange on payment_mode_id that did the following:
  - If payment mode and its payment_type=outbound and its payment method has bank account required, set partner_bank_id to False
  - If no payment mode, set partner_bank_id to False

Now that partner_bank_id is a computed field, we shouldn't set its value to False if the move is not a vendor bill or doesn't have a payment mode.

The computation of partner_bank_id must instead reflect the conditions we used to have, that is:
  - If we don't have a payment_mode, set value to False
  - If we have a payment mode
    - If type=in_invoice and payment mode's payment method has not bank account required, set partner_bank_id to False (as setting the bank account from the partner is done in the call to super)
    - If type=out_invoice
      - if payment mode's payment method has bank account required - set partner_bank_id to the bank account linked to the fixed journal of payment mode if defined - do not set anything if that isn't the case since setting the bank account from the company is done in the call to super)
      - if payment mode's payment method does not have bank account required, set partner_bank_id to False

Closes: #1091

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -90,12 +90,29 @@ class AccountMove(models.Model):
     def _compute_partner_bank_id(self):
         res = super()._compute_partner_bank_id()
         for move in self:
-            # No bank account assignation is done for out_invoice as this is only
-            # needed for printing purposes and it can conflict with
-            # SEPA direct debit payments. Current report prints it.
-            if move.move_type != "in_invoice" or not move.payment_mode_id:
+            payment_mode = move.payment_mode_id
+            if payment_mode:
+                if (
+                    move.move_type == "in_invoice"
+                    and payment_mode.payment_type == "outbound"
+                    and not payment_mode.payment_method_id.bank_account_required
+                ):
+                    move.partner_bank_id = False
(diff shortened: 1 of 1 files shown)


yvaucher (approved):
LGTM

To me it is legit to change this computation field to be commanded by the payment mode. And shouldn't be empty if the payment mode defines a fixed bank account :+1:

ivantodorovich (approved):
LGTM.

Perhaps add unit tests for the case(s) you're fixing.
It'd make the issue easier to understand and review IMO

PR #1106, opened 2023-06-24
[16.0][IMP] account_payment_line_create wizard - UI improvements

luc-demeyer:


luc-demeyer:
There are several issues with the 16.0  account_payment_line_create wizard which are resolved via this PR:

1) stack trace when clicking on the 'Add a line' in the move_line_ids M2M field

You get this stack trace when the required date field has not been filled in.
IMHO this an OWL bug but I fixed it by simply hiding this field as long as the date has not been completed which makes the look and feel of the wizard cleaner.
image

2) deprecated domain setting in onchange has been replaced by a computed domain

3) hide fields and buttons that make no sense when the filter options result in no transactions and replace this by a proper alter-info message for the end user
image

4) clean up UI (proper colspan for move_line_ids field

5) add 16.0 compliant hotkeys for the buttons

luc-demeyer:
I have rebased this PR and adapted the commit message since the stack trace has been fixed in the meanwhile by https://github.com/OCA/bank-payment/pull/1125

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -39,8 +40,14 @@ class AccountPaymentLineCreate(models.TransientModel):
         selection=[("same", "Same"), ("same_or_null", "Same or Empty"), ("any", "Any")],
     )
     move_line_ids = fields.Many2many(
-        comodel_name="account.move.line", string="Move Lines"
+        comodel_name="account.move.line",
+        string="Move Lines",
+        domain="[('id', 'in', selectable_aml_ids)]",
```
pedrobaeza: This is totally anti-perfomant when this list gets bigger. Please use another approach using an "absolute domain".

pedrobaeza (changes requested):

account_payment_order/wizard/account_payment_line_create.py
luc-demeyer: imho https://github.com/OCA/bank-payment/pull/1106/commits/0fc94a34651cf888f11df459833749a4dad6c564 fixes this potential performance issue.

pedrobaeza: I don't think so. The problem is that the web client has to download a very large string containing the whole IDs list, which takes a lot of time whilst this list gets bigger. Try this PR in a DB with let's say 5.000 AR/AP pending moves to see what I'm talking about.

luc-demeyer:
@pedrobaeza 
PR is imho now ready for merge (domain replaced by absolute domain).

account_payment_order/wizard/account_payment_line_create_view.xml
```diff
@@ -35,53 +39,86 @@
...
-                    />
                 </group>
+                <div
+                    id="no_date_selected"
+                    class="alert alert-info"
+                    role="alert"
+                    attrs="{'invisible': [('date_selected', '=', True)]}"
+                    colspan="2"
+                >
+                    <span>THE DATE FIELD HAS NOT BEEN SET</span>
```
pedrobaeza: Don't put everything in upper letters.

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -64,6 +71,28 @@ def default_get(self, field_list):
...
+        "invoice",
+        "target_move",
+        "allow_blocked",
+        "payment_mode",
+        "partner_ids",
+    )
+    def _compute_move_line_ids_domain(self):
+        domain = self._prepare_move_line_domain()
+        self.move_line_ids_domain = domain
+        self.has_open_items = self.env["account.move.line"].search_count(domain)
```
pedrobaeza: This one is draining too much resource while changing selection. I think you should remove it.

luc-demeyer:
@pedrobaeza 
imho we are ok now

ghost:
@luc-demeyer @pedrobaeza 
**Open question:** 
Does the addition of the field "has_open_items" really worth it in terms of performance ? 
Searching for accounting entries within the defined criteria every time a criteria is changed can lead to untimely loading from the user standpoint, no ?

Could adding a "Check for open entries" button be an alternative ?

pedrobaeza:
Yes, this is still not performance-wise, as it's doing a search each time, and having more clicks at the end. I still prefer the old method (although not ideal also).

luc-demeyer:
Imho this PR gives a better user experience without performance impact in the real world.

Let me explain:
A company with a high volume (e.g. 10 - 100K) of open vendor bills that need to be added manually into a payment order will not exist since it's not managable. High volume trading companies do an automated population of their payment orders.
For high volume direct debit it's in my experience also always automated.

Bills for regular business transactions are mostly added directly to payment orders during the vendor bill encoding process via the 'add to payment order' button.

Hence we are talking here about the remaining set of various bills that are added manually, that's probably worth a few milliseconds for a postgres query (we perform a _search to compute the domain and the ids list is not sent to the UI).

I think that we should focus on User Experience and not on a non-existing since hypothetical performance impact.

pedrobaeza:
I don't agree. I have a customer that generates 42k invoices, and do the SEPA DD order using this wizard only selecting the filters, clicking the "Add items", and clicking OK, so that's a clear regression for them.

luc-demeyer:
@pedrobaeza 
Hence my assumption was wrong, but I can hardly imagine that the majority of those SDD invoices cannot be added automatically to the payment order. This looks like a business opportunity to me: sell them a couple of days and they'll save a lot of manual labour effort. They can also do this operation automatically via services like https://www.twikey.com/.

Since the domain setting on @api.onchange is deprecated, we need to replace this code and I am not sure if we need to do special tricks for an exceptional customer who treats 42k invoices manually (daily ? weekly ? monthly ?).

A small thing we could do for the 'has_open_items' boolean is to compute this one seperate from the domain with _search, limit=1.

pedrobaeza:
I know direct DD services, but they have good fees with their current provider and don't want to change on that. They need very few time for doing this, so that's not a real problem (the generation of the records takes a lot of time though, but that's another history). Anyway, I would remove the manual entry on that wizard, removing the domain, and just keep the batch adding through criteria. For manual entries, we can do it for the order itself.

PR #1111, opened 2023-07-17
[FIX] account_banking_sepa_credit_transfer: required bank account

JaumeBforgeFlow:
Only raise UserError when the payment method requires a bank account. Before it was raising the error regardless if bank account was not required.

### DIFF
--- a/account_banking_sepa_credit_transfer/models/account_payment_order.py
+++ b/account_banking_sepa_credit_transfer/models/account_payment_order.py
@@ -174,7 +174,10 @@ class AccountPaymentOrder(models.Model):
                 instructed_amount.text = "%.2f" % line.amount_currency
                 amount_control_sum_a += line.amount_currency
                 amount_control_sum_b += line.amount_currency
-                if not line.partner_bank_id:
+                if (
+                    self.payment_method_id.bank_account_required
+                    and not line.partner_bank_id
+                ):
                     raise UserError(
                         _(
                             "Bank account is missing on the bank payment line "


GuillemCForgeFlow (approved):
Makes sense 👍🏽

AaronHForgeFlow (approved):
:+1:

MiquelRForgeFlow:
@JaumeBforgeFlow please do forward ports.

JaumeBforgeFlow:
> @JaumeBforgeFlow please do forward ports.

Here you have them: https://github.com/OCA/bank-payment/pull/1114, https://github.com/OCA/bank-payment/pull/1115 and https://github.com/OCA/bank-payment/pull/1116

pedrobaeza:
This is not correct, and it should be reverted. The bank account required flag is for invoices, not for payment lines. If you want that, you should add a new flag (and it should be at payment mode level IMO), but not corrupt the function of the current one.

JordiMForgeFlow:
@pedrobaeza I can see that the flag is used for payment lines here (https://github.com/OCA/bank-payment/blob/21a1a1a8dbf12348422bf14d4ebe60b26cc49782/account_payment_order/models/account_payment_line.py#L28) and then used to do the check on the payment line here (https://github.com/OCA/bank-payment/blob/21a1a1a8dbf12348422bf14d4ebe60b26cc49782/account_payment_order/models/account_payment_line.py#L153). Why is it not valid to use it?

PR #1113, opened 2023-07-19
[14.0][FIX]account_payment_order_notification: Email template fix

manuelregidor:
The email template is not properly formed as a "for" loop adding rows to a table cannot be included into an email template. As a result, the "for" section of the template is automatically moved to the top, which generates an error at rendering. This fix sorts that problem out.


manuelregidor:
@ValentinVinagre @HaraldPanten

HaraldPanten:
@pedrobaeza That might be interesting for you, as well

HaraldPanten (approved):

ValentinVinagre (approved):
LGTM 👍🏻

account_payment_order_notification/data/mail_template_data.xml
```diff
@@ -14,36 +14,40 @@
...
-            <tr>
-                % if payment_line.move_line_id.move_id.move_type in ("in_invoice", "in_refund") and payment_line.move_line_id.move_id.ref:
-                    <td>${payment_line.move_line_id.move_id.ref}</td>
-                % else:
-                    <td>${payment_line.communication}</td>
-                % endif
-                <td>${format_date(payment_line.date)}</td>
-                <td>${format_amount(payment_line.amount_currency, payment_line.currency_id)}</td>
+            <tr style="border-bottom:2px solid #dee2e6;">
+                <td style="width:40%;"><strong>Number</strong></td>
```
pedrobaeza, line 28: Please don't force sizes.

pedrobaeza (dismissed):

account_payment_order_notification/data/mail_template_data.xml
manuelregidor: My solution is based on other Odoo email templates such as "Sales Order: Confirmation Email" template, which also use tables to show lines . As you can see in that template, since multiple tables are used, the width is forced so all the cells are aligned. It would be appreciated if you could suggest a different way to fix the problem using tables but not forcing the cells' width. Thank you.

pedrobaeza: The problem indicated in the PR has nothing to do with this (or I don't see the relation), so please limit the fix to that problem or separate it in different commits or PRs, but anyway documenting all of them properly.

HaraldPanten: Pedro, we're not sure on how to solve the issue we found, without forcing what we are doing. How can we set the correct structure without forcing the sizes as Manuel has done?

Do you mean that we can do that FIX (as Manuel did), but only changing the description/commit message?

THX.

pedrobaeza: Can you please rephrase the problem? I don't understand it quite well. Maybe with screenshots...

ValentinVinagre: Hi Pedro,
The problem is that it is not possible to perform a %for inside `<table>.` Therefore, it has to be extracted, when extracting that %for from the `<table>`s it is not possible to perform the traversal through the `<td>` tag and therefore 2 `<table>`s have to be performed separately. the first `<table>` is used for the headers and the second `<table>` for the lines containing the dynamic information. Since they are `<table>` different if the size of the `<td>` is not forced, it is not possible to have an equal size in both tables, that's where the force comes from.
What do u think?

victoralmau:
Sorry for the confusion, but I don't see the for moving, i.e. IMO the table is displayed correctly (without this change).

**Chatter**:
ejemplo-1

**MailHog**:
ejemplo-2

**Mail wizard**:
ejemplo-3

**Html**:
ejemplo-4

manuelregidor:
@victoralmau Maybe you're generating the email using a translated term, and translated terms do work. This is the code of the template in the runbot:

`<div style="font-family:'Lucida Grande', Ubuntu, Arial, Verdana, sans-serif;font-size: 12px; color: rgb(34, 34, 34); background-color: #FFF; ">
    <p>Hello ${object.partner_id.name or ''},</p>
    % if object.order_id.payment_type == 'inbound':
        <p>A debit order including at least one of your invoices has been processed and sent to the bank.</p>
    % else:
        <p>A payment order including at least one of your invoices has been processed and sent to the bank.</p>
    % endif
    <p>It includes the following transactions:</p>
    
        % for payment_line in object.payment_line_ids:
            
                % if payment_line.move_line_id.move_id.move_type in ("in_invoice", "in_refund") and payment_line.move_line_id.move_id.ref:
                    
                % else:
                    
                % endif
                
        % endfor
        <table>
        <thead>
            <tr style="height:15px;background-color: #d8d8d8;font-size: 13px;font-weight:bold;" align="center" valign="middle">
                <td>Number</td>
                <td>Date</td>
                <td>Amount</td>
            </tr>
        </thead>
        <tbody><tr><td>${payment_line.move_line_id.move_id.ref}</td><td>${payment_line.communication}</td><td>${format_date(payment_line.date)}</td>
                <td>${format_amount(payment_line.amount_currency, payment_line.currency_id)}</td>
            </tr></tbody>
    </table>
    <p>Do not hesitate to contact us if you have any questions.</p>
</div>`

As you can see, the for loop is displaced. However, if you generate a translated term of the malformed template code, it works properly.

pedrobaeza:
Even if the loop is misplaced, that doesn't mean that you need to put fixed widths. Just put the for in the proper place for iterating `<tr>` elements.

ValentinVinagre:
> Even if the loop is misplaced, that doesn't mean that you need to put fixed widths. Just put the for in the proper place for iterating `<tr>` elements.
Commented here 😄 
https://github.com/OCA/bank-payment/pull/1113#discussion_r1270336736

account_payment_order_notification/data/mail_template_data.xml
pedrobaeza: It has been possible always. I don't get why you say it isn't. Check for example: https://github.com/odoo/odoo/blob/4c3b6ad80ffd9015e7dfcb9e93a6455cbc319c90/addons/website_sale/data/mail_template_data.xml#L41

manuelregidor: In that example tr elements are not inserted straight into a table through the for loop, but instead a table is inserted inside another table on each iteration. Please, also note that because of that, td width is forced.

pedrobaeza: Well, if you need to insert such table here, please do it, but I still don't understand why the example put by Víctor is working. If it's something of the translation, please check Spanish translation and copy its syntax. If not possible, put the "table in table" trick without sizes to see.

victoralmau: Ok, i think i remember what happens. if we try to edit the email template in odoo it automatically moves the `for` before the table. 

This is something that happens with other odoo email templates (I don't remember an example right now, but I'll look for an example).

pedrobaeza: But email templates never should be edited with the rich text editor, but with HTML code.

manuelregidor: It happens when you edit it with the HTML editor. For some reason, that structure does not work and the "for" loop is displaced and I think it should be fixed (as @victoralmau said, there are other examples in which that very same structure is used and do not work either). My proposal is based on other email templates that do work. If any of you have a better idea idea in which column width is not forced, please, let me know and I will make the changes.

victoralmau: The same problem occurs in v14 at https://github.com/odoo/odoo/blob/14.0/addons/lunch/data/lunch_data.xml#L112 but it is already solved in v15 https://github.com/odoo/odoo/blob/15.0/addons/lunch/data/mail_template_data.xml#L67.

Do you think it is necessary to redo the email content to edit the template in odoo and avoid this problem that in v15 is already solved?

HaraldPanten: It's a good point having it solved in latest versions, but I think that users should be able to edit the email translations after the corresponding training. 

IMO we should apply the FIX for V14.

manuelregidor: @victoralmau It is exactly the same problem you mentioned in your comment. However, I don't think that a solution in the style of this one https://github.com/odoo/odoo/blob/15.0/addons/lunch/data/mail_template_data.xml#L67  can be applied to an email template in v14. That's why my solution is based on email templates present in v14 base Odoo modules.

ValentinVinagre: @pedrobaeza @victoralmau We do not want to leave this point hanging. Do you see this as a viable solution?

pedrobaeza:
OK, if you have it clear there's no other option, let's go:

/ocabot merge minor

PR #1118, opened 2023-07-21
[15.0][IMP] account_payment_order: Add a custom message if the invoice does not have a payment method.

victoralmau:
Add a custom message if the invoice does not have a payment method.

The message will be displayed when trying to create a line of a payment order to better clarify the reason.

@Tecnativa TT44551

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -57,16 +57,21 @@ class AccountMove(models.Model):
         for move in self:
             if move.state != "posted":
                 raise UserError(_("The invoice %s is not in Posted state") % move.name)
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
There's already a check for no payment mode, but the code to get there is not correct and it should be done first:

https://github.com/OCA/bank-payment/pull/1118/files#diff-bb6441f4ea06462fe1aae1f093f82c4e57460864b29b209676a6c7ce14e49ef9L81

victoralmau:
Changes done.

pedrobaeza (approved):

carlosdauden (approved):

PR #1121, opened 2023-07-25
[14.0][FIX] account_payment_order: incorrect inbound partial reconciliations

hildickethan:
- Fix a bug caused by https://github.com/OCA/bank-payment/commit/00d9e89f3d9300b4cc8b831876fc72befcea18ca changes that only partially reconciles moves that should be fully reconciled.
- Added a test that fails without the change, as the inbound reconciles to partial when it should be full
- Swapped to `float_compare` to avoid possible rounding errors

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -8,6 +8,7 @@ import base64
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare
 
 
 class AccountPaymentOrder(models.Model):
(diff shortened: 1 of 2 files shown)


account_payment_order/models/account_payment_order.py
```diff
@@ -433,7 +433,10 @@ def generated2uploaded(self):
             for line in payment.payment_line_ids:
                 if not line.move_line_id:
                     continue
-                if line.amount_currency != -line.move_line_id.amount_residual_currency:
+                sign = -1 if payment.payment_order_id.payment_type == "outbound" else 1
+                if line.amount_currency != (
```
pedrobaeza: Should we take the opportunity to use `float_compare` here?

hildickethan: Yes that's probably a good idea just in case

hildickethan: @pedrobaeza Swapped

pedrobaeza (approved):

ioans73 (approved):

pedrobaeza:
/ocabot merge patch

hildickethan:
> Please fw-port it to upper versions.

- https://github.com/OCA/bank-payment/pull/1127
- https://github.com/OCA/bank-payment/pull/1128
15 one is mergeable, 16 tests aren't cooperating

PR #1122, opened 2023-07-25
[16.0][IMP] account_payment_order_grouped_output: Don't generate grouped move when 1 line

pedrobaeza:
@Tecnativa


pedrobaeza:
Fast-tracking, as already merged in previous branches + it has tests:

/ocabot merge patch

PR #1123, opened 2023-07-26
[FIX] `account_banking_sepa_direct_debit`: invalid field used to check mandate

acsonefho:
The field `payment_ids` doesn't exist on `account.payment.order` model.

https://github.com/OCA/bank-payment/pull/979#discussion_r1274391891


pedrobaeza:
You can see here that the field exists:

https://github.com/OCA/bank-payment/blob/14.0/account_payment_order/models/account_payment_order.py#L124

PR #1125, opened 2023-07-26
[16.0][FIX] account_payment_partner: fetchPreloadedData bug using selection…

ljsalvatierra-factorlibre:
… widget

When clicking on `CREATE TRANSACTIONS` without a Due Date or Journal Entry Date.

imagen

imagen

imagen

```js
Uncaught (in promise) Error: The following error occurred in onWillStart: ""
    OwlError owl.js:87
    wrapError owl.js:2607
    onWillStart owl.js:2650
    useModel model.js:141
    setup list_controller.js:65
    ComponentNode owl.js:2359
    createComponent owl.js:5779
    slot1 owl.js:5553
    callSlot owl.js:2971
    template owl.js:5553
    _render owl.js:1725
    render owl.js:1717
    initiateRender owl.js:2381
Caused by: Error: 
    UnimplementedRouteError sample_server.js:14
    mockRpc sample_server.js:156
    fakeRPC sample_server.js:825
    call orm_service.js:109
    preloadSelection selection_field.js:89
    fetchPreloadedData relational_model.js:856
    loadPreloadedData relational_model.js:866
    _load relational_model.js:1199
```

This is caused when using a `widget` that is preloaded on data registry:

https://github.com/OCA/OCB/blob/16.0/addons/web/static/src/views/relational_model.js#L864

```js
async loadPreloadedData() {
...
            // @FIXME type should not be get like this
            const type = activeField.widget || this.fields[fieldName].type;
            if (!this.isInvisible(fieldName) && preloadedDataRegistry.contains(type)) {
                proms.push(fetchPreloadedData(preloadedDataRegistry.get(type), fieldName));
            }
...
```

### DIFF
--- a/account_payment_partner/views/account_move_line.xml
+++ b/account_payment_partner/views/account_move_line.xml
@@ -15,7 +15,6 @@
                     <field name="reconciled" invisible="1" />
                     <field
                         name="payment_mode_id"
-                        widget="selection"
                         force_save="1"
                         attrs="{'invisible': [('account_type', 'not in', ['asset_receivable', 'liability_payable'])], 'readonly': ['|', ('account_type', 'not in', ['asset_receivable', 'liability_payable']), ('reconciled', '=', True)]}"
                     />
@@ -34,7 +33,6 @@
                 <field
                     name="payment_mode_id"
                     optional="hide"
-                    widget="selection"
                     force_save="1"
                     attrs="{'invisible': [('account_type', 'not in', ['asset_receivable', 'liability_payable'])], 'readonly': ['|', ('account_type', 'not in', ['asset_receivable', 'liability_payable']), ('reconciled', '=', True)]}"
                 />


JavierIniesta (approved):
LGTM

Gelo-fl (approved):
LGTM  https://www.awesomescreenshot.com/video/19973755?key=dda9126e11c2405853dd2f6b19acd6b6

pedrobaeza (changes requested):
The solution should be to put the field due date as required.

ljsalvatierra-factorlibre:
> The solution should be to put the field due date as required.

The field `due_date` is required when selected the option. The bug appears when selected both options, Date Due or Journal Entry Date.

imagen

I explain the origin of the bug in the description of the PR.

pedrobaeza:
I see in the text `When clicking on CREATE TRANSACTIONS without a Due Date.`, so then hide the button `Create transactions` if no date is introduced, but the other is a change with too much impact, and this is anyway something to be fixed on Odoo core.

I introduce @alexis-via and @sbidoul here, which are core contributors of this module, to see if there's a strong reason to keep the widget selection on the payment modes. Maybe switching to m2o is better, and we avoid the problem.

astirpe:
We faced this issue with module `l10n_nl_tax_statement`, when clicking on the drilldown icon that opens the move lines list.
Screenshot from 2023-09-20 14-24-49

The cause is `widget="selection"` of `account_payment_partner`. Removing the widget is solving the problem. I hope this PR is merged soon. Thank you!

astirpe (approved):

pedrobaeza (approved):
OK, let's go with it. I think this is solving the symptom, not the cause, that comes from bad OWL conversion of the widget, but this is really not a big deal, as many2one widget is also OK on that fields.

/ocabot merge patch

PR #1049, opened 2023-02-24
[14.0] account_payment_order: Fix grouped partial reconcile

dannyadair:
When a payment order uses a payment mode with "Group Transactions in Payment Orders", and multiple invoices for the same supplier are added, reducing e.g. the first invoice payment used to result in the last one to be reconciled against the reduced amount(s) instead.

We first do partial reconciliations for over-/underpayments, only then pass on to .reconcile() for automatically reconciling the rest.

PR #1128, opened 2023-07-31
[16.0][FW] account_payment_order: Fix grouped partial reconcile + incorrect inbound partial reconciliations

hildickethan:
Forward ports of
- https://github.com/OCA/bank-payment/pull/1049
- https://github.com/OCA/bank-payment/pull/1121

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -444,15 +444,42 @@ class AccountPaymentOrder(models.Model):
         return action
 
     def generated2uploaded(self):
+        """Post payments and reconcile against source journal items
+
+        Partially reconcile payments that don't match their source journal items,
+        then reconcile the rest in one go.
+        """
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Sorry, I missed this one.

Merging as already merged in previous versions:

/ocabot merge patch

pedrobaeza:
@hildickethan-S73 please rebase and check the CI problem.

hildickethan:
> @hildickethan-S73 please rebase and check the CI problem.

The issue is that the 100€ lines are 115 due to the default 15% tax, I could remove the taxes from the invoices in that test to get by but it feels like covering up some underlying difference in this module between 15.0 and 16.0.
If I remove the taxes at the `setupclass` level since that would be the most normal difference, it fails in another one `test_supplier_manual_refund` to `No pending AR/AP lines to add on BILL/2023/12/0001`, which is what is making doubt

PR #1129, opened 2023-08-03
[13.0][FIX] account_banking_sepa_credit_transfer: required bank account UserError

GuillemCForgeFlow:
Reopen https://github.com/OCA/bank-payment/pull/1111 as it should be fine to use the `bank_account_required` flag from the `account.payment.method` model as it is already in use in other places for payment lines: https://github.com/OCA/bank-payment/pull/1111#issuecomment-1663615139

cc @JaumeBforgeFlow @JordiMForgeFlow

### DIFF
--- a/account_banking_sepa_credit_transfer/models/account_payment_order.py
+++ b/account_banking_sepa_credit_transfer/models/account_payment_order.py
@@ -174,7 +174,10 @@ class AccountPaymentOrder(models.Model):
                 instructed_amount.text = "%.2f" % line.amount_currency
                 amount_control_sum_a += line.amount_currency
                 amount_control_sum_b += line.amount_currency
-                if not line.partner_bank_id:
+                if (
+                    self.payment_method_id.bank_account_required
+                    and not line.partner_bank_id
+                ):
                     raise UserError(
                         _(
                             "Bank account is missing on the bank payment line "


pedrobaeza:
SEPA credit transfer requires the partner bank account. Why do you want to overpass such check? You'll end up with an invalid XML file.

JordiBForgeFlow:
@pedrobaeza the problem is that pain001 is more than SEPA credit transfer, and Odoo does not support a more generic pain001 credit transfer. pain001 is an ISO standard https://wiki.xmldation.com/General_Information/Payment_Standards/ISO_20022/pain.001

For example you could use pain001 for a company in EU to pay to a supplier in India, but you could not use SEPA, because SEPA would only be possible if you pay in EUR, within the EU.

pain001, allows payments to other companies, allowing different ways to identify the creditor and debitor accounts. In some countries like in Denmark companies can pay to each other using something called the FIK number, which is associated to each invoice, and is not a bank account number.

So, either we implement an account_banking_pain_001_credit_transfer, or we need to alter account_banking_sepa_credit_transfer to accept other ways to send payments other than EUR and within EU

pedrobaeza:
So my initial claim was right: you have to develop something extra (a config option, glue module...), but not disable this by default based on that flag, that has other meaning. There are some conditional checks around the code based on the PAIN standard. You can use such mechanism and add the standard PAIN in the account.payment.method, just like Switzerland does here: https://github.com/OCA/l10n-switzerland/tree/12.0/l10n_ch_pain_credit_transfer

JordiMForgeFlow:
@pedrobaeza thank you for the feedback and the reference.

We will look into handling it in a separated module. I am closing the PR

PR #1134, opened 2023-08-25
[IMP] Improving tests account_banking_sepa_credit_transfer

BT-rmartin:
By not reassigning the company of the partners in the test the current tests continue working whilst combining this module with account_banking_mandate doesn't lead to a problem of "You cannot change the company of Partner Bank %s, as there exists mandates referencing it that belong to another company"


BT-rmartin:
@dzungtran89 @ValentinVinagre @CarlosRoca13 Could you please review this tiny change?

pedrobaeza (changes requested):
This is not correct, as that code is for avoiding other problems with several OCA modules combined together. You must patch the other module, or better, remove demo data in these tests and use data created on the fly from scratch.

PR #1135, opened 2023-08-31
[14.0][IMP] account_payment_sale: add bank accounts to report_saleorder_document

hparfr:
Inspired of what is done in invoices (account_payment_partner)

### DIFF
--- a/account_payment_sale/README.rst
+++ b/account_payment_sale/README.rst
@@ -69,6 +69,7 @@ Contributors
 * Danimar Ribeiro
 * Raphaël Valyi
 * Raf Ven <raf.ven@dynapps.be>
+* Raphaël Reverdy <raphael.reverdy@akretion.com>
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


nayatec (approved):

pedrobaeza:
Why not extracting this common logic for not having duplicated code here and in the payment order module, and also create a separated QWeb template with the same content and call such template from both the invoice and the sale?

hparfr:
>Why not extracting this common logic for not having duplicated code here and in the payment order module, 

I have the feeling  acount_payment_partner needs some code cleaning / doc. It's not obivious for me why there is this all this logic with partner_bank_id or the maybedeadcode self.mandate_id

I prefer introduce this code as is, with this trade off of few duplicate lines and may be proprose some day a refactoring, looks to me the odds to have this merged are heigher that way.

>and also create a separated QWeb template with the same content and call such template from both the invoice and the sale?

I have seen a multiple times where the sale and the invoice reports diverge slightly and this kind of common code are at the end more work.

pedrobaeza:
Well, that's not the case here, so I think it's worth.

Issue #1131, opened 2023-08-14
V16 account_banking_sepa_direct_debit


cesarsc79:
## Module
account_banking_sepa_direct_debit

## Describe the bug

When the SEPA file is generated, the debtor's IBAN is filled in with the creditor's IBAN.
Debtors' IBANs do not appear in the XML file

## To Reproduce
V16:

generate a SEPA DEBITOR xml file

**Expected behavior**
 the debtor's IBAN is filled in with the debtor's  IBAN

10537:
I confirm that, we have the same error. @rvalyi @ovnicraft @azmeuk

10537:
Hi @cesarsc79 

You can used this [PULL-1136](https://github.com/OCA/bank-payment/pull/1136) for resolve your problem until this will be merge to the branch 16.0

best regards,

cesarsc79:
> Hi @cesarsc79
> 
> You can used this [PULL-1136](https://github.com/OCA/bank-payment/pull/1136) for resolve your problem until this will be merge to the branch 16.0
> 
> best regards,

Thanks

PR #1136, opened 2023-09-01
[16.0][FIX]Fixed error on XML creation data in account_banking_sepa_direct_debit.

10537:
This changes fix a error when the xml is create, you can view more details in this issue: [Issue 1131](https://github.com/OCA/bank-payment/issues/1131)

### DIFF
--- a/account_banking_sepa_direct_debit/models/account_payment_order.py
+++ b/account_banking_sepa_direct_debit/models/account_payment_order.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from lxml import etree
-
 from odoo import _, exceptions, fields, models
 from odoo.exceptions import UserError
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
This solution is not correct. It's forcing to use the first bank account of the partner, when you have the possibility of choosing another. Your problem probably comes from having incorrectly filled the bank account on your invoice.

cesarsc79:
Pedro:

The bank account is filled correctly on the invoice

pedrobaeza:
And in the payment line?

cesarsc79:
The bank account is filled correctly in the payment line too

pedrobaeza:
Then this line should dump such bank account to the XML file:

https://github.com/OCA/bank-payment/blob/927df0fe880e329abb2b9ca449b2c28831705f4e/account_banking_sepa_direct_debit/models/account_payment_order.py#L251

PR #1137, opened 2023-09-04
[14.0][FIX] account_payment_order: change result_payorder_ids to list to show action correctly

pedrobaeza:
Forward-port of #1013

@Tecnativa


pedrobaeza:
Fast-tracking, as already merged in previous version and the problem persists here:

/ocabot merge patch

PR #1138, opened 2023-09-04
[15.0][FIX] account_payment_order: change result_payorder_ids to list to show action correctly

pedrobaeza:
Forward-port of #1013 and #1137

@Tecnativa

PR #1139, opened 2023-09-04
[16.0][FIX] account_payment_order: change result_payorder_ids to list to show action correctly

pedrobaeza:
Forward-port of #1013, #1137 and #1138

@Tecnativa


pedrobaeza:
The last one in the chain:

/ocabot merge patch

PR #1141, opened 2023-09-12
[IMP]sepa credit transfer : allow to customise transaction grouping key

luc-demeyer:
For a project with JP Morgan I have been asked to move the PmtTpInf block from the PmtInf block to the CdtTrfTxInf block (hence move e.g. InstrPrty or LclInstrm fown to the transaction level).

According to  pain.001.001.03.xsd both places are allowed.
According to the German subset pain.001.001.03_GBIC_3.xsd the current implement is the recommended one.

This PR adds a method allowing to customise (move) this PmtTpInf  block without changing the default behaviour (without inherit the generated pain file remains unchanged).


luc-demeyer:
@pedrobaeza 
can you remove the 'stale' label ?

PR #1142, opened 2023-09-19
[14.0][FIX+IMP] account_payment_order: Define the value of the communication field correctly

victoralmau:
Superseed https://github.com/OCA/bank-payment/pull/1071

Changes done:
- [x] Define the value of the `communication` field correctly
- [x] Add `payment_ref` field to `account.payment` tree and form views

Example use case:
factura
order-pago-1
orden-pago-2
orden-pago-3

@Tecnativa TT45153

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -39,6 +39,31 @@ class AccountMove(models.Model):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _get_payment_order_communication(self):
+        """
+        Retrieve the communication string for the payment order
+        """
+        communication = self.payment_reference or self.ref or self.name or ""
+        if self.is_invoice():
+            if (self.reference_type or "none") != "none":
+                communication = self.ref
+            elif self.is_purchase_document():
+                communication = self.ref or self.payment_reference
+            else:
+                communication = self.payment_reference or self.name
+        # If we have credit note(s) - reversal_move_id is a one2many
+        if self.reversal_move_id:
+            references = []
+            references.extend(
+                [
+                    move._get_payment_order_communication()
+                    for move in self.reversal_move_id
+                    if move.payment_reference or move.ref
+                ]
+            )
+            communication += " " + " ".join(references)
+        return communication
+
     def _prepare_new_payment_order(self, payment_mode=None):
         self.ensure_one()
         if payment_mode is None:
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -53,16 +53,7 @@ class AccountMoveLine(models.Model):
         # Build a recordset to gather moves from which references have already
         # taken in order to avoid duplicates
         reference_moves = self.env["account.move"].browse()
-        # If we have credit note(s) - reversal_move_id is a one2many
-        if self.move_id.reversal_move_id:
-            references.extend(
-                [
-                    move.payment_reference or move.ref
-                    for move in self.move_id.reversal_move_id
-                    if move.payment_reference or move.ref
-                ]
-            )
-            reference_moves |= self.move_id.reversal_move_id
+        reference_moves |= self.move_id.reversal_move_id
         # Retrieve partial payments - e.g.: manual credit notes
         for (
             _,
@@ -73,7 +64,7 @@ class AccountMoveLine(models.Model):
             if payment_move not in reference_moves and (
                 payment_move.payment_reference or payment_move.ref
             ):
-                references.append(payment_move.payment_reference or payment_move.ref)
+                references.append(payment_move._get_payment_order_communication())
         return references
 
     def _get_communication(self):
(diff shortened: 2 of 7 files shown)


account_payment_order/models/account_move_line.py
```diff
@@ -57,7 +57,11 @@ def _get_linked_move_communication(self):
         if self.move_id.reversal_move_id:
             references.extend(
                 [
-                    move.payment_reference or move.ref
+                    (
+                        move.ref
```
pedrobaeza: It's not mandatory to fill the field:

```suggestion
                        move.ref or move.payment_reference
```

account_payment_order/models/account_move_line.py
```diff
@@ -73,7 +77,11 @@ def _get_linked_move_communication(self):
             if payment_move not in reference_moves and (
                 payment_move.payment_reference or payment_move.ref
             ):
-                references.append(payment_move.payment_reference or payment_move.ref)
+                references.append(
+                    payment_move.ref
```
pedrobaeza:
```suggestion
                    payment_move.ref or payment_move.payment_reference
```

account_payment_order/models/account_move_line.py
```diff
@@ -95,7 +103,7 @@ def _get_communication(self):
                     self.move_id.move_type in ("in_invoice", "in_refund")
                     and self.move_id.ref
                 ):
-                    communication = self.move_id.payment_reference or self.move_id.ref
+                    communication = self.move_id.ref
```
pedrobaeza: Why this is removed and why so many places to take the same thing? Shouldn't be one of there should be a reason to be some of them, and you need to take each use case apart.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -407,6 +407,7 @@ def test_supplier_refund_reference(self):
         self.refund = self._create_supplier_refund(self.invoice)
         with Form(self.refund) as refund_form:
             refund_form.ref = "R1234"
+            refund_form.payment_reference = "FR/1234"
```
pedrobaeza, line 424: This is not tested anywhere

account_payment_order/views/account_payment_views.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_account_payment_tree" model="ir.ui.view">
+        <field name="name">account.payment.tree</field>
+        <field name="model">account.payment</field>
+        <field name="inherit_id" ref="account.view_account_payment_tree" />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <field name="payment_reference" optional="hide" />
```
pedrobaeza: This is being added to global payment tree view, so not sure to be done this way. Maybe a primary view and then it's shown always?

pedrobaeza:
I would like to rationalize a bit all of this communication thing, trying to centralize it, and counting with all the possible cases. Can you please do it?

account_payment_order/tests/test_payment_order_outbound.py
victoralmau: This data is set and then checked to make sure that this value has not been set in communication field.

victoralmau:
The code has been refactored a bit to set `communication` in a single method that clarifies the use cases, tests have been added to check the value in different use cases.

account_payment_order/models/account_move.py
```diff
@@ -39,6 +39,21 @@ def _compute_payment_order_ok(self):
...
+        """
+        communication = self.ref or self.name or ""
+        if self.is_invoice():
+            if (self.reference_type or "none") != "none":
+                communication = self.ref
+            elif self.is_purchase_document():
+                communication = self.ref or self.payment_reference
+            else:
+                # Force to only put invoice number here
+                communication = self.payment_reference or self.name
```
pedrobaeza, line 53: If we force the invoice number, then it should be `self.name` first.

account_payment_order/models/account_move_line.py
```diff
@@ -57,7 +57,7 @@ def _get_linked_move_communication(self):
         if self.move_id.reversal_move_id:
             references.extend(
                 [
-                    move.payment_reference or move.ref
+                    move._get_communication()
                     for move in self.move_id.reversal_move_id
                     if move.payment_reference or move.ref
```
pedrobaeza: This condition should be reflected in the method and removed.

account_payment_order/models/account_move.py
```diff
@@ -39,6 +39,21 @@ def _compute_payment_order_ok(self):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _get_communication(self):
```
pedrobaeza: Be more specific on the name, as the method is located in account.move:

```suggestion
    def _get_payment_order_communication(self):
```

account_payment_order/models/account_move.py
victoralmau: I agree with you, this is the same as it is currently (https://github.com/OCA/bank-payment/blob/14.0/account_payment_order/models/account_move_line.py#L101) and changing it will change the expected result so far, do we change it anyway (set `self.name` first) or leave it as it is by removing the comment to avoid confusion?

pedrobaeza: OK, let's remove the comment.

account_payment_order/models/account_move.py
```diff
@@ -39,6 +39,31 @@ def _compute_payment_order_ok(self):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _get_payment_order_communication(self):
+        """
+        Retrieve the communication string for the payment order
+        """
+        communication = self.ref or self.name or ""
```
pedrobaeza: I think we should finally complement this this way:

```suggestion
        communication = self.payment_ref or self.ref or self.name or ""
```

pedrobaeza (approved):

carlosdauden (approved):

PR #1143, opened 2023-09-19
[14.0][IMP] account_banking_mandate: Allow to set a specific mandate reference or sequence

victoralmau:
Changes done:
- [x] Allow sequence to be multi-company.
- [x] Allow to set a specific mandate reference or sequence.

@Tecnativa

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -61,7 +61,7 @@ class AccountBankingMandate(models.Model):
         default=lambda self: self.env.company,
     )
     unique_mandate_reference = fields.Char(
-        string="Unique Mandate Reference", tracking=10, copy=False
+        string="Unique Mandate Reference", tracking=10, copy=False, default="/"
     )
     signature_date = fields.Date(
         string="Date of Signature of the Mandate",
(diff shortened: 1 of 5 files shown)


account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -178,7 +178,7 @@ def _check_valid_state(self):
     @api.model
     def create(self, vals=None):
         unique_mandate_reference = vals.get("unique_mandate_reference")
-        if not unique_mandate_reference or unique_mandate_reference == "New":
+        if not unique_mandate_reference or unique_mandate_reference == "/":
```
pedrobaeza:
```suggestion
        unique_mandate_reference = vals.get("unique_mandate_reference", "/")
        if unique_mandate_reference == "/":
```

pedrobaeza (approved):

account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -177,8 +177,8 @@ def _check_valid_state(self):
 
     @api.model
```
pedrobaeza: Please take the occasion to convert this to multi-record (`@api.model_create_multi`)

PR #1144, opened 2023-09-20
Fix mandates tree view in res.partner.bank for better ergonomy

moctarjallo:


pedrobaeza:
Why is it better ergonomy? Please put screenshots of before and after. Isn't this simply to put `col=2` on existing definition?

moctarjallo:
`col=2` wouldn't work.

Here's what it looks like before and after.

Before
After

pedrobaeza (approved):
It works most of the times with col=2, but your proposal is also good.

bofiltd (approved):

bosd (approved):
LGTM

PR #1148, opened 2023-09-27
[16.0][IMP] account_payment_order: adds restriction for duplicated move lines in payment order

ghost:
Restriction added so that payment lines are not duplicated

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -221,6 +221,20 @@ class AccountPaymentOrder(models.Model):
                         )
                     )
 
+    @api.constrains("payment_line_ids")
+    def _check_payment_lines(self):
+        for order in self:
+            move_line_ids = [x.move_line_id.id for x in order.payment_line_ids]
+            if len(move_line_ids) != len(set(move_line_ids)):
+                raise ValidationError(
+                    _(
+                        "There are several lines pointing to the same pending "
+                        "balance. This is probably caused by a manual line creation. "
+                        "Please remove this duplication for being able to save the "
+                        "order."
+                    )
+                )
+
     @api.depends("payment_line_ids", "payment_line_ids.amount_company_currency")
     def _compute_total(self):
         for rec in self:
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -83,6 +83,16 @@ class TestPaymentOrderInboundBase(AccountTestInvoicingCommon):
 
 @tagged("-at_install", "post_install")
 class TestPaymentOrderInbound(TestPaymentOrderInboundBase):
+    def _line_creation(self, inbound_order):
+        vals = {
+            "order_id": inbound_order.id,
+            "partner_id": self.partner.id,
+            "currency_id": inbound_order.payment_mode_id.company_id.currency_id.id,
+            "amount_currency": 200.38,
+            "move_line_id": self.invoice.invoice_line_ids[0].id,
+        }
+        return self.env["account.payment.line"].create(vals)
+
     def test_constrains_type(self):
         with self.assertRaises(ValidationError):
             order = self.env["account.payment.order"].create(
@@ -136,3 +146,15 @@ class TestPaymentOrderInbound(TestPaymentOrderInboundBase):
         payment_order.cancel2draft()
         payment_order.unlink()
         self.assertEqual(len(self.payment_order_obj.search(self.domain)), 0)
+
+    def test_constrains_payment_line(self):
+        inbound_order = self.env["account.payment.order"].create(
+            {"payment_mode_id": self.inbound_mode.id, "journal_id": self.journal.id}
+        )
+        # Create two payment lines with the same
+        # move line id and try to assign them to inbound order
+        payment_line_1 = self._line_creation(inbound_order)
+        inbound_order.payment_line_ids |= payment_line_1
+        with self.assertRaises(ValidationError):
+            payment_line_2 = self._line_creation(inbound_order)
+            inbound_order.payment_line_ids |= payment_line_2
--- a/account_payment_order/tests/test_payment_order_outbound.py
+++ b/account_payment_order/tests/test_payment_order_outbound.py
@@ -225,6 +225,16 @@ class TestPaymentOrderOutbound(TestPaymentOrderOutboundBase):
         self.assertEqual(order.move_ids[0].date, order.payment_ids[0].date)
         self.assertEqual(order.state, "uploaded")
 
+    def _line_creation(self, outbound_order):
+        vals = {
+            "order_id": outbound_order.id,
+            "partner_id": self.partner.id,
+            "currency_id": outbound_order.payment_mode_id.company_id.currency_id.id,
+            "amount_currency": 200.38,
+            "move_line_id": self.invoice.invoice_line_ids[0].id,
+        }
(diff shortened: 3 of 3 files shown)


pedrobaeza:
AFAIK, this restriction already exists (but not implemented that way). And it's legit to have several times the same journal item if I don't remember bad. For example, a returned payment that is again added to another payment order.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -260,6 +260,19 @@ def test_constrains(self):
...
+                    "order_id": outbound_order.id,
+                    "partner_id": self.partner.id,
+                    "communication": "Test Outbound Payment Line " + str(i),
+                    "currency_id": outbound_order.payment_mode_id.company_id.currency_id.id,
+                    "amount_currency": 200.38,
+                    "move_line_id": self.invoice.invoice_line_ids[0].id,
+                }
+                payment_line = self.env["account.payment.line"].create(vals)
+                outbound_order.payment_line_ids |= payment_line
+                i += 1
```
aritzolea: This line is not necessary

ghost: Done!

aritzolea (approved):

SoniaViciana (approved):
LGTM

Hello @pedrobaeza 

Thanks for the review. This improvement prevents to being added the same journal item, multiple times in the same payment order. With the upgrade, we can do that using the "add a line" wizard, but at least, you get an error message at the confirmation of the order. Without the improvement, you can confirm an order like this:


imagen

If I review de journal items, there is only one of them (no 3):

imagen



On the other hand, the returned payment can be treated with another method. If i'm not mistaken, there are some modules for this:

imagen

Thanks for your help,

account_payment_order/models/account_payment_order.py
```diff
@@ -196,6 +196,26 @@ def check_date_scheduled(self):
...
+            payment_lines = order.payment_line_ids
+            move_lines_ids = payment_lines.mapped("move_line_id").ids
+            existing_move_lines = []
+            for move_line_id in move_lines_ids:
+                existing_move_line = payment_lines.filtered(
+                    lambda l: l.move_line_id.id == move_line_id
+                )
+                if len(existing_move_line) > 1:
+                    existing_move_lines.append(existing_move_line[0].communication)
+            if existing_move_lines:
```
pedrobaeza: Avoid `s` in the middle of the variable name, as they provoke often typos, and this is also very unperformant. You can avoid for this case to specify which is duplicated, as it's a very rough edge:

```suggestion
            move_lines = payment_lines.move_line_id
            if len(move_lines) != len(payment_lines):
```

ghost: Thank you Pedro for your time.
I have changed `move_lines_ids` to `move_line_ids` as you suggested.
I consider necessary to specify which line is being duplicated, as some payment orders can have numerous lines, and it would be harder to find the duplicated lines for the user.

pedrobaeza: But as said, this is anti-performant, with a quadratic complexity that cannot be accepted. If you don't find something with a linear complexity to get the same result, you should remove it.

ghost: OK, it has been modified. Thank you @pedrobaeza.

account_payment_order/models/account_payment_order.py
```diff
@@ -196,6 +196,12 @@ def check_date_scheduled(self):
                         )
                     )
 
+    @api.constrains("payment_line_ids")
+    def _check_payment_lines(self):
+        for order in self:
+            if len(order.payment_line_ids.move_line_id) != len(order.payment_line_ids):
+                raise ValidationError(_("Duplicated lines found."))
```
pedrobaeza: Be less cryptic:

```suggestion
                raise ValidationError(_("There are several lines pointing to the same pending balance. This is probably caused by a manual line creation. Please remove this duplication for being able to save the order."))
```

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -81,6 +81,16 @@ def _create_customer_invoice(self):
 
 @tagged("-at_install", "post_install")
 class TestPaymentOrderInbound(TestPaymentOrderInboundBase):
+    def line_creation(self, inbound_order):
```
pedrobaeza: Extracting this as a method, being used only in one place, and more a public one, is not very convenient. I think you can do it inline, but if you think it can be reused, please put it at least as private (with a leading `_`)

pedrobaeza:
I'm also thinking that the cases can be more complex, as you can be interested (by any reason) to put 2 times the same move line, but with amounts lower or equal than the total amount. Example: move line with a balance of 100 €. You do 2 lines with 50 € each. This can be legal. Not sure if it's a real case though.

SoniaViciana:
I can't think of any case in which you would want to duplicate lines in payment orders that don't correspond to accounting.

For the example shown (2 payments for the same item), the correct way would be to establish `date_maturity` at the `account.move.line`

pedrobaeza:
I know that it's very weird, but right now the constraint doesn't exist and they are able to do it. If we avoid this now, there can be side effects. Let's add the line commenting it in the ROADMAP.

account_payment_order/models/account_payment_order.py
```diff
@@ -196,6 +196,12 @@ def check_date_scheduled(self):
                         )
                     )
 
+    @api.constrains("payment_line_ids")
+    def _check_payment_lines(self):
+        for order in self:
+            if len(order.payment_line_ids.move_line_id) != len(order.payment_line_ids):
```
ghost: This constraint will break payment orders with payment lines without journal items. The journal item is not mandatory on the payment line. If a payment line is added manually without journal item, the constraint will be triggered.

ljsalvatierra-factorlibre:
Hi, we are going to work on this over the following weeks.

ljsalvatierra-factorlibre (approved):

aritzolea:
@pedrobaeza @jdidderen-noviat could you please review the last changes?

account_payment_order/models/account_payment_order.py
```diff
@@ -196,6 +196,23 @@ def check_date_scheduled(self):
                         )
                     )
 
+    @api.constrains("payment_line_ids")
+    def _check_payment_lines(self):
+        for order in self:
+            found_lines = self.env["account.move.line"]
+            for pay_line in order.payment_line_ids:
```
pedrobaeza: This is anti-performance, so it should be avoided. It also adds on each iteration the record to a recordset, having to recreate it and consuming memory. I think a better formula is:

```python
move_line_ids = [x.move_line_id.id for x in order.payment_line_ids]
if len(move_line_ids) != len(set(move_line_ids)):
    raise ...
```

pedrobaeza (changes requested):

aritzolea:
@pedrobaeza pre-commit is failing because a pre-commit failure in another module from this repo
`****po-python-parse-format****
account_payment_method_fs_storage/i18n/es.po:74 Translation string couldn't be parsed correctly using str.format KeyError('detalles') - [po-python-parse-format]`

pedrobaeza:
Can you please fix the branch in other PR ?

aritzolea:
> Can you please fix the branch in other PR ?

Branch has been fixed

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to v17

PR #1149, opened 2023-09-27
[16.0] Add warning when payment order is not SEPA

alexis-via:
The field 'sepa' on account.payment.order is only display for SEPA payment methods.
If the option "show warning if not SEPA" is enabled on the payment method, a warning banner is now displayed on payment orders with a SEPA payment method which are not SEPA.

### DIFF
--- a/account_banking_pain_base/models/account_payment_method.py
+++ b/account_banking_pain_base/models/account_payment_method.py
@@ -16,6 +16,7 @@ class AccountPaymentMethod(models.Model):
         "the corresponding unaccented character, so that only ASCII "
         "characters are used in the generated PAIN file.",
     )
+    warn_not_sepa = fields.Boolean(string="Warn If Not SEPA")
 
     def get_xsd_file_path(self):
         """This method is designed to be inherited in the SEPA modules"""
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -24,7 +24,12 @@ logger = logging.getLogger(__name__)
 class AccountPaymentOrder(models.Model):
     _inherit = "account.payment.order"
 
-    sepa = fields.Boolean(compute="_compute_sepa", readonly=True, string="SEPA Payment")
+    sepa = fields.Boolean(compute="_compute_sepa", string="SEPA Payment")
+    sepa_payment_method = fields.Boolean(
+        compute="_compute_sepa",
+        string="SEPA Payment Method",
+    )
+    show_warning_not_sepa = fields.Boolean(compute="_compute_sepa")
     charge_bearer = fields.Selection(
         [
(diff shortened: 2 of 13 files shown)


alexis-via:
Related v14 PR: https://github.com/OCA/bank-payment/pull/1050

alexis-via:
non_sepa_warn

account_banking_pain_base/models/account_payment_method.py
```diff
@@ -16,6 +16,7 @@ class AccountPaymentMethod(models.Model):
         "the corresponding unaccented character, so that only ASCII "
         "characters are used in the generated PAIN file.",
     )
+    warn_not_sepa = fields.Boolean(string="Warn If Not SEPA")
```
pedrobaeza, line 19: Again I think you don't need this flag. You can know which payment methods are SEPA ones without this flag, and this makes that any installation won't benefit of this feature if they don't configure it.

pedrobaeza (changes requested):

PaulGoubert (approved):
Working fine.   Thank you Alexis.

account_banking_pain_base/models/account_payment_method.py
alexis-via: True, a migration script was missing, to set it to true by default on existing installations for sepa_direct_debit and sepa_credit_transfer. I just added the migration script.

houzefa-abba (approved):

account_banking_pain_base/models/account_payment_method.py
alexis-via: FYI, I'm now doing some tests with another french integrator on ISO20022 wire transfers from France to non-SEPA countries. It requires 2 small additions in the XML, but works well with the module account_banking_sepa_credit_transfer. I plan to contribute it soon.
So I really think that account_banking_sepa_credit_transfer can be useful for users out of the SEPA zone to generate ISO 20022 files which are non-SEPA. And these users would be annoyed to always have the yellow banner that the payment order is not SEPA.

pedrobaeza: OK, then please put it as `default=True`, and I'm happy with this.

alexis-via: For some reasons that I can't explain, the migration scripts I talked about in my answer of september 30 where not present in this PR... but they were still on my hard drive. I just added them to the PR today.
It doesn't make sense to put it to default=True on the field level, because that field is on account.payment.method, and you have payment methods that have no link with sepa. But it is initialized to True in the 2 OCA modules that provide sepa payment methods and the migration scripts are provided to set it to True on existing Odoo databases.

pedrobaeza (approved):
OK then about the migration scripts + mark the field on data.

/ocabot merge minor

PR #1150, opened 2023-09-27
[16.0] account_payment_order: re-implement allow_blocked

alexis-via:
allow_blocked has been moved from the wizard that adds payment lines to the payment mode. That way, users cannot modify it. And they cannot by-pass it by using the button "add to payment order" on the invoice. The 'blocked' status is also check upon confirmation of the payment order (in case it has been marked as blocked AFTER being added to the payment order).

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -110,6 +110,18 @@ class AccountMove(models.Model):
                 for line in applicable_lines.filtered(
                     lambda x: x.payment_mode_id == payment_mode
                 ):
+                    if line.blocked and not payment_mode.allow_blocked:
+                        raise UserError(
+                            _(
+                                "The journal item '%(line)s' related to "
+                                "journal entry '%(move)s' is marked as litigation, "
+                                "and the payment mode '%(pay_mode)s' "
+                                "doesn't allow litigation lines.",
+                                line=line.display_name,
+                                move=move.display_name,
+                                pay_mode=payment_mode.display_name,
+                            )
+                        )
                     line.create_payment_line_from_move_line(payorder)
                     count += 1
                 if new_payorder:
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -177,6 +177,21 @@ class AccountPaymentLine(models.Model):
             )
         if not self.communication:
             raise UserError(_("Communication is empty on payment line %s.") % self.name)
+        if (
+            self.move_line_id
+            and self.move_line_id.blocked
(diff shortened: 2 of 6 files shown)


alexis-via:
If you need this on v14, it is integrated in this PR: https://github.com/OCA/bank-payment/pull/1050

pedrobaeza (changes requested):
I don't see this as an improvement, but a regression, as my users want to be able to select if they want to include them or not. You shouldn't remove it from the wizard, although you can add an extra option at payment mode level with 3 levels: yes, no or selectable, and act in consequence.

ghost (commented):
I'm a bit puzzled about this change. 

On one hand, I can see this as an improvement for people who wants to have more protections around this flag.

On the other hand, I also agree with Pedro that it can be see as a degression for some people.

An idea would be to have it computed at the payment order level based on the payment mode (in readonly = false and invisible) and keep the flag in the wizard so the user can change the flag on the payment order from the wizard. And hide the flag in the wizard behind a security groups or something like that in order of having both approaches working

pedrobaeza:
The three-valued selection I suggested can be compatible with this last suggestion: on payment mode, you decide which behavior for allow_blocked:

- Selectable. The default one. Everything continues the same, having it in the wizard.
- Yes. You allow them without asking.
- No. You don't allow them.

alexis-via:
Taking into account your comments, I changed the implementation:
Option "allow_blocked" on payment mode:
- if False : ALWAYS block, no workaround possible: blocked when trying to push an invoice on a payment order, block in the wizard to pull move lines on payment order (no option displayed to user), block upon confirmation of payment order
- if True : No systematic blocking. Show the option "allow_blocked" in the wizard, to give the option to the user to select blocked move lines if he wants to.

ghost (approved):
Thanks for the changes :)
LGTM

houzefa-abba (approved):

pedrobaeza:
For UX, I still prefer the three-value option. If not, people get confused about what means not having "Allow block" checked.

alexis-via:
I really prefer a solution with 2 possibilities instead of 3. The third option "allow without asking" doesn't make sense to me. It undermines the importance of having a litigation with a supplier (or a customer).

pedrobaeza:
Please check again my comment. I haven't said in any moment "allow without asking". The options are:


* Selectable. The default one. Everything continues the same, having it in the wizard.
* Yes. You allow them without asking.
* No. You don't allow them.

pedrobaeza:
The "yes" option makes sense to me, but if not, another rephrasing should be done for the boolean. You can put a selection with 2 options then.

luc-demeyer (approved):

PR #1147, opened 2023-09-22
[16.0][FIX] account_payment_order: Add extra form view only to avoid open move lines in edit mode

victoralmau:
FWP from 15.0: https://github.com/OCA/bank-payment/pull/1146

Add extra form view only to avoid open move lines in edit mode.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT45211

PR #1151, opened 2023-09-28
[14.0][FIX] account_payment_order: Remove readonly=1 to move lines to allow to remove records

victoralmau:
Related to https://github.com/OCA/bank-payment/pull/1145

Remove readonly=1 to move lines to allow to remove records

Undo https://github.com/OCA/bank-payment/pull/1145/commits/f9a1487f8950652fa7b7639f205580aa06978de9

Another approach is https://github.com/OCA/bank-payment/pull/1147/commits/1e5d146d07a03d41a947bd4ae100e3757ba56a63, but it only works from v16.

@Tecnativa TT45211

### DIFF
--- a/account_payment_order/wizard/account_payment_line_create_view.xml
+++ b/account_payment_order/wizard/account_payment_line_create_view.xml
@@ -49,8 +49,6 @@
                     <field
                         name="move_line_ids"
                         nolabel="1"
-                        readonly="1"
-                        force_save="1"
                         context="{'tree_view_ref': 'account_payment_order.view_move_line_tree'}"
                     >
                         <tree>


pedrobaeza (commented):
And what about the specific form view for not allowing to edit things?

victoralmau:
> And what about the specific form view for not allowing to edit things?

Another approach is https://github.com/OCA/bank-payment/pull/1147/commits/1e5d146d07a03d41a947bd4ae100e3757ba56a63, but it only works from v16.

pedrobaeza:
Yes, we want that here, but if not, you should revert the previous commit, instead of creating a new one.

giarve (approved):
Previous commit should be reverted as it breaks "Add line" functionality and the message displayed on the same form confuses the user.

pedrobaeza:
Let's do the same for v15

PR #1153, opened 2023-10-02
[15.0][FIX+IMP] account_payment_order: Define the value of the communication field correctly

victoralmau:
FWP 14.0: https://github.com/OCA/bank-payment/pull/1142

Changes done:
- [x] Define the value of the `communication` field correctly
- [x] Add `payment_ref` field to `account.payment` tree and form views

Example use case:
factura
order-pago-1
orden-pago-2
orden-pago-3

@Tecnativa TT45153

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -38,6 +38,54 @@ class AccountMove(models.Model):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _get_payment_order_communication_direct(self):
+        """Retrieve the communication string for this direct item."""
+        communication = self.payment_reference or self.ref or self.name or ""
+        if self.is_invoice():
+            if (self.reference_type or "none") != "none":
+                communication = self.ref
+            elif self.is_purchase_document():
+                communication = self.ref or self.payment_reference
+            else:
+                communication = self.payment_reference or self.name
+        return communication
+
+    def _get_payment_order_communication_full(self):
+        """Retrieve the full communication string for the payment order.
+        Reversal moves and partial payments references added.
+        Avoid having everything in the same method to avoid infinite recursion
+        with partial payments.
+        """
+        communication = self._get_payment_order_communication_direct()
+        references = []
+        # Build a recordset to gather moves from which references have already
+        # taken in order to avoid duplicates
+        reference_moves = self.env["account.move"].browse()
+        # If we have credit note(s) - reversal_move_id is a one2many
+        if self.reversal_move_id:
+            references.extend(
+                [
+                    move._get_payment_order_communication_direct()
+                    for move in self.reversal_move_id
+                ]
+            )
+            reference_moves |= self.reversal_move_id
+        # Retrieve partial payments - e.g.: manual credit notes
+        for (
+            _x,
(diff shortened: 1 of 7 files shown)


carlosdauden (approved):

sergio-teruel (approved):

account_payment_order/models/account_move_line.py
```diff
@@ -53,16 +53,7 @@ def _get_linked_move_communication(self):
         # Build a recordset to gather moves from which references have already
         # taken in order to avoid duplicates
         reference_moves = self.env["account.move"].browse()
-        # If we have credit note(s) - reversal_move_id is a one2many
-        if self.move_id.reversal_move_id:
-            references.extend(
```
pedrobaeza, line 58: `references` doesn't contain all the needed ones. You are not calling `_get_payment_order_communication()` for this case.

pedrobaeza (changes requested):

account_payment_order/models/account_move_line.py
victoralmau: This condition is included in `_get_payment_order_communication()` method of `accont.move` (https://github.com/OCA/bank-payment/pull/1142#discussion_r1333933409).

`_get_communication()` method of `account.move.line` calls `_get_payment_order_communication()` so it already has this information.

Is there anything else I am not taking into account?

pedrobaeza: But you are returning here other references directly hard-coded in https://github.com/OCA/bank-payment/pull/1153/files#diff-c44811bc7189570a631438d2dca538756d8ad24b26ced88bfd5df3ab44b74f60R67

Why this is not constant?

victoralmau: Ok, it's already changed to be consistent.

Do you think the current behavior is correct?

account_payment_order/models/account_move.py
```diff
@@ -38,6 +38,55 @@ def _compute_payment_order_ok(self):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _get_payment_order_communication(self):
+        """
+        Retrieve the communication string for the payment order
+        """
```
pedrobaeza:
```suggestion
    def _get_payment_order_communication_direct(self):
        """Retrieve the communication string for this direct item."""
```

account_payment_order/models/account_move.py
```diff
@@ -38,6 +38,55 @@ def _compute_payment_order_ok(self):
...
+                communication = self.ref
+            elif self.is_purchase_document():
+                communication = self.ref or self.payment_reference
+            else:
+                communication = self.payment_reference or self.name
+        return communication
+
+    def _get_full_payment_order_communication(self):
+        """
+        Retrieve the full communication string for the payment order.
```
pedrobaeza:
```suggestion
    def _get_payment_order_communication_full(self):
        """Retrieve the full communication string for the payment order.
```

pedrobaeza (commented):
That's better readable, indeed! Some minor comments:

- I have put a naming scheme with the same prefix, and docstring have the first line with the first comment line without line break.

pedrobaeza (approved):
/ocabot merge patch

PR #1160, opened 2023-10-16
[16.0][IMP] bank-payment : test performance improvement 

josep-tecnativa:
CC @Tecnativa TT45458


pedrobaeza:
Don't do this PR until the 15.0 is finished, as there are comments in the 15.0 one.

account_banking_mandate/tests/test_invoice_mandate.py
```diff
@@ -12,26 +13,17 @@
 
 class TestInvoiceMandate(TransactionCase):
     def test_post_invoice_01(self):
+        self.invoice._onchange_partner_id()
```
pedrobaeza, line 16: Why these changes?

pedrobaeza:
ping @josep-tecnativa

pedrobaeza:
Let's close it in favor of #1188

PR #1163, opened 2023-11-04
[IMP] grouped invoices: " - ".join in stead of "-".join

luc-demeyer:
Payment Reference (communication) in case of grouped invoices into a single payment: " - ".join in stead of "-".join since we see many vendors using "-" in their invoice numbering which makes the "-".join confusing.

image

### DIFF
--- a/account_banking_sepa_credit_transfer/tests/test_sct.py
+++ b/account_banking_sepa_credit_transfer/tests/test_sct.py
@@ -202,7 +202,7 @@ class TestSCT(SavepointCase):
             ),
             0,
         )
-        self.assertEqual(agrolait_bank_line.payment_reference, "F1341-F1342-A1301")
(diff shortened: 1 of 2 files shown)


luc-demeyer:
@victoralmau 
You missed out this small usability improvement when you superceeded https://github.com/OCA/bank-payment/pull/1071 (I should have made a seperate PR for this small change).

pedrobaeza:
Why not using `|` instead?

luc-demeyer:
@pedrobaeza 
| is not allowed, cf. https://github.com/OCA/bank-payment/blob/9bcdea9feada58b59c7841071dcce072a831e830/account_banking_pain_base/models/account_payment_order.py#L164

pedrobaeza (approved):
Ouch, that's very limited...

Let's fast-track it:

/ocabot merge patch

PR #1164, opened 2023-11-04
[FIX]payment order - fix compute_partner_bank_id

luc-demeyer:
The bank account number of then invoice is not selected when a supplier has multiple bank account numbers and there is no payment mode set on the vendor bill.
This PR fixes this issue.

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -31,7 +31,7 @@ class AccountMoveLine(models.Model):
             if (
                 ml.move_id.move_type in ("in_invoice", "in_refund")
                 and not ml.reconciled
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

ghost (approved):

pedrobaeza (changes requested):
Wait, I'm now thinking that this is going to fill the bank account and fix it when it's imported in payment orders, which disables the feature of filling it on that case. Why do you want such behavior?

alexis-via (approved):

pedrobaeza:
@alexis-via I have put a requested changes review because I have now doubts. Can you please check them?

PR #1167, opened 2023-11-05
[16.0] account_payment_purchase: replace api.onchange by computed field on purchase.order

alexis-via:
replace api.onchange by computed field on purchase.order
Add company_id in the domain of the fields of purchase.order
Add check_company=True on the 2 fields of purchase.order

### DIFF
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -10,16 +10,26 @@ class PurchaseOrder(models.Model):
 
     supplier_partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
+        compute="_compute_payment_mode",
+        readonly=False,
+        store=True,
+        precompute=True,
         string="Supplier Bank Account",
-        domain="[('partner_id', '=', partner_id)]",
+        domain="[('partner_id', '=', partner_id), ('company_id', 'in', [False, company_id])]",
+        check_company=True,
         help="Select the bank account of your supplier on which your company "
         "should send the payment. This field is copied from the partner "
         "and will be copied to the supplier invoice.",
     )
     payment_mode_id = fields.Many2one(
         comodel_name="account.payment.mode",
+        compute="_compute_payment_mode",
+        readonly=False,
+        store=True,
+        precompute=True,
         string="Payment Mode",
-        domain="[('payment_type', '=', 'outbound')]",
+        domain="[('payment_type', '=', 'outbound'), ('company_id', '=', company_id)]",
+        check_company=True,
     )
 
     @api.model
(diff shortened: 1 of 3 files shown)


account_payment_purchase/models/purchase_order.py
```diff
@@ -7,19 +7,31 @@
...
     supplier_partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
+        compute="_compute_payment_mode",
+        readonly=False,
+        store=True,
+        precompute=True,
         string="Supplier Bank Account",
-        domain="[('partner_id', '=', partner_id)]",
+        domain="[('partner_id', '=', partner_id), "
+        "'|', ('company_id', '=', company_id), ('company_id', '=', False)]",
```
pedrobaeza: You can simplify the domain this way:

`[("company_id", "in", [False, company_id])]`

account_payment_purchase/models/purchase_order.py
```diff
@@ -7,19 +7,31 @@
 
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
+    _check_company_auto = True  # missing in the native purchase module !!!
```
pedrobaeza: I think this side change shouldn't be done here.

account_payment_purchase/views/purchase_order_view.xml
```diff
@@ -9,11 +9,7 @@
         <field name="inherit_id" ref="purchase.purchase_order_form" />
         <field name="arch" type="xml">
             <field name="payment_term_id" position="after">
-                <field
-                    name="payment_mode_id"
-                    domain="[('payment_type', '=', 'outbound')]"
-                    widget="selection"
-                />
+                <field name="payment_mode_id" widget="selection" />
```
pedrobaeza: AFAIK, this widget is conflicting in certain cases in v16. Better to use m2o

rvalyi: out of curiosity do you have some references or more details about the problems?

pedrobaeza: #1125

account_payment_purchase/models/purchase_order.py
alexis-via: done

account_payment_purchase/models/purchase_order.py
alexis-via: Of course ! It should be done in the purchase module !!!

account_payment_purchase/views/purchase_order_view.xml
alexis-via: done

account_payment_purchase/models/purchase_order.py
pedrobaeza: Yes, but not being in core, propose it there, but do another module for this, not put this on a traversal module that has nothing to do with it.

luc-demeyer (approved):

luc-demeyer:
Code review & testing ok.

I am wondering why precompute=True is set on the computed fields.
I tested by changing to precompute=False and the PR works in both cases.
@pedrobaeza 
do you think we need precompute=True in this case ?
When reading the doc in the ORM one would think it's required but it seems not to be the case.

pedrobaeza:
It's because some fields are already precompute, and the dependency chain should be completely that way for avoiding double computing.

account_payment_purchase/models/purchase_order.py
alexis-via: done

alexis-via:
I also converted (0, 0, {}) to Command.create([})
Ready to merge

pedrobaeza (approved):

PR #1172, opened 2023-11-06
[16.0][IMP] account_payment_order: add payment lines smart button on invoice

luc-demeyer:


alexis-via (approved):

account_payment_order/models/account_move.py
```diff
@@ -207,3 +214,23 @@ def create_account_payment_line(self):
                 }
             )
         return action
+
+    def action_payment_lines(self):
+        self.ensure_one()
```
ghost, line 221: Is 'self.ensure_one()' mandatory ? From my point of view, a future/possible improvement or an open door at the level of a particular project would be an action in the list view for a selection of invoices.

ghost (approved):
LGTM (Just a little open question on the action method)

account_payment_order/models/account_move.py
```diff
@@ -38,6 +39,12 @@ def _compute_payment_order_ok(self):
                 payment_mode = move.payment_mode_id
             move.payment_order_ok = payment_mode.payment_order_ok
 
+    def _compute_payment_line_count(self):
+        for move in self:
+            move.payment_line_count = self.env["account.payment.line"].search_count(
```
pedrobaeza: Use `read_group` instead for better performance.

account_payment_order/models/account_move.py
```diff
@@ -207,3 +214,23 @@ def create_account_payment_line(self):
...
+    def action_payment_lines(self):
+        self.ensure_one()
+        pl_ids = (
+            self.env["account.payment.line"]
+            ._search([("move_line_id", "in", self.line_ids.ids)])
+            ._result
+        )
+        return {
+            "type": "ir.actions.act_window",
+            "name": _("Payment Lines"),
```
pedrobaeza: Instead of creating an action from scratch, requiring another translation and full definition, just read the existing action (using `_for_xml_id`).

account_payment_order/views/account_invoice_view.xml
```diff
@@ -39,6 +39,21 @@
...
             </button>
+            <div name="button_box" position="inside">
+                <button
+                    name="action_payment_lines"
+                    type="object"
+                    attrs="{'invisible': [('payment_line_count', '=', 0)]}"
+                    icon="fa-bars"
+                    help="Payment Order Lines"
+                >
+                <field
```
pedrobaeza:
```suggestion
                    <field
```

account_payment_order/views/account_payment_line.xml
```diff
@@ -1,12 +1,14 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
+
```
pedrobaeza: Don't add useless empty lines for nothing. In fact, auto-formatters tend to remove them (not for now on OCA), but in this case that they don't exist before, it's not correct to add them.

pedrobaeza (changes requested):

pedrobaeza:
And finally, please follow commit message guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message, putting the module after the tag.

account_payment_order/models/account_move.py
luc-demeyer: this is an interesting one.
I didn't imagine that Odoo uses def search in stead of def _search in search_count.
Any idea why they do this ?

luc-demeyer:
@pedrobaeza 
imho we  are ok now.

account_payment_order/models/account_move.py
pedrobaeza: Well, I was talking about doing something similar to:

https://github.com/OCA/bank-payment/blob/2f3def4c0306dc4dd1c0808383451dc344e4d94e/account_payment_order_grouped_output/models/account_payment_order.py#L21-L32

account_payment_order/models/account_move.py
```diff
@@ -207,3 +216,25 @@ def create_account_payment_line(self):
                 }
             )
         return action
+
+    def action_payment_lines(self):
+        self.ensure_one()
+        pl_ids = (
+            self.env["account.payment.line"]
+            ._search([("move_line_id", "in", self.line_ids.ids)])
```
pedrobaeza: You can put directly this domain in the action.

account_payment_order/models/account_move.py
luc-demeyer: In Odoo 17.0 the search_count performance issue has been fixed, cf.  https://github.com/odoo/odoo/blob/17.0/odoo/models.py#L1582

account_payment_order/models/account_move.py
luc-demeyer: done.

pedrobaeza (approved):
Thanks for the improvement:

/ocabot merge minor

PR #1174, opened 2023-11-07
[16.0][BIG] Large refactoring/improvement/cleanup of OCA/bank-payment

alexis-via:
NEW FEATURES
- activate 'in_payment' payment_state on invoices... at last !
- payment order numbering: add ability to have a sequence specific to a particular payment mode (optional)
- add show account number policy "first_last" to show N first and N last characters of bank account number
- add support for regulatory reporting
- replace unstructured address by structured address in SEPA XML file (mandatory starting 11/2025)
- add support for pain.008.01.08 (SDD) and pain.001.001.09 (SCT), which are now the recommended versions of the EPC

SMALL IMPROVEMENTS
- account.payment.method: rename field "payment_order_only" to "payment_order_ok" for naming consistency with the field of account.payment.mode. Auto-set "payment_order_ok" of payment mode from the equivalent field of payment method. Migration scripts provided for account.payment.method and the data of SEPA credit transfer and SEPA direct debit.
- payment order attachment: add M2O payment_file_id that points to the payment file. No more need for a simplified attachment form view.
- keep the spaces in the scrambled bank account number, so that the user easily understands that it is a "scrambled" bank account number
- add field acc_number_scrambled on res.partner.bank for easy and direct use of scrambled account number
- improve the layout of the sepa section of the accounting config page
- hide field mandate_id on form + tree views of payment lines if payment mode is mandate_required=False
- automatically open download pop-up when clicking on "Generate payment file".
- set no_debit_before_maturity to True by default on payment modes
- set default_date_prefered to "due" by default on inbound payment modes with payment_order_ok = True
- search on partner_id from account.payment.order
- add tracking on payment_mode_id on sale.order

SIMPLIFICATION / CODE CLEANUP
- big cleanup of XML code generation: use lxml.objectify instead of lxml.etree, add support for currencies with decimal places != 2, stop using safe_eval()
- remove field partner_bank_id on account.move.line
- move code for scrambled account number from qweb report to res.partner.bank class
- remove boolean field show_bank_account_from_journal from account.payment.mode: we now consider that this option is always true i.e. when account_payment_partner is installed, Odoo takes the bank accounts to display on the customer invoice from the payment mode (if the payment mode is empty, it displays all the bank accounts of the company). As a consequence, hide partner_bank_id on customer invoice form view.
- remove fields inbound_payment_order_only and outbound_payment_order_only from account.journal : these 2 fields were not used!
- add check_company=True where it was missing and remove code of constraint that checks company consistency
- re-organise form view of payment order to equilibrate between left column and right column
- remove support for pain.001.001.02/04/05 (SCT) and pain.008.01.03/04 (SDD) which have never been selected by the EPC

TECHNICAL:
- replace @api.onchange by computed readonly=False store=True fields
- remove (or reduce) the use of demo data in tests
- speed-up tests by using tracking_disable=True
- Replace (6, 0, []) and (0, 0, {}) by Command.set() and Command.create()
- add sql unicity constraint on payment order number per company

FIXES:
- bug #1336 
- bug #1337 disallow the creation of a SEPA mandate linked to a non-IBAN bank account
- bug #1209 
- bug #1477


alexis-via:
It's the module account_payment_order_grouped_output which make the tests for payment_status in_payment vs paid fail. I have never used account_payment_order_grouped_output  ; according to README, it has been developed to mitigate the drawbacks of the switch to account.payment !!!

alexis-via:
For some strange reasons that I don't understand, when you upgrade to this branch, Odoo fails because a view has been deleted in the source code. You can solve the issue by executing this SQL request:

```
delete from ir_ui_view where id=(select res_id from ir_model_data where module='account_payment_order' and model='ir.ui.view' and name='view_move_line_form');
```

alexis-via:
Tests are not green.

alexis-via:
**Payment order, in_payment and bank statement reconciliation**

In this PR, I activate the "in_payment" payment status on invoice. For me, it's one of the major advantages of using the object account.payment, that was implemeted by @pedrobaeza in this big PR https://github.com/OCA/bank-payment/pull/979
Surprisingly, Pedro didn't activate the in_payment status in his PR. In fact, activating such a feature is very easy, you just have to inherit the method _get_invoice_in_payment_state() in account.move and return "in_payment" instead of "paid":
https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L4345
Only 1 line of code!

One of the drawbacks in the switch to account.payment is that, when you validate a payment order with 100 different partners, you get 100 different entries in the Outstanding receipts account. So, when you have the bank statement line with the total amount of the payment order, you have to select 100 counter-part entries in the Outstanding receipts account !!!

The module account_payment_order_grouped_output is designed to solve this problem:
https://github.com/OCA/bank-payment/tree/16.0/account_payment_order_grouped_output
It generates an additional account.move:
- 100 lines in the Outstanding receipts account with reconciliation
- 1 counter part line for the total amount... in the Outstanding receipts account
Advantage: in the bank statement reconcile interface, you just have to select 1 counterpart line.
Drawbacks: it breaks the "in_payment" payment status, because the counter-part line of the account.move generated by the account.payment is reconciled when the payment order is validated and not when the bank statement line is processed ! Looking at the code about this in the "account" module, it's not easy to modify that.

I was told that, in the Enterprise module, they have a special tab in the bank statement reconcile interface for payments by lot: so, when you process the bank statement lines with the total amount of the payment order, Odoo proposes the payment orders not yet reconciled with the bank statement lines ; when you select the appropriate payment order, it will automatically set the 100 counterpart lines to reconcile the bank statement line.
It seems that the OCA module account_reconcile_payment_order
https://github.com/OCA/account-reconcile/tree/15.0/account_reconcile_payment_order
is designed to propose a similar feature in the community bank statement reconcile interface. But this module is not available in Odoo v16 with the new account_reconcile_oca module.

So, how should be move forward on this issue ? @pedrobaeza @etobella 

My opinion:
1) account_payment_order_grouped_output is not a good option: we would have to modify the native behavior of the computation of the in_payment state, which is not easy and not desirable. In addition, it adds an extra account.move.
2) a solution similar to Odoo Enterprise seemed a bit strange to me at first... but, after some discussions with other community member, I eventually think it may be a good solution. If account_reconcile_payment_order is designed for that, we could port it to v16 and make it depend on account_reconcile_oca. It's probably not easy, but it's really necessary if we want to have both the "in_payment" payment_status on invoices and an easy reconcile process in the bank statement, even for payment orders with a lot of different partners.

pedrobaeza:
Yes, `account_reconcile_payment_order` adapted to the new reconcilation system sees the way to go. This module can add a new tab "Payment/Debit orders" that allows to select the non-reconciled ones, and clicking on one of them, it adds all the AR/AP generated lines (but allowing to remove/re-add some of them if needed).

miquelalzanillas (approved):
What a great PR! 

Tested functionallity in a fresh new db and seems is working.

etobella:
@alexis-via Can you rebase?

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -657,7 +654,7 @@ def generate_remittance_info_block(self, parent_node, line, gen_args):
                 creditor_ref_info_type_issuer = etree.SubElement(
                     creditor_ref_info_type, "Issr"
                 )
-                creditor_ref_info_type_issuer.text = communication_type
+                creditor_ref_info_type_issuer.text = "ISO"
```
sbidoul: @alexis-via please keep communication_type. We need it for belgium where (IIRC) it must become bba.
https://github.com/OCA/l10n-belgium/tree/10.0/l10n_be_iso20022_pain

cc/ @luc-demeyer

luc-demeyer: @sbidoul 
I fully agree.
Small remark on l10n_be_iso20022_pain: should be 'BBA' (uppercase), cf. my comment on 16.0 PR.

luc-demeyer:
@alexis-via 
I am still waiting on merge of PR's we made during OCA days and rebase of this PR before I start in-depth code review and testing  of this PR.

Mat-moran:
Hi @luc-demeyer , which PR are you waiting for? Can we help to push this? Is blocking migration to v17

pedrobaeza:
Please rebase and solve conflicts.

alexis-via:
I'll continue to work on this next month (but not in the coming days)

account_banking_pain_base/models/account_pain_regulatory_reporting.py
```diff
@@ -0,0 +1,49 @@
...
+    ]
+
+    @api.depends("code", "name")
+    def name_get(self):
+        res = []
+        for rec in self:
+            res.append((rec.id, "[%s] %s" % (rec.code, rec.name)))
+        return res
+
+    def _name_search(
```
pedrobaeza: Instead of this, you can just put `_rec_names_search = ["code", "name"]`.

account_banking_pain_base/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_pain_regulatory_reporting_read,Read access on account.pain.regulatory.reporting,model_account_pain_regulatory_reporting,base.group_user,1,0,0,0
+access_account_pain_regulatory_reporting_full,Full access on account.pain.regulatory.reporting,model_account_pain_regulatory_reporting,account.group_account_manager,1,1,1,1
```
pedrobaeza: Shouldn't be the full right access the "Payment" group?

account_banking_pain_base/views/account_pain_regulatory_reporting.xml
```diff
@@ -0,0 +1,80 @@
...
+                    bg_color="bg-danger"
+                    attrs="{'invisible': [('active', '=', True)]}"
+                />
+            <group name="main">
+                <field name="code" />
+                <field name="name" />
+                <field name="country_id" />
+                <field name="active" invisible="1" />
+        </group>
+</form>
```
pedrobaeza, line 25: Incorrect indentation

account_banking_sepa_credit_transfer/migrations/16.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,11 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    sct_method = env.ref("account_banking_sepa_credit_transfer.sepa_credit_transfer")
+    sct_method.write({"payment_order_ok": True})
```
pedrobaeza: Don't force this. Just rename the column of the field on the source one.

account_banking_sepa_credit_transfer/models/account_payment_order.py
```diff
@@ -191,10 +191,11 @@ def generate_payment_file(self):  # noqa: C901
                     gen_args,
                     line,
                 )
-                line_purpose = line.payment_line_ids[:1].purpose
-                if line_purpose:
-                    purpose = etree.SubElement(credit_transfer_transaction_info, "Purp")
-                    etree.SubElement(purpose, "Cd").text = line_purpose
+                payment_line = line.payment_line_ids[0]
```
pedrobaeza, line 139: Why changing this? Are you sure there's always going to be a payment_line_id?

account_banking_sepa_credit_transfer/tests/test_sct.py
```diff
@@ -353,25 +384,24 @@ def create_invoice(
...
+        line_data = {
+            "name": "Great service",
+            "account_id": cls.account_expense.id,
+            "price_unit": price_unit,
+            "quantity": 1,
+        }
         data = {
             "partner_id": partner_id,
-            "reference_type": "none",
+            "reference_type": "free",
```
pedrobaeza, line 391: Why do you need this change?

account_banking_sepa_direct_debit/migrations/16.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,11 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    sct_method = env.ref("account_banking_sepa_direct_debit.sepa_direct_debit")
+    sct_method.write({"payment_order_ok": True})
```
pedrobaeza: The same about not forcing

pedrobaeza (commented):
Half of the files reviewed.

rousseldenis:
@alexis-via Any news on this ?

flotho:
Hi @alexis-via any chance to trigger the CI ? regards

alexis-via:
I'm back on this PR... at last ! :)

Any objection to removal of pain.001.001.02 ?
I'm re-writing the code of XML generation using lxml.objectify and removing pain.001.001.02 would simplify the code. As EPC asks for pain.001.001.03, I don't think anyone is using pain.001.001.02.
I also plan to remove safe_eval() in account_banking_pain_base.
FYI, I already added support for currencies with decimal places != 2 (not yet commited).

pedrobaeza:
Please do each of the changes grouped by logic sets in smaller PRs. If not, this is hardly going to be accepted due to the impossibility of reviewing everything and the hell of later blame operations they mean.

account_banking_pain_base/models/account_pain_regulatory_reporting.py
alexis-via: Done. Thanks for your remark!

alexis-via:
I implemented structured addresses, as per EPC guidelines (mandatory starting 11/2025) https://www.europeanpaymentscouncil.eu/document-library/guidance-documents/guidance-use-structured-addresses-2025-sepa-payment-schemes

alexis-via:
About PAIN versions: there is a new version of pain.001.001.xx almost every year, cf https://www.iso20022.org/catalogue-messages/iso-20022-messages-archive?page=5  The latest version is pain.001.001.12 (pain.008.001.11 for SDD) !
So there is no point in trying to support every pain version, we should focus on the versions "selected" by the EPC. Until 2023, the versions in the implementation guidelines of the EPC were pain.001.001.03 for SCT and pain.008.01.02 for SDD. But, if you read the up-to-date version of the implementation guidelines of the EPC, you'll find they are now based on pain.001.001.09 for SCT and pain.008.01.08 for SDD.

So I added support for pain.001.001.09 and pain.008.01.08 and I removed the versions that have never been selected by EPC (pain.001.001.02/04/05 and pain.008.01.03/04)

pedrobaeza (changes requested):
As said, a PR modifying 90 files and doing a lot of stuff is something that is not reviewable. Please split it properly into several PRs by the corresponding logical blocks (PAIN version, Address support, etc...)

alexis-via:
Yes, it's a big PR, because this project really needs a big cleanup/rewrite and there are many new things published by the European Payment Council (I wasn't aware of all that until this week !!!)

I think that this PR will never be merged in v16, but it's a "technology preview" of what we should have in v18. I deploy this PR on all my Odoo v16 deployments, so it's production stable. I plan to port OCA/bank-payment to v18 as soon as v18 is released, taking inspiration from the improvements of this PR. So, if some people disagree about the changes introduced by this PR, it's time to discuss now, before it ends up in v18.

We followed the same path with my big simplification/update/cleanup intrastat PR for v14 https://github.com/OCA/intrastat-extrastat/pull/198 It was never merged in v14, but it ended up in v16 during the v15 to v16 migration. Advantage : it was already used in production by my v14 customers, so everyone could be sure that it was something serious, and not some crazy code to make the revolution. And we had time to discuss my proposed changes on the PR before the v16 migration work started.

pedrobaeza:
The same problem will exist in v18. You mix everything together without reason, just because "you are now working on it after 1 year of no caring". Merging it in v18 will be the same problem: there's no way to follow the reasons for most of the changes, ending up with a huge effort to understand and to fix things that have been changed by you because they are not of your taste.

Sorry, but in something so critical like this set of modules (Intrastat are just a niche), we can't allow that. You have been the exception for so many time having big PRs merged. Please do like the rest of the contributors, splitting things properly, and gradually merging things.

alexis-via:
I haven't merged big PRs on OCA/bank-payment for a very long time. The last big PR I merged here was the rewrite I made in v9 during the Sorrento code sprint, and it was very necessary at that time.

The last big PR that was merged on OCA/bank-payment was your PR that introduced the use of account.payment (but didn't add support for the "in_payment" status on invoice, I was very surprised about that). It was a big revolution in the code. There is still one big drawback following the use of account.payment : the reconciliation on bank statements. Before, we could configure the payment mode so that we only had 1 journal item in the outbound payment account for many payments to many different suppliers. Now, it's not possible any more: if you have a payment order for 10 different suppliers executed on the same day and there is only 1 line of the bank statement for the total amount, the accountant will have to manually select the 10 journal items in the outbound payment account. I have many accountant among my customers that complain about this and tell me "it was better before" ! Using the OCA module "account_payment_order_grouped_output" is not a solution because it breaks the "in_payment" status on invoices. In the end, your PR for account.payment was merged although there was nothing ready for the bank statement reconciliation. We really need to work on it now.

OK, this PR is difficult to review because there are many changes in the code. I tried to separate the changes in different commits with accurate commit messages, but it's not always possible, in particular when I re-wrote the code of the SEPA XML generation. The code that generated the SEPA XML was so horrible and verbose that it was really needed. I wrote that code in v5 when I started learning Python and Odoo and it seems that the code survived until today ! Moving to lxml.objectify was a "technology test" for me. Now that it's implemented in this PR, we can discuss if it's a good option for v18 and if we should move to qweb : at least we have a working implementation, so can be judge the code. Let's talk about these important choices!

I update the first comment on this PR to accurately list all the changes, so that the features and changes can be discussed by functional experts that don't look at the code.

rvalyi:
@alexis-via in OCA/l10n-brazil we use/extend account_payment_partner, account_payment_mode and account_payment_order for the boleto and CNAB, our SEPA Brazilian equivalent, so I propose my help for splitting this huge PR.

The strategy could be: first do small changes in these central modules and adapt the satellites modules (PAIN and SEPA) to keep test passing but without trying to refactor them too much initially. And finally do the the final refactor in the satellite modules. We can stack the sub-PRs one atop of the other using https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

But the thing is: for each sub-PR, we will need to get tests passing and avoid doing crazy changes in the critical modules during the intermediary steps.

I can help with the git rebase wizardry and some work, but I think we should discuss first what can be postponed to the latest steps in the satellite modules such as the PAIN and SEPA modules.

@pedrobaeza is sadly right about the same problem will occur with v18. Because nowadays we tend to forward and back port fixes between branches and if we keep the crazy refactor for v18 altogether, critical features from v16/v17 will be broken and these branches will loose synergy with v18. So my point is: for some big changes (the SEPA XML generation?) that might be inevitable and so be it. But eventually some smaller changes in the central modules can be injected in v16 in smaller bits.

I propose we talk about this. Sending you an email.

cc @renatonlima @mbcosta @marcelsavegnago @antoniospneto

alexis-via:
@rvalyi I don't think it's worth the effort to try to merge this PR in v16 and try to forward port to v17. We are just one month away from v18, so we should now focus on v18. It's too late for v16.

What's important here is to discuss the features and changes I propose : are they good for the OCA/bank-payment project or not ? Is it good to use lxml.objectify ? It is good to drop safe_eval in _prepare_field method ? It is good to drop the pain versions that have never been selected by EPC or are they used in real life by real users that want to keep them ? It the datamodel for regulatory reporting OK ? It is good to drop the fields "inbound_payment_order_only" and "outbound_payment_order_only" from account.journal... they don't seem to be used anywhere !!!

I can understand that having so many new ideas and changes can be a bit scary. But having the current big technical debt with dead code and awful XML generation code for such a long time scares me even more. It seems we didn't follow the EPC news these last years and we didn't implement the changes that we have to implement. Reading EPC documentation is not fun and I guess only very few devs do it. So I think the migration to v18 will be the right opportunity to implement the changes of this PR that users want/need (and drop the changes that users disapprove). If the community wants these changes/features, my v18 migration PRs will be used and merged by the community. I read that @pedrobaeza doesn't understand the reasons for most of the changes. I'm ready to give more details and more explanations on the changes that are not clear and explain the motivations. I also read on this PR that @miquelalzanillas likes the proposed changes. I would love the have the point of view of other people!

pedrobaeza:
Just open one PR for each of the ideas you have, to be discussed separately, and avoiding small changes of your taste that are introduced unadvertised. I'm just asking that.

rousseldenis:
The changes you introduced here are great!

But, IMHO, @pedrobaeza arguments sound reasonable if you want to attract other people to review as the current PR is too huge and too much changes at once (maybe some want to review one part as the one they know well and not another). Of course some changes have no sense without others, but you'll get better progress by splitting this than waiting a full approval.

My two cents

alexis-via:
OK, but what do you think of the features and changes I propose @rousseldenis ? Do you agree with them ?

rousseldenis:
> OK, but what do you think of the features and changes I propose @rousseldenis ? Do you agree with them ?

I read quickly, I need to take time to check feature by feature. Not sure I can do it today but I'll check next week if you want full review

account_banking_pain_base/models/account_payment_order.py
alexis-via: Thanks for pointing this out. Let's talk about this during the next code sprint, because the datamodel for structured communication type is really messy and unclean. I'm confident that we'll agree on a nice and clear datamodel on this, that will make the l10n_be_iso20022_pain simpler and make the XML generation code simpler.

account_banking_pain_base/security/ir.model.access.csv
alexis-via: done

account_banking_pain_base/views/account_pain_regulatory_reporting.xml
alexis-via: I let prettier do the formatting work and pre-commit doesn't give any error. Strange !

account_banking_sepa_credit_transfer/migrations/16.0.1.2.0/post-migration.py
alexis-via: But the payment method for SEPA credit transfer should always have payment_order_ok = true, no ? Otherwise, what's the point of installing account_banking_sepa_credit_transfer ??

account_banking_sepa_credit_transfer/tests/test_sct.py
alexis-via: I changed the value in the module account_payment_order to have the same value ("free") on account.move and account.payment.line for unstructured communication. Avoids a headache for maintainers. Migration scripts are provided in account_payment_order, cf https://github.com/akretion/banking/blob/16-payment_order-improve_and_cleanup/account_payment_order/migrations/16.0.2.0.0/pre-migration.py#L14

account_banking_sepa_credit_transfer/models/account_payment_order.py
alexis-via: It's just the same code that was moved from account_banking_sepa_credit_transfer to account_banking_pain_base, to mutualize the code with account_banking_sepa_direct_debit. On a payment order, an account.payment is always linked to N account.payment.line, and we read the purpose on the account.payment.line because that's where the information is. I didn't change the logic.

alexis-via:
I fixed the layout of the config page

BEFORE : 

before


AFTER : 

page_config_sepa

pedrobaeza:
Please do such fix in a separate PR, and it will be immediately merged.

alexis-via:
My re-write of the code that generate the address block fixes the bug reported here:
https://github.com/OCA/bank-payment/issues/1209
I saw the bug when I was working on that code... I just wasn't aware someone had reported it !

alexis-via:
While reading the code, I found a new bug on direct debit, that I later confirmed through tests. I opened a bug report https://github.com/OCA/bank-payment/issues/1336 and I fixed it in my last commit.

flotho:
Hi @alexis-via 
Thanks for this PR. I still have an issue when generating the sepa file . All the account are set to the wrong value.
I've tested this with the latest changes of your PRs
I made a proposal here https://github.com/OCA/bank-payment/pull/1212

Regards

MiquelRForgeFlow:
> It's too late for v16.

We still have few clients in v12. Never it's too late.

alexis-via:
Yesterday, I added a small but nice improvement : automatically open download pop-up when clicking on "Generate payment file".

alexis-via:
I made 2 other minors changes to have "good practise" by default when configuring payment modes:
- set no_debit_before_maturity to True by default on payment modes
- set default_date_prefered to "due" by default on inbound payment modes with payment_order_ok = True

flotho:
Hi @alexis-via ,

Any chance to cherry pick this commit https://github.com/flotho/bank-payment/commit/da2028cafdae0604d9371cb726046fd676f49628 to FIX https://github.com/OCA/bank-payment/issues/1193
and FIX https://github.com/OCA/bank-payment/issues/1131

Moreover, I faced this issue after an upgrade of the pain base module while trying to generate a file : 

image

Any tips to find which state field is concerned ?

Regards

flotho:
> Hi @alexis-via ,
> .../... 
> Moreover, I faced this issue after an upgrade of the pain base module while trying to generate a file :
> 
> image
> 
> Any tips to find which state field is concerned ?
> 
> Regards

Found, state_id was missing on the partner

flotho:
Hi @alexis-via , any chance to have a rebase here ?

dnplkndll:
checking the status on this refactor. is the plan still to release for 16.0?

pedrobaeza:
No on my side. This contains too many changes and not split. See my comments at https://github.com/OCA/bank-payment/pull/1174#pullrequestreview-2255314509, https://github.com/OCA/bank-payment/pull/1174#issuecomment-2306393466, https://github.com/OCA/bank-payment/pull/1174#issuecomment-2307146912, etc

There's no real intention of this to be merged by the contributor, unless we all accept all the changes he has though on his side and a lot of breaking things. It's a de facto fork.

dnplkndll:
I had the chance to read over the comments. really we just need to know where to start with an ACH ( USA ) portal feature to allow the payments of statements ( sets of invoices due for the month. ) It would like nice to add rules portal users could apply like the pay on due date or on the 20th of month via bank or set a payment on the due date from the reminder email. looks like payment.acquirer connection and the portal views. 

We were hoping to build this on 16.0 then migrate to 18 at the same time.

pedrobaeza:
Interesting feature. I think you can mount it on top of the current merged code, as even if this is accepted, it will be on 18, not on this versions that is already almost 3 years old.

PR #1185, opened 2023-11-26
[17.0][MIG] account_payment_purchase: Migration to 17.0 

ramiadavid:

PR #1187, opened 2023-11-26
[17.0][MIG] account_payment_purchase_stock: Migration to 17.0

ramiadavid:
Depends on:

- [x] #1185


account_payment_mode/tests/test_account_payment_mode.py
```diff
@@ -21,7 +21,9 @@ def setUpClass(cls):
         cls.company = cls.env.ref("base.main_company")
 
         # Company 2
-        cls.company_2 = cls.env["res.company"].create({"name": "Company 2"})
+        cls.company_2 = cls.env["res.company"].create(
+            {"name": "Company 2", "po_lead": 0}
```
etobella: I am not sure about this. What will happen if we test this module without installint account_payment_purchase_stock?

ramiadavid: @etobella I'm sorry, but I don't understand the question.

etobella: You modified a test of another module for some reason and it has no sense IMO.

You cannot expect that purchase_stock is installed for testing `account_payment_mode`.

You should remove this change

ramiadavid: Deleted, I put it because the tests gave errors, but I don't know why now it no longer gives errors

etobella (approved):
No major changes

pedrobaeza (approved):
/ocabot merge nobump

PR #1191, opened 2023-11-30
[ADD][14.0]account_payment_order_currency

syera94:
@florian-dacosta 
@alexis-via 

It s for check .
It could be nice if we can talk about it tmw


account_payment_order_currency/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Syera BONNEAUX <syera.bonneaux@gakretion.com>
```
florian-dacosta: Small mistake in email address same for the below line:

account_payment_order_currency/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds currency choice on model account.payment.mode and account.payment.line.create
```
florian-dacosta: Something more functional friendly like 
This module adds currency choice during the payment order filtering step. (this is not perfect at all, but you should not put some model name in this functional description IMO)

account_payment_order_currency/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,35 @@
+# © 2009 EduSense BV (<http://www.edusense.nl>)
+# © 2011-2013 Therp BV (<https://therp.nl>)
+# © 2014-2015 ACSONE SA/NV (<https://acsone.eu>)
```
florian-dacosta, line 3: You can remove all these lines, they did not participate to this module

account_payment_order_currency/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,35 @@
...
+class AccountPaymentLineCreate(models.TransientModel):
+    _inherit = "account.payment.line.create"
+
+    currency_ids = fields.Many2many(
+        comodel_name="res.currency", string="Currency Filter"
+    )
+
+    @api.model
+    def default_get(self, field_list):
+        res = super(AccountPaymentLineCreate, self).default_get(field_list)
```
florian-dacosta: It is really minor and not blocking at all, but you can simplify removing the default super arguments :  `super().default_get(field_list)`

account_payment_order_currency/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,35 @@
...
+        res.update(
+            {
+                "currency_ids": mode.default_currency_ids,
+            }
+        )
+        return res
+
+    def _prepare_move_line_domain(self):
+
+        res = super(AccountPaymentLineCreate, self)._prepare_move_line_domain()
```
florian-dacosta: same here about super default arguments

account_payment_order_currency/wizard/account_payment_line_create_view.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  © 2013-2016 Akretion (https://www.akretion.com)
+  @author: Alexis de Lattre <alexis.delattre@akretion.com>
+  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
florian-dacosta: You can remove these commented lines I guess

account_payment_order_currency/readme/CONTRIBUTORS.rst
syera94: done

account_payment_order_currency/readme/DESCRIPTION.rst
syera94: done

account_payment_order_currency/wizard/account_payment_line_create.py
syera94: done

account_payment_order_currency/wizard/account_payment_line_create_view.xml
syera94: done

florian-dacosta (commented):
One more issue, you need to recompute the domain of the `move_line_ids` field in case the currency change. 
In the module, it is done here : https://github.com/OCA/bank-payment/blob/14.0/account_payment_order/wizard/account_payment_line_create.py#L170

Since I believe we can't just add a field in the `api.onchange` decorator of a method, I would create a new method, with an api.onchange of your currency field, and this method could call the `move_line_filters_change` method of `account_payment_order` module.

PR #1192, opened 2023-11-30
[ADD][14.0]account_payment_order_validation_move_line

syera94:
@florian-dacosta 
@alexis-via 
for check.
Same than account_payment_order_currency, it should be great if we can talk about it tmw.


account_payment_order_validation_move_line/models/account_payment_mode.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class AccountPaymentMode(models.Model):
+
+    _inherit = "account.payment.mode"
+
+    default_selected_for_payment_id = fields.Boolean(string="To Pay Filter")
```
florian-dacosta: rename field to `default_selected_for_payment`, by convention in Odoo, the _id prefix is kind of reserved for many2one

account_payment_order_validation_move_line/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Syera BONNEAUX <syera.bonneaux@gakretion.com>
```
florian-dacosta: Fix the email address please (same for the line below)

account_payment_order_validation_move_line/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds the possibility to choose only validated invoice for the account.payment.line.create and it can be set from theaccount.payment.mode model and account.payment.line.create
```
florian-dacosta: The module adds the possibility to filter only invoices that have been marked as "To pay" on the payment order.

Or something like that (do not speak about technical models here as any functional person won't be able to understand)

account_payment_order_validation_move_line/views/account_payment_mode.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="account_payment_mode_form" model="ir.ui.view">
+        <field name="name">account_payment_order.account.payment.mode.form</field>
+        <field name="model">account.payment.mode</field>
+        <field name="inherit_id" ref="account_payment_mode.account_payment_mode_form" />
+        <field name="arch" type="xml">
+            <field name="default_date_type" position="after">
+                <field name="default_selected_for_payment_id" />
```
florian-dacosta: + rename field here too

account_payment_order_validation_move_line/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,24 @@
+# © 2015-2016 Akretion (<https://www.akretion.com>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class AccountPaymentLineCreate(models.TransientModel):
+    _inherit = "account.payment.line.create"
+
+    select_for_payment_move_line = fields.Boolean(
```
florian-dacosta: Rename to `selected_for_payment_filter` for example (not sure that move_line is really relevant here)

account_payment_order_validation_move_line/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,24 @@
...
+        string="Take only to pay",
+        related="order_id.payment_mode_id.default_selected_for_payment_id",
```
florian-dacosta: This field should not be related.
beeing related with readonly=False, if the user changes it on the wizard, it will also change on the payment mode, changing the default for next time!

It should be a regular field, which can be filled in the default.

account_payment_order_validation_move_line/wizard/account_payment_line_create.py
```diff
@@ -0,0 +1,24 @@
...
+        store=True,
+        readonly=False,
+    )
+
+    def _prepare_move_line_domain(self):
+
+        res = super(AccountPaymentLineCreate, self)._prepare_move_line_domain()
+
+        if self.select_for_payment_move_line:
+            res += [("move_id.selected_for_payment", "=", True)]
```
florian-dacosta: I wonder if the condition should be more complicated, because, this selected_for_payment is only meant to be set on invoices, not in regular accounting entry. But this payment order can accept accounting entry that are not invoices, depending on the filter `invoice`.

So I guess the domain as it is now is ok in case `invoice` field is True, but if not, we'd have to have a domain like 
['|', '&', ('move_type', 'in', ("in_invoice", "out_invoice", "in_refund", "out_refund")), ("move_id.selected_for_payment", "=", True), ('move_type', '=', 'entry')]
So I would check
`if self.invoice and self.select_for_payment_move_line`
then add the domain` [("move_id.selected_for_payment", "=", True)]`
`elif self.select_for_payment_move_line`
then add the more complex domain to accept regular accounting entries

florian-dacosta (commented):
I'd change the module name.

Maybe something like account_payment_order_select_for_payment to be consistent as it is kind of a glue module between `account_invoice_select_for_payment` and `account_payment_order`

Also, there is a missing part in the module. Once the payment order is validated (I guess at the step the reconciliation is done), the paid invoices should be unflaggued (the selected_for_payment should be set to False on the invoices, once the payment is done.

PR #1198, opened 2023-12-27
[16.0][FIX]account_banking_mandate: mandate_partner_bank_form columns

Tisho99:
On V16, the mandate_ids field in the mandate_partner_bank_form is not correctly displayed. It has a width of one column and the content of its mandate records can not be properly seen. This PR increments that width.

Before:
image

After:
image

### DIFF
--- a/account_banking_mandate/views/res_partner_bank_view.xml
+++ b/account_banking_mandate/views/res_partner_bank_view.xml
@@ -15,6 +15,7 @@
                         name="mandate_ids"
                         context="{'mandate_bank_partner_view': True}"
                         nolabel="1"
+                        colspan="4"
                     />
                 </group>
             </xpath>


Tisho99:
@HaraldPanten @ValentinVinagre

account_banking_mandate/views/res_partner_bank_view.xml
```diff
@@ -10,11 +10,12 @@
         <field name="inherit_id" ref="base.view_partner_bank_form" />
         <field name="arch" type="xml">
             <xpath expr="//form/sheet/group" position="inside">
-                <group name="mandates" string="Direct Debit Mandates">
+                <group name="mandates" string="Direct Debit Mandates" col="4">
```
pedrobaeza: You shouldn't change the `col`. If the problem is that the width is not completed, please check on the migration guide how it should be done:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

and for these things, always put a screenshot of the before and after for easing reviews.

pedrobaeza (changes requested):

Tisho99:
Col attribute removed

account_banking_mandate/views/res_partner_bank_view.xml
```diff
@@ -15,6 +15,7 @@
                         name="mandate_ids"
                         context="{'mandate_bank_partner_view': True}"
                         nolabel="1"
+                        colspan="4"
```
pedrobaeza, line 18:
```suggestion
                        colspan="2"
```

Tisho99: This is the result with colspan="2"
image

pedrobaeza (approved):
OK then.

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):
LGTM

PR #1199, opened 2024-01-03
[13.0][FIX]account_payment_purchase: return result from onchange and get properties with correct company

GuillemCForgeFlow:
Before the change, the result of the onnchange was not being returned when it should always be for onchange methods.
Also, set the context with the `force_company` again to get the correct value for the `supplier_payment_mode_id` `res.partner` property.

cc @ForgeFlow


GuillemCForgeFlow:
@pedrobaeza Perhaps this can be fast-tracked.

pedrobaeza (approved):
/ocabot merge patch

PR #1200, opened 2024-01-03
[17.0] [MIG] account payment order

angelmoya:


angelmoya:
WIP

I need help to fix this:

account_payment_order/models/account_move.py:171:52: B023 Function definition does not bind loop variable `payment_mode`
account_payment_order/models/account_payment_order.py:426:47: B023 Function definition does not bind loop variable `payment`

PR #1213, opened 2024-01-31
[16.0][FIX] account_payment_order: Keep partner_bank_id correct value on payments belonging to payment order

aritzolea:
Payments (`account.payment`) belonging to a payment order (`account.payment.order`) are assigned the partner bank account on their creation which doesn't follow the pattern defined on the `_compute_available_partner_bank_ids` from the `account` module.

This is not important while the compute method is not triggered, but if it gets triggered the original partner_bank_id value gets replaced by the value defined in the compute method (journal's bank in case of incoming payment orders). This affects to the creation of sepa files, because the bank accounts listed in the file are not correct.

This fix overrides `_compute_available_partner_bank_ids` function on `account_payment_order` module. For payments which belong to a payment order and are generated from payment lines (`account.payment.line`), it assigns the partner_bank_id from the origin payment lines.

### DIFF
--- a/account_payment_order/models/account_payment.py
+++ b/account_payment_order/models/account_payment.py
@@ -2,6 +2,7 @@
 # Copyright 2022 Tecnativa - Pedro M. Baeza
 # Copyright 2023 Noviat
 # Copyright 2024 Tecnativa - Víctor Martínez
+# Copyright 2024 FactorLibre - Aritz Olea
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
@@ -87,3 +88,14 @@ class AccountPayment(models.Model):
         for vals in vals_list:
             vals["date_maturity"] = self.payment_line_ids[0].date
         return vals_list
+
+    def _compute_available_partner_bank_ids(self):
+        """Overrides available_partner_bank_ids compute method for payments which
+        belong to a payment order. Without this override, in case compute method is
+        triggered, these payments' bank value is replaced by an incorrect value."""
+        order_pays = self.filtered(lambda p: p.payment_order_id and p.payment_line_ids)
+        for pay in order_pays:
+            pay.available_partner_bank_ids = pay.payment_line_ids[0].partner_bank_id
+        return super(
+            AccountPayment, self - order_pays
+        )._compute_available_partner_bank_ids()
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -26,6 +26,7 @@ class TestPaymentOrderInboundBase(AccountTestInvoicingCommon):
         cls.partner = cls.env["res.partner"].create(
             {
                 "name": "Test Partner",
+                "bank_ids": [(0, 0, {"acc_number": "XX15001559627230"})],
             }
         )
         cls.inbound_mode = cls.env["account.payment.mode"].create(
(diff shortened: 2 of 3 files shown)


ljsalvatierra-factorlibre (approved):

gerard-vacas (approved):
LGTM

account_payment_order/models/account_payment.py
```diff
@@ -33,3 +34,11 @@ def _compute_payment_method_line_fields(self):
                     )
                 )
         return res
+
+    def _compute_available_partner_bank_ids(self):
+        order_pays = self.filtered(lambda p: p.payment_order_id and p.payment_line_ids)
```
pedrobaeza, line 96: If this finally arrives, you should put a docstring in this override stating why is put.

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -136,3 +137,25 @@ def test_creation(self):
         payment_order.cancel2draft()
         payment_order.unlink()
         self.assertEqual(len(self.payment_order_obj.search(self.domain)), 0)
+
+    def test_payment_partner_bank_value(self):
+        payment_order = self.inbound_order
+        self.assertEqual(len(payment_order.ids), 1)
+
```
pedrobaeza: Better to not add empty lines inside the method for new code. It adds more scroll with no real value.

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -136,3 +137,25 @@ def test_creation(self):
...
+        payment_order.draft2open()
+
+        self.assertEqual(len(payment_order.payment_ids), 1)
+        payment = payment_order.payment_ids[0]
+        self.assertEqual(len(payment.payment_line_ids), 1)
+        payment_line = payment.payment_line_ids[0]
+        self.assertEqual(payment.partner_bank_id, payment_line.partner_bank_id)
+
+        # Assert payment line's bank has not changed
+        payment._compute_available_partner_bank_ids()
```
pedrobaeza: Tests shouldn't launch manually private or compute methods. They should be triggered through business flows.

pedrobaeza (changes requested):
I don't really see the use case for the re-triggering you are talking about, and the tests are just launching manually the compute method.

account_payment_order/models/account_payment.py
aritzolea: Done

account_payment_order/tests/test_payment_order_inbound.py
aritzolea: Done

account_payment_order/tests/test_payment_order_inbound.py
aritzolea: Modified. I'm not calling the `compute` method directly.

account_payment_order/models/account_payment.py
ljsalvatierra-factorlibre: Instead of `incorrect value` i would state which value, the bank from the company.

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -136,3 +137,23 @@ def test_creation(self):
...
+        self.assertEqual(len(payment_order.payment_ids), 1)
+        payment = payment_order.payment_ids[0]
+        self.assertEqual(len(payment.payment_line_ids), 1)
+        payment_line = payment.payment_line_ids[0]
+        self.assertEqual(payment.partner_bank_id, payment_line.partner_bank_id)
+        # Force payment's available banks compute to check parner_bank_id
+        # fields value doesn't change
+        with Form(payment) as payment_form:
+            payment_form.partner_id = self.env["res.partner"]
+            payment_form.partner_id = payment_line.partner_id
```
pedrobaeza: This test case is still a non real one and doesn't reflect a business case. What are you trying to accomplish? Maybe what we have to do is to forbid changing account.payment linked to payment orders.

aritzolea: Modified tests to a more "likely" use case un which a user can accidentaly modify a field on `account.payment` model's form view. `partner_bank_id` field would get an incorrect value.

I recorded a video reproducing the issue. In this case field turns to False because journal has no bank: 

https://github.com/OCA/bank-payment/assets/31987854/894d3bbb-da7e-4c0d-a38e-5fabd3f417c3

ljsalvatierra-factorlibre: Hi @pedrobaeza is it ok?

pedrobaeza: Thanks for the video. I think the proper fix should be to block any change (through an override in the `write` method of the `account.payment`) if it's linked to a payment order.

pedrobaeza:
This is better fixed IMO here: #1346

Do you have anyway the trigger conditions?

aritzolea:
> This is better fixed IMO here: #1346
> 
> Do you have anyway the trigger conditions?

A trigger condicion can be seen in the video I recorded for this PR a few months ago

pedrobaeza:
But I wasn't able to reproduce it on the same conditions, @aritzolea. It seems something that depends on the module stack. Did you do it on runboat? Anyway, as said, the other PR is more focused IMO.

pedrobaeza:
Do you agree then to switch to the other PR?

aritzolea:
> But I wasn't able to reproduce it on the same conditions, @aritzolea. It seems something that depends on the module stack. Did you do it on runboat? Anyway, as said, the other PR is more focused IMO.

I was able to reproduce the error in 16.0 runboat, and the solution in your PR's runboat

pedrobaeza:
OK, then are you ok with continuing with the other PR?

aritzolea:
> OK, then are you ok with continuing with the other PR?

Yes, it can be continued on the other PR.

pedrobaeza:
Thanks.

PR #1214, opened 2024-02-06
[16.0][FIX] account_banking_mandate: Use correct company when search for mandate_id

hugosantosred:
In the context of a multi company user with two companies selected (Company A and Company B) and the Company B as default. If you have a Customer with two mandates (one by each company) and select it in an invoice of the company A, the mandate that is set belongs to the Company B.

This PR adds the company context of the account.move to fix this.

### DIFF
--- a/account_banking_mandate/models/account_move.py
+++ b/account_banking_mandate/models/account_move.py
@@ -22,9 +22,10 @@ class AccountMove(models.Model):
         related="payment_mode_id.payment_method_id.mandate_required", readonly=True
     )
 
-    @api.depends("payment_mode_id", "partner_id")
+    @api.depends("company_id", "payment_mode_id", "partner_id")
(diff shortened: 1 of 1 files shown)


account_banking_mandate/models/account_move.py
```diff
@@ -25,6 +25,7 @@ class AccountMove(models.Model):
     @api.depends("payment_mode_id", "partner_id")
```
pedrobaeza:
```suggestion
    @api.depends("company_id", "payment_mode_id", "partner_id")
```

hugosantosred: Thanks for the review. Changes Added

pedrobaeza (approved):

ljsalvatierra-factorlibre (approved):

PR #1216, opened 2024-02-12
[16.0][FIX] account_payment_order: set invoice name in communication if payment order is inbound

ghost:
Sets invoice name in communication (`account.payment.line`) if payment order is inbound.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -55,6 +55,16 @@ class AccountMove(models.Model):
                 communication = self.payment_reference or self.ref
             else:
                 communication = self.payment_reference or self.name
+            default_payment_type = self._context.get("default_payment_type", False)
+            default_payment_order_id = self._context.get("default_payment_order_id", [])
+            if not default_payment_type and default_payment_order_id:
+                default_payment_type = (
+                    self.env["account.payment.order"]
+                    .browse(default_payment_order_id)
+                    .payment_type
+                )
+            if default_payment_type == "inbound":
+                communication = self.name
         return communication or ""
 
     def _get_payment_order_communication_full(self):
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -60,7 +60,9 @@ class AccountMoveLine(models.Model):
 
     def _prepare_payment_line_vals(self, payment_order):
         self.ensure_one()
-        communication_type, communication = self._get_communication()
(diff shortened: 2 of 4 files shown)


account_payment_order/models/account_move.py
```diff
@@ -57,6 +57,9 @@ def _get_payment_order_communication_direct(self):
                 communication = self.ref or self.payment_reference
             else:
                 communication = self.payment_reference or self.name
+        default_payment_type = self._context.get("default_payment_type", False)
+        if default_payment_type == "inbound":
+            communication = self.name
```
aritzolea: Should the name also de assigned if the account.move is not an invoice?

ghost: Changed so that this only applies when it's an invoice

aritzolea (approved):

pedrobaeza:
Remove draft state when ready to be reviewed

SoniaViciana (approved):
Functional review done. With this change, when you create a payment order (charge), the communication is the invoice number instead of the sale order number. I think it makes more sense.

Thanks!

ghost:
Hello @OCA/banking-maintainers, 

Could you review this? 

Thank you!

alexey-pelykh (commented):
could you please also add tests without default_payment_type in context to ensure nothing breaks for those cases?

account_payment_order/models/account_move.py
```diff
@@ -57,6 +57,16 @@ def _get_payment_order_communication_direct(self):
                 communication = self.ref or self.payment_reference
             else:
                 communication = self.payment_reference or self.name
+            default_payment_type = self._context.get("default_payment_type", False)
```
etobella, line 58: Why are you using the default_payment_type?

If you want to use name, just leave payment_reference and ref to false :thinking:

pedrobaeza (changes requested):
This is not correct.

The payment reference field is clear: it's for setting the reference of the payment. Setting it, it should have preference over the invoice number. And anyway, by default, the payment reference is set with the invoice number.

Which is the problem this is trying to fix?

SoniaViciana:
When this PR was proposed the "payment reference" was completed by default with the sales order number instead of invoice number.

This should have been solved in parallel since it does not happen today:

image

We can close this issue.

pedrobaeza:
For other times, it's good to add proper explanations since the beginning and in the commit message, so when recalling this, the reasons are clear and can be contrasted.

PR #1218, opened 2024-02-13
[14.0] account_payment_partner: remove dead code about sepa

hparfr:
...or you may call it indirect depency

looks like account_payment_mandate is not in the dependencies since the introduction of this code in v10.


hparfr:
please have a look @pedrobaeza


Ref https://github.com/OCA/bank-payment/pull/1135#issuecomment-1938176670

pedrobaeza (approved):
Yes, it seems weird to have that code in the base module. We don't assign anyway banks nor mandate in invoices. We delay its assignation till the order creation.

pedrobaeza:
Closing in favor of #1219 

Now I understand better that code. It's for showing the customer bank account for SEPA DD.
 I saw it bad before thinking that it was code for assigning the mandate in the invoice.

PR #1219, opened 2024-02-13
[14.0] account_payment_partner: refactor partner_banks_to_show

hparfr:
move partner_bank_to_show into account_payment_mode to allow it's reuse from other models like sale_order

depends of #1218


account_payment_partner/models/account_move.py
```diff
@@ -142,22 +142,7 @@ def partner_banks_to_show(self):
         self.ensure_one()
         if self.partner_bank_id:
             return self.partner_bank_id
-        if self.payment_mode_id.show_bank_account_from_journal:
-            if self.payment_mode_id.bank_account_link == "fixed":
```
pedrobaeza, line 146: Include an override of `partner_banks_to_show` in SEPA DD module including this, indicating that the aim of this is to show the bank account where the SEPA DD is going to be charged.

hparfr: [account_banking_sepa_direct_debit](https://github.com/OCA/bank-payment/blob/14.0/account_banking_sepa_direct_debit/__manifest__.py#L14C47-L14C70) do not depend of this module.

pedrobaeza: Then that's the reason for doing such "ugly thing". We need to show the customer account for SEPA DD.

hparfr: The real thing would have to create a glue module for that.

Anyway, this small change (<10 len if), start to cost me a lot in refactorings for features I don't need.

pedrobaeza: Yeah, the feature was there before the refactoring. We can't lose it. I think we can keep the hack without doing a glue module, although it doesn't seem very elegant.

hparfr:
glue module added

florian-dacosta (approved):
LGTM

PR #1221, opened 2024-02-16
[16.0][ADD] account_payment_method_fs_storage: init

AnizR:
[New module]

**account_payment_method_fs_storage**

This Odoo addon enhances the payment mode functionality by introducing the ability to designate a specific storage location. Users can now specify a storage destination within the payment method, facilitating the automatic transfer of generated files to the designated location upon completing a payment transaction. 

Co-author: @acsonefho


acsonefho (approved):

account_payment_method_fs_storage/models/account_payment_method.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountPaymentMethod(models.Model):
+
+    _inherit = "account.payment.method"
+
+    storage_id = fields.Many2one(
```
adrienpeiffer: I have my doubts on this point. What are the access rights on this model? Accountants will be able to select any record. Not sure that's a good thing. I think it would be better to define on a storage that it can be used in this case and filter the result.

adrienpeiffer (changes requested):

account_payment_method_fs_storage/models/account_payment_method.py
AnizR: My proposition:

- Introduce a boolean on fs.storage that tells if storage can be used on Payment Method
- Do not give access to fs.storage but run in sudo the push operation to storage
- Use a selection "storage" that represents the list of storage that can be used on Payment Method

account_payment_method_fs_storage/models/fs_storage.py
```diff
@@ -0,0 +1,28 @@
...
+class FsStorage(models.Model):
+
+    _inherit = "fs.storage"
+
+    use_on_payment_method = fields.Boolean(
+        help="Tells if storage can be set on payment methods",
+    )
+
+    @api.constrains("use_on_payment_method")
+    def _check_use_on_payment_method(self):
```
acsonefho, line 17: You should add a test for this

acsonefho (changes requested):

account_payment_method_fs_storage/models/fs_storage.py
AnizR: Done :+1:

acsonefho (approved):
GJ :muscle:

adrienpeiffer (approved):
Code review

adrienpeiffer:
Thanks @acsonefho @AnizR

PR #1222, opened 2024-02-16
[16.0][IMP] account_payment_order: No error on unknown payment method codes

pedrobaeza:
Forward-port of #1217

Since v14, the tool for changing the target transfer account is to create new payment methods, and put a method line in the journal with such account.

Thus, there will be "fake" payment methods with no export handler with the only goal of having a different target account.

Raising an error on such methods when finishing the payment order, forcing to add custom code for returning (False, False), is too much, so better to simply return that by default and don't block in these cases.

@Tecnativa TT47723

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -376,17 +376,13 @@ class AccountPaymentOrder(models.Model):
         return True
 
     def generate_payment_file(self):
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

flotho (approved):
Thanks,
code review
Functionnal understood
LGTM

PR #1225, opened 2024-02-28
[16.0][IMP] account_payment_purchase: Add propagation of payment mode from purchase to invoice

atorreiro:
When you set a payment mode in the purchase order and then you generate an invoice, the payment mode is not propagated. I propose to inherit _prepare_invoice method and return the payment mode.

### DIFF
--- a/account_payment_purchase/models/account_move.py
+++ b/account_payment_purchase/models/account_move.py
@@ -19,8 +19,7 @@ class AccountMove(models.Model):
             self.purchase_vendor_bill_id.purchase_order_id.supplier_partner_bank_id.id
             or self.purchase_id.supplier_partner_bank_id.id
         )
-
-        res = super()._onchange_purchase_auto_complete() or {}
+        res = {}
         if self.payment_mode_id and new_mode and self.payment_mode_id.id != new_mode:
             res["warning"] = {
                 "title": _("Warning"),
@@ -37,4 +36,5 @@ class AccountMove(models.Model):
             return res
         elif self.partner_bank_id.id != new_bank:
             self.partner_bank_id = new_bank
+        res = super()._onchange_purchase_auto_complete()
         return res
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -53,3 +53,9 @@ class PurchaseOrder(models.Model):
             else:
                 order.supplier_partner_bank_id = False
                 order.payment_mode_id = False
+
+    def _prepare_invoice(self):
+        """Propagate payment mode to invoice"""
+        invoice_vals = super()._prepare_invoice()
+        invoice_vals["payment_mode_id"] = self.payment_mode_id.id
+        return invoice_vals
--- a/account_payment_purchase/tests/test_account_payment_purchase.py
+++ b/account_payment_purchase/tests/test_account_payment_purchase.py
@@ -111,6 +111,16 @@ class TestAccountPaymentPurchase(TransactionCase):
         self.assertEqual(
             result and result.get("warning", {}).get("title", False), "Warning"
         )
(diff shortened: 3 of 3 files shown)


aritzolea (changes requested):
Adapt tests so they don't fail and also check on tests the value is propagated to the invoice when it's been created from the purchase order, please

aritzolea:
@atorreiro Are you continuing the PR?

aritzolea (approved):

atorreiro:
> @atorreiro Are you continuing the PR?

Sorry for the delay, I have added the tests and fixes in a new commit. I have changed the order of the super in method '_onchange_purchase_auto_complete' because this changes the payment mode of invoice and is never returned a warning if the new mode is different from the previous payment mode of invoice.

PabloMartFL (approved):
Funtional review ok, when the purchase order has a payment mode, the invoice take that payment mode

https://www.awesomescreenshot.com/video/26350246?key=9b63b1e27735d75c19be31c1147a0fb1

Just wanna point out that, if the partner has a payment mode established, the invoice allways take the payment mode in the purchase order

https://www.awesomescreenshot.com/video/26350200?key=bfe3620fe26e3b96eca41aa7da3ef4d3

account_payment_purchase/models/account_move.py
```diff
@@ -37,4 +36,5 @@ def _onchange_purchase_auto_complete(self):
             return res
         elif self.partner_bank_id.id != new_bank:
             self.partner_bank_id = new_bank
+        res = super()._onchange_purchase_auto_complete()
```
pedrobaeza, line 39: Why this change not commented anywhere?

pedrobaeza (changes requested):
And please do some community reviewing PRs from other companies. A related one is for example in #1247

account_payment_purchase/models/account_move.py
atorreiro: Do you mean commenting it in the code? The reason of changing the order of super is because when it is called, the payment mode of invoice changes and then self.payment_mode_id and new_mode are the same at the point it is checked. If you check "_onchange_purchase_auto_complete" method in purchase, it calls the _prepare_invoice and changes the payment mode, so it was not returning a warning when new mode was different from the previous one.

pedrobaeza: I mean that you are not mentioning this change, its purpose, and so on, and it's totally unrelated to the main purpose of this PR. Remove it, and you can propose it in another one.

atorreiro: But I mean that it is related to the change I have made in _prepare_invoice method to propagate the payment mode, in my opinion it is not an independent PR. Previously, it was correct because when the payment mode was different a warning was returned, but now with the change in _prepare_invoice, it is required to change the order of the super or the payment mode of invoice will change before checking if it is the different or not to the previous one.

pedrobaeza: But the idea is to change it and show the warning. With this change, there will never be any warning...

atorreiro: Here I show examples with purchase orders P0010 and P0011, that have different payment modes.

When creating an invoice directly from purchase order:
https://github.com/OCA/bank-payment/assets/128492020/434c570b-b1b9-4dab-a179-1e312fc2e7a7

First creating an invoice and then adding the purchase orders to the invoice:
https://github.com/OCA/bank-payment/assets/128492020/7c15a3cc-3ba7-4a19-90ec-bda4d3cd5e4b

This is not the way it should work?

pedrobaeza: Yes, as the lines are added as they were before. You just receive the warning.

aritzolea:
@atorreiro Could you rebase? Branch has conflicts

PR #1237, opened 2024-03-07
[15.0][FIX] account_payment_order: Filter bank on company

hhgabelgaard:
In multi company setup, we need to filter res.partner.bank for current (payment order) company, or shared partner bank account (No company set)

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
-from odoo.fields import first
 
 
 class AccountMoveLine(models.Model):
@@ -70,7 +69,11 @@ class AccountMoveLine(models.Model):
             # in this case
         if payment_order.payment_type == "outbound":
             amount_currency *= -1
-        partner_bank_id = self.partner_bank_id.id or first(self.partner_id.bank_ids).id
+        partner_bank_id = self.partner_bank_id.id
+        if not partner_bank_id:
+            partner_bank_id = self.partner_id.bank_ids.filtered(
+                lambda bank: not bank.company_id or bank.company_id == self.company_id
+            )[:1].id
         vals = {
             "order_id": payment_order.id,
             "partner_bank_id": partner_bank_id,
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -142,7 +142,9 @@ class AccountPaymentLine(models.Model):
     def partner_id_change(self):
         partner_bank = False
         if self.partner_id.bank_ids:
-            partner_bank = self.partner_id.bank_ids[0]
+            partner_bank = self.partner_id.bank_ids.filtered(
+                lambda bank: not bank.company_id or bank.company_id == self.company_id
+            )[:1]
         self.partner_bank_id = partner_bank
 
     @api.onchange("move_line_id")


account_payment_order/models/account_move_line.py
```diff
@@ -70,7 +69,14 @@ def _prepare_payment_line_vals(self, payment_order):
             # in this case
         if payment_order.payment_type == "outbound":
             amount_currency *= -1
-        partner_bank_id = self.partner_bank_id.id or first(self.partner_id.bank_ids).id
+
```
pedrobaeza: Please don't add extra empty lines inside a method.

pedrobaeza (commented):
I understand that the problem is that you have selected more than one company at the same time when preparing the payment order. I think the cleanest patch should be to do `self.partner_id.with_context(allowed_company_ids=self.company_id).bank_ids` instead. What do you think?

hhgabelgaard:
> I understand that the problem is that you have selected more than one company at the same time when preparing the payment order.

No, Only one company selected. But this shouldn't depend on access / ir.rule
> I think the cleanest patch should be to do 
`self.partner_id.with_context(allowed_company_ids=self.company_id).bank_ids` instead. What do you think?

Didn't know that trick - And it seems to depend on access / ir.rules. i.e won't work in a background job run by the system user
My inspiration was how core does it, e.g. https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L532

account_payment_order/models/account_payment_line.py
```diff
@@ -142,7 +142,11 @@ def payment_line_hashcode(self):
     def partner_id_change(self):
         partner_bank = False
         if self.partner_id.bank_ids:
-            partner_bank = self.partner_id.bank_ids[0]
+            bank_ids = self.partner_id.bank_ids.filtered(
+                lambda bank: bank.company_id is False
+                or bank.company_id == self.company_id
+            )
+            partner_bank = bank_ids and bank_ids[0]
```
pedrobaeza:
```suggestion
            partner_bank = self.partner_id.bank_ids.filtered(
                lambda bank: not bank.company_id
                or bank.company_id == self.company_id
            )[:1]
```

account_payment_order/models/account_move_line.py
```diff
@@ -70,7 +69,14 @@ def _prepare_payment_line_vals(self, payment_order):
...
+        partner_bank_id = self.partner_bank_id.id
+        if not partner_bank_id:
+            bank_ids = self.partner_id.bank_ids.filtered(
+                lambda bank: bank.company_id is False
+                or bank.company_id == self.company_id
+            )
+            partner_bank_id = bank_ids and bank_ids[0].id
```
pedrobaeza:
```suggestion
        partner_bank_id = self.partner_bank_id.id
        if not partner_bank_id:
            partner_bank_id = self.partner_id.bank_ids.filtered(
                lambda bank: not bank.company_id
                or bank.company_id == self.company_id
            )[:1].id
```

hhgabelgaard:
@pedrobaeza Your suggestions has been applied.
Thanks for the input/review

pedrobaeza:
You have modified slightly my suggestions, why? Mine is more compact and descriptive.

hhgabelgaard:
Because I assumed this [:1].id would fail with IndexError: tuple index out of range if the filtered list is empty....
As [0] does...
But It seems to work beautifully, and yes, it is more compact and descriptive.

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

PR #1247, opened 2024-04-04
[15.0][FIX] account_payment_purchase: Leave the partner_bank_id field empty in purchase invoice process

victoralmau:
Leave the `partner_bank_id` field empty in purchase invoice process

Example use case:
- Create a Supplier and set a bank account.
- Creating a purchase order with a Supplier Bank Account
- The payment method of the payment method must have the Bank Account Required box unchecked.
- Confirm the purchase order and create the invoice with the "Create Bill" button.
- The invoice must have the bank account empty.

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT46502

### DIFF
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -29,7 +29,7 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("partner_id", "company_id")
     def onchange_partner_id(self):
-        res = super(PurchaseOrder, self).onchange_partner_id()
+        res = super().onchange_partner_id()
         if self.partner_id:
             self = self.with_company(self.company_id)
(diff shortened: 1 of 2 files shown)


victoralmau:
IMO there are "_inconsistent_" things in the purchase orders, specifically the _Supplier Bank Account_ field. I think that it should be a compute store field and have the same logic as `account_payment_partner` (https://github.com/OCA/bank-payment/blob/15.0/account_payment_partner/models/account_move.py#L129). What do you think?

pedrobaeza (approved):
The computed writable field will be an improvement, but let's limit for now this to the current problem.

victoralmau:
Ping @sergio-teruel

pilarvargas-tecnativa (approved):

PR #1248, opened 2024-04-07
[FIX][17.0] account_payment_order: Fix dynamic domain

ramiadavid:
The return domain in onchange is deprecated in v17 and does not work. A calculated Json field was created to store the domain.

### DIFF
--- a/account_payment_order/wizard/account_payment_line_create.py
+++ b/account_payment_order/wizard/account_payment_line_create.py
@@ -41,6 +41,7 @@ class AccountPaymentLineCreate(models.TransientModel):
     move_line_ids = fields.Many2many(
         comodel_name="account.move.line", string="Move Lines"
     )
+    move_line_domain = fields.Binary(compute="_compute_move_line_domain")
 
     @api.model
     def default_get(self, field_list):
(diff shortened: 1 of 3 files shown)


ramiadavid:
@pedrobaeza I have detected this problem that I did not detect when I migrated the module. I create this PR to solve it.

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -41,6 +41,7 @@ class AccountPaymentLineCreate(models.TransientModel):
     move_line_ids = fields.Many2many(
         comodel_name="account.move.line", string="Move Lines"
     )
+    move_line_domain = fields.Json()
```
pedrobaeza: Usually this is saved as `Binary`, not `Json`. Has this changed in v17?

account_payment_order/wizard/account_payment_line_create.py
```diff
@@ -178,8 +179,7 @@ def populate(self):
     )
     def move_line_filters_change(self):
         domain = self._prepare_move_line_domain()
-        res = {"domain": {"move_line_ids": domain}}
-        return res
+        self.move_line_domain = domain
```
pedrobaeza: Do it a computed non stored field instead of this.

ramiadavid: Updated

account_payment_order/wizard/account_payment_line_create.py
ramiadavid: Well apparently it works fine with both, I'll change it to Binary

pedrobaeza (approved):
Perfect! Fast-tracking this for fixing it: 

/ocabot merge minor

PR #1249, opened 2024-04-09
[14.0][FIX] account_payment_order: exclude cancel journal entries

alexis-via:
Always exclude cancel journal entries in the wizard that adds payment lines to payment order.

Same behavior as in the rest of Odoo (official and OCA modules). For example, this is the behavior of the OCA module account_financial_report
https://github.com/OCA/account-financial-reporting/blob/14.0/account_financial_report/report/trial_balance.py#L75


pedrobaeza (approved):

rvalyi (approved):

pedrobaeza:
15.0 and 16.0 are pending fw-ports:

/ocabot merge patch

PR #1256, opened 2024-04-25
[17.0][FIX] account_banking_mandate: change name_get method to use _compute…

MarwanBHL:
switch `name_get` to `_compute_display_name`

`name_get` is deprecated in 17.0 see : https://github.com/odoo/odoo/pull/122085


pedrobaeza (approved):
Fast-tracking this straight-forward patch:

/ocabot merge patch

PR #1259, opened 2024-05-03
[15.0][IMP] account_banking_mandate_contact: Improve tests (remove mock)

victoralmau:
`account_banking_mandate_contact`: Improve tests
Changes done:
- Add _get_payment_method_information() compatibility from account_banking_mandate
- Remove mock from account_banking_mandate_contact tests

@Tecnativa

### DIFF
--- a/account_banking_mandate/models/account_payment_method.py
+++ b/account_banking_mandate/models/account_payment_method.py
@@ -1,7 +1,7 @@
 # Copyright 2016-2020 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class AccountPaymentMethod(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This still seems to be incorrect, but anyway, doing the mock sounds very weird and a hack. We shouldn't need to do any mock on these tests. Do you know the original intention of them? Can we remove it and fix tests without the mocks?

victoralmau:
This fix is the same as the one done in the commit I mention and is a side effect that affects many modules causing all tests to fail. What other change do you propose to solve it?

pedrobaeza:
Well,tests are red, so it's not working, but I'm just saying that a mock seems weird.

victoralmau:
Extra changes done and mocks have been removed + update the PR description

account_payment_order/models/account_payment_method.py
```diff
@@ -13,3 +14,14 @@ class AccountPaymentMethod(models.Model):
         "some payment methods.",
         default=False,
     )
+
+    @api.model
+    def _get_payment_method_information(self):
+        """Add extra codes only from account_payment_order tests."""
+        res = super()._get_payment_method_information()
+        if config["test_enable"] and self.env.context.get("test_account_payment_order"):
```
rousseldenis: @victoralmau That's a regression IMHO. You can't embed test feature in code... it's nonsense !

lmignon (changes requested):
@victoralmau IMO, replacing mock by test feature into the original implementation is not an improvement it's a hack. Can you keep the mock plz.

victoralmau:
`account_payment_order` changes removed.

pedrobaeza (approved):

rousseldenis (approved):

PR #1260, opened 2024-05-03
[16.0][FIX] account_payment_partner: payment mode filter for credit note/refund

jbaudoux:
Allow to freely choose inbound or outbound payment mode for credit note/refund

When you create a credit note, you may want to include it in the next debit order to decrease the amount debited from open invoices. For example, partnerA has an invoice for $500 and a credit note for $100. You want to make a debit order for $400. For this you need to put the same payment mode on the invoice and credit note.

This behavior has been lost in the migration to v13:
* In v12: https://github.com/OCA/bank-payment/blob/12.0/account_payment_partner/views/account_invoice_view.xml#L29
* In migration PR: https://github.com/OCA/bank-payment/pull/649/files#diff-560b5e454acd270e25223d56b85967eaec36fe9a08eac999124990650aa32f6cR46

cc @rven @alexis-via @pedrobaeza @lmignon @sbidoul

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -9,9 +9,6 @@ from odoo import api, fields, models
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    payment_mode_filter_type_domain = fields.Char(
-        compute="_compute_payment_mode_filter_type_domain"
-    )
     partner_bank_filter_type_domain = fields.Many2one(
         comodel_name="res.partner", compute="_compute_partner_bank_filter_type_domain"
     )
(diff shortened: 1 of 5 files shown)


pedrobaeza:
AFAIK, this was explicitly removed by @alexis-via in his v9 refactoring, but not sure if it was re-added later. The idea is that you first reconcile the $100 refund with the $500 invoice, and the add the invoice with remaining $400.

I'm not against to have this though if everything is under control.

jbaudoux:
> AFAIK, this was explicitly removed by @alexis-via in his v9 refactoring, but not sure if it was re-added later. The idea is that you first reconcile the $100 refund with the $500 invoice, and the add the invoice with remaining $400.
> 
> I'm not against to have this though if everything is under control.

Thanks for the quick feedback. That's indeed a workaround. However, my issue is the consequence for the debited party that won't get the proper detail of the debit order as he will see a partial debit for an invoice instead of a full debit of an invoice and a credit of a credit note.

pedrobaeza:
Yes, that's the only cons I see also: or you change the communication manually to reflect both the credit note and the invoice, or it's confusing for the other party. Another option is to reflect automatically in the communication the invoice and possible reconciled credit notes. It still requires the extra step of reconciling both though.

jbaudoux:
test fixed

lmignon (approved):
LGTM (Code review)

pedrobaeza:
I think this change should include at least a couple of tests for a payment order with both credit and normal operation checking the whole final amount, and another with only credit to see if there's an error due to the negative amount.

lmignon:
@jbaudoux CAn you reopen this one? It should be merged since it's used in production with success for more than 2 years.

PR #1161, opened 2023-10-18
[16.0][MIG] account_payment_order_tier_validation: Migration to 16.0

vincent-hatakeyama:

PR #1262, opened 2024-05-03
[17.0][ADD] account_payment_order_tier_validation: new module to add tier validation to payment orders

alan196:
This module allows to configure a Tier Validation for account payment orders


StefanRijnhart:
The same module already exists: here is the V16 migration of it: https://github.com/OCA/bank-payment/pull/1161. I like that one better because it has more comprehensive tests. Would you consider to migrate that one instead?

PR #1246, opened 2024-04-03
[IMP] account_payment_sale: Add payment mode in invoicing grouping criteria

rousseldenis:

PR #1268, opened 2024-05-16
[16.0][FIX] account_payment_sale: Force add payment mode in invoice vals

ramiadavid:
Since this PR #1246 was added, an error occurs when there are orders without payment_mode_id along with others that do have them when comparing None > int
`TypeError: '<' not supported between instances of 'NoneType' and 'int' 

The error occurs here:
https://github.com/odoo/odoo/blob/fe2fb05892e273587996a2f68cd53fc4f8604c6e/addons/sale/models/sale_order.py#L1149-L1156

Fixed by forcing payment_mode_id to be added even if it is empty because False can be compared with integer

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -26,15 +26,15 @@ class SaleOrder(models.Model):
                 order.payment_mode_id = False
 
     def _get_payment_mode_vals(self, vals):
-        if self.payment_mode_id:
-            vals["payment_mode_id"] = self.payment_mode_id.id
-            if (
-                self.payment_mode_id.bank_account_link == "fixed"
-                and self.payment_mode_id.payment_method_id.code == "manual"
-            ):
-                vals[
-                    "partner_bank_id"
-                ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
+        vals["payment_mode_id"] = self.payment_mode_id.id
+        if (
+            self.payment_mode_id
+            and self.payment_mode_id.bank_account_link == "fixed"
+            and self.payment_mode_id.payment_method_id.code == "manual"
+        ):
+            vals[
+                "partner_bank_id"
+            ] = self.payment_mode_id.fixed_journal_id.bank_account_id.id
 
     def _prepare_invoice(self):
         """Copy bank partner from sale order to invoice"""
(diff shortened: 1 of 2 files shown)


francesco-ooops:
@rousseldenis can you confirm the issue?

ramiadavid:
@francesco-ooops I have added a test to see the error

ramiadavid:
Pre-commit fails due to another module: `account_banking_pain_base`

PicchiSeba:
@ramiadavid Can you rebase? This might fix the pre-commit error

PicchiSeba (approved):
Code Review: LGTM

pedrobaeza:
Although I agree there should be a solution, the intention of no setting such vals is to keep the partner's payment mode in the moment of the invoice creation, so what about assigning such payment mode following the same rules as if you select the partner in the invoice?

rousseldenis:
> @rousseldenis can you confirm the issue?

I'll check but indeed this seems correct.

rousseldenis (approved):

pedrobaeza (changes requested):
Please check my comment

rousseldenis:
> Although I agree there should be a solution, the intention of no setting such vals is to keep the partner's payment mode in the moment of the invoice creation, so what about assigning such payment mode following the same rules as if you select the partner in the invoice?

Because payment mode could be set manually on the order side and should be kept. If I have understood your question

pedrobaeza:
It's not a direct question, but how to implement the fix.

pedrobaeza:
I clarify more: I'm asking for not putting `vals["payment_mode_id"] = False`, but in case of not having any payment mode in the sales order, to assign the payment mode following this logic:

https://github.com/OCA/bank-payment/blob/2f446b3dce73b327a0528369c348dd708b79aca0/account_payment_partner/models/account_move.py#L62

which is what happens currently.

pedrobaeza:
The other option is to strip out the vals entry before creation if False for letting the computed to act.

This can be more optimal for avoiding to duplicate code.

rousseldenis:
> I clarify more: I'm asking for not putting `vals["payment_mode_id"] = False`, but in case of not having any payment mode in the sales order, to assign the payment mode following this logic:
> 
> https://github.com/OCA/bank-payment/blob/2f446b3dce73b327a0528369c348dd708b79aca0/account_payment_partner/models/account_move.py#L62
> 
> which is what happens currently.

Yes, I've just seen it.

@ramiadavid Could you maybe try to remove the get_payment_mode_vals() function and override the compute methods defined in account_payment_partner instead ?

PR #1269, opened 2024-05-17
[16.0][IMP] account_payment_sale: Improve tests to prevent error with pricelist_id field

victoralmau:
Improve tests to prevent error with `pricelist_id` field

Remove the `pricelist_id` assignment (leave the default one) because it may fail if the user does not have the `product.group_product_product_pricelist` group

`AssertionError: can't write on invisible field pricelist_id`

Locked by:
- [x] `fs_storage`: https://github.com/OCA/storage/pull/372 (Related to `account_payment_method_fs_storage` tests)

@Tecnativa

### DIFF
--- a/account_payment_sale/tests/test_sale_order.py
+++ b/account_payment_sale/tests/test_sale_order.py
@@ -10,7 +10,6 @@ class TestSaleOrder(CommonTestCase):
     def create_sale_order(self, payment_mode=None):
         with Form(self.env["sale.order"]) as sale_form:
             sale_form.partner_id = self.base_partner
-            sale_form.pricelist_id = self.env.ref("product.list0")
             for (_, p) in self.products.items():
                 with sale_form.order_line.new() as order_line:
                     order_line.product_id = p


account_payment_sale/tests/test_sale_order.py
```diff
@@ -8,16 +8,21 @@
...
-                    order_line.product_uom = p.uom_id
-                    order_line.price_unit = p.list_price
+        # Apply default to avoid error if user does not have
+        # product.group_product_pricelist group
+        sale_form = Form(
+            self.env["sale.order"].with_context(
+                default_pricelist_id=self.env.ref("product.list0").id
+            )
+        )
+        sale_form.partner_id = self.base_partner
```
pedrobaeza: Here the pricelist is changed to the one of the partner, so the approach shouldn't be valid.

victoralmau: Ok I understand, better then remove the `pricelist_id` assignment and leave the corresponding one when defining the partner.

pedrobaeza:
But if you don't put the previous error, we can't know if this is the solution. Please always be the most verbose possible.

victoralmau:
> But if you don't put the previous error, we can't know if this is the solution. Please always be the most verbose possible.

Sorry, I thought the explanation that the user may not have permissions (`product.group_product_product_product_pricelist`) was enough.

I have added the error shown in the PR description.

Do you think another approach is better to avoid this possible error? IMO I think that you should not change `pricelist_id` field in a sales order, if necessary in some case you define the corresponding default pricelist in the partner to be used.

pedrobaeza (approved):
As tests don't depend on the exact price (as there may be the possible problem of a currency switch), it's OK to not fix the pricelist.

Merging for continuing the integration tests cleaning:

/ocabot merge patch

PR #1270, opened 2024-05-20
[16.0] [FIX] account_payment_sale: correction for grouping criteria on invoices creation when payment_mode_id is not set

sergiobstoj:
When having invoices where `payment_mode_id` is `None` and then trying to create invoices from tree view with other invoices with `payment_mode_id` assigned is returning an Error:
```
File "/opt/odoo/odoo-base/odoo-server/addons/sale/models/sale_order.py", line 1151, in _create_invoices
    invoice_vals_list = sorted(
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```
Here we create inactive records to handle this issue, assigning `payment_mode_momentary` to the invoices with no `payment_mode_id` before the creation of the invoices. After that, we unassign `payment_mode_id` from the invoices.

### DIFF
--- a/account_payment_sale/data/account_payment_mode.xml
+++ b/account_payment_sale/data/account_payment_mode.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="payment_method_momentary" model="account.payment.method">
+        <field name="name">Momentary Payment Method</field>
+        <field name="code">0000</field>
+        <field name="payment_type">inbound</field>
+        <field name="active" eval="False" />
+    </record>
+
+    <record id="payment_mode_momentary" model="account.payment.mode">
+        <field name="name">Momentary Payment Mode</field>
+        <field name="payment_method_id" ref="payment_method_momentary" />
+        <field name="company_id" ref="base.main_company" />
+        <field name="bank_account_link">variable</field>
+        <field name="active" eval="False" />
+    </record>
+</odoo>
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -52,3 +52,17 @@ class SaleOrder(models.Model):
         if "payment_mode_id" not in keys:
             keys.append("payment_mode_id")
         return keys
+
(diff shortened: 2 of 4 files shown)


account_payment_sale/data/account_payment_mode.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="payment_method_momentary" model="account.payment.method">
+        <field name="name">Momentary Payment Method</field>
+        <field name="code">0000</field>
+        <field name="payment_type">inbound</field>
+        <field name="active" eval="False" />
+    </record>
+
```
aritzolea, line 9: Avoid empty spaces

account_payment_sale/models/sale_order.py
```diff
@@ -52,3 +52,17 @@ def _get_invoice_grouping_keys(self) -> list:
         if "payment_mode_id" not in keys:
             keys.append("payment_mode_id")
         return keys
+
+    def _create_invoices(self, grouped=False, final=False, date=None):
+        momentary_mode = self.env.ref("account_payment_sale.payment_mode_momentary")
+        for order in self:
+            if not order.payment_mode_id:
+                order.payment_mode_id = momentary_mode
+
```
aritzolea, line 61: Avoid empty spaces

account_payment_sale/models/sale_order.py
```diff
@@ -52,3 +52,17 @@ def _get_invoice_grouping_keys(self) -> list:
...
+        res = super()._create_invoices(grouped=grouped, final=final, date=date)
```
aritzolea, line 62: Invoices are being created with that new payment mode, you're unassigning it from the orders but not from the invoice.

aritzolea:
Add `[16.0]` on the title, just at the beginning. Without that you cannot know the versión on the pull request list view.

Also change from `IMP` to `FIX`

aritzolea:
Add also a test in which the traceback is thrown without this fix

pedrobaeza:
You should find another way, as creating a record just for that is not correct. Imagine all the possible error scenarios, like someone removing such payment mode in the DB.

PR #1276, opened 2024-05-28
[16.0][IMP] account_payment_order: Mark invoices in a payment order as in payment

pedrobaeza:
Odoo core already includes a mechanism, only activated in enterprise, to mark invoices reconciled against a payment as "In payment", not as "Paid", and then when the bank reconciliation is done, passed to "Paid".

As we already refactored the payment order transfer entries generation to be done by payments, the only missing bits is to override the method that returns the payment state to put.

We have limited this new payment state though to those invoices that are part of a payment order, not all including those which were paid pressing the button "Register payment", as that's out of the scope of this module.

@Tecnativa TT49386

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -233,3 +233,12 @@ class AccountMove(models.Model):
             }
         )
         return action
+
+    @api.model
+    def _get_invoice_in_payment_state(self):
+        """Called from _compute_payment_state method.
+        Consider in_payment all the moves that are included in a payment order.
+        """
+        if self.line_ids.payment_line_ids:
+            return "in_payment"
+        return super()._get_invoice_in_payment_state()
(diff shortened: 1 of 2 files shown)


ValentinVinagre:
@HaraldPanten @Jesarregui FYI could u review 😄 ?

carolinafernandez-tecnativa (approved):
Functional review LGTM, Thanks!

victoralmau (approved):

carlosdauden (approved):

HaraldPanten:
@pedrobaeza Just to understand it better:

What's the point on considering only the invoices related to a payment order as "In payment" instead of considering all the paid invoices but unreconciled?

pedrobaeza:
It's for limiting the scope of the change to the scope of the module: having `account_payment_order` installed that modifies also "Register payment" button to go to "In payment" doesn't seem correct to me.

HaraldPanten (approved):
Functional review 👍

ValentinVinagre (approved):
LGTM 😄

PR #1279, opened 2024-05-31
[IMP] Allow to use the storage either from the payment mode or from the payment method

lmarion-source:
Initially, the storage was found only on the payment method. We added the storage at the payment mode level to allow different storage for different companies.


account_payment_method_or_mode_fs_storage/models/account_payment_mode.py
```diff
@@ -0,0 +1,25 @@
...
+        help="Storage where to put the file after generation",
+    )
+
+    def _get_selection_storage(self):
+        """
+        This is to avoid giving access to the model fs.storage
+        at groups other than base.group_system
+        """
+        storages = (
+            self.env["fs.storage"].sudo().search([("use_on_payment_method", "=", True)])
```
AnizR: I think that it should be 'use_on_payment_mode'?

account_payment_method_or_mode_fs_storage/tests/test_account_payment_method_or_mode_fs_storage.py
```diff
@@ -127,3 +133,69 @@ def test_check_use_on_payment_method(self):
 
         self.assertFalse(self.payment_method.storage)
         self.assertFalse(self.fs_storage.use_on_payment_method)
+
+    def test_payment_mode_fs_storage(self):
+        self.company.update(
```
AnizR, line 213: I think that you need to adapt the test to change the 'source' of the fs_storage

AnizR (changes requested):
Great initiative :+1:

AnizR:
@acsonefho ping

account_payment_method_or_mode_fs_storage/models/fs_storage.py
```diff
@@ -12,6 +12,9 @@ class FsStorage(models.Model):
     use_on_payment_method = fields.Boolean(
         help="Tells if storage can be set on payment methods",
     )
+    use_on_payment_mode = fields.Boolean(
```
acsonefho: As discusses, IMO it's not a good idea to always adding a boolean on the storage itself to know where it should be available.
The idea is to add (into config) which storage are available for payment mode (and same for payment method).
The `use_on_payment_mode` shouldn't be an attribute of the storage.

account_payment_method_or_mode_fs_storage/models/account_payment_mode.py
```diff
@@ -0,0 +1,25 @@
...
+        help="Storage where to put the file after generation",
+    )
+
+    def _get_selection_storage(self):
+        """
+        This is to avoid giving access to the model fs.storage
+        at groups other than base.group_system
+        """
+        storages = (
+            self.env["fs.storage"].sudo().search([("use_on_payment_mode", "=", True)])
```
acsonefho: I suppose this should be updated with my previous comment

account_payment_method_or_mode_fs_storage/models/account_payment_order.py
```diff
@@ -12,12 +12,17 @@ class AccountPaymentOrder(models.Model):
...
     def _must_be_exported_to_storage(self):
         self.ensure_one()
-        return bool(self.payment_method_id.storage)
+        return bool(self._get_storage())
 
     def _export_to_storage(self, file_content, filename):
-        storage_id = int(self.payment_method_id.storage)
+        storage_id = int(self._get_storage())
         # user that confirms order may not have access to storage
         export_storage = self.env["fs.storage"].sudo().search([("id", "=", storage_id)])
```
acsonefho: Please can you use a `browse(storage_id)`.
Both are the same but technically, the `browse` could use the existing cache/prefetch and `search` force using a `flush` (on specific columns) who is not necessary.

account_payment_method_or_mode_fs_storage/models/fs_storage.py
```diff
@@ -26,3 +29,17 @@ def _check_use_on_payment_method(self):
 
         if any(mapped_data.get(str(rec.id), 0) > 0 for rec in storage_not_on_payment):
             raise UserError(_("Storage is already used on at least one payment method"))
+
+    @api.constrains("use_on_payment_mode")
+    def _check_use_on_payment_mode(self):
+        storage_not_on_payment = self.filtered(lambda r: not r.use_on_payment_mode)
```
acsonefho: Based on my previous comment, I suppose this should be moved too.

account_payment_method_or_mode_fs_storage/models/res_company.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    fs_storage_source_payment = fields.Selection(
```
acsonefho, line 11: IMO this is strange. It depends on where the storage is filled.
That means `mode` and `method` can not be mixed?

acsonefho (changes requested):

account_payment_method_or_mode_fs_storage/migrations/16.0.1.0.4/post-migrate.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def _rename_modules(env):
+    modules = [
+        (
+            "account_payment_method_fs_storage",
+            "account_payment_method_or_mode_fs_storage",
+        )
+    ]
+    openupgrade.update_module_names(env.cr, modules, merge_modules=True)
```
acsonefho, line 13: I'm not sure about `merge_modules=True`.

account_payment_method_or_mode_fs_storage/models/account_payment_method.py
```diff
@@ -0,0 +1,44 @@
...
+                    "account_payment_method_or_mode_fs_storage"
+                    f".fs_storage_ids_{self.env.company.id}"
+                )
+            )
+            if ids:
+                storages = self.env["fs.storage"].browse(ast.literal_eval(ids))
+                return [(str(r.id), r.display_name) for r in storages]
+        return []
+
+    def write(self, values):
```
acsonefho: Useless

account_payment_method_or_mode_fs_storage/models/account_payment_order.py
```diff
@@ -12,12 +12,17 @@ class AccountPaymentOrder(models.Model):
 
     _inherit = "account.payment.order"
 
+    def _get_storage(self):
+        if self.env.company.fs_storage_source_payment == "mode":
+            return self.payment_mode_id.storage
+        return self.payment_method_id.storage
```
acsonefho: IMO you should do a `else` section.
And if doesn't fit to your case: `return ''` (in case of another module extend your `fs_storage_source_payment`.

account_payment_method_or_mode_fs_storage/models/account_payment_order.py
acsonefho: up :pray:

account_payment_method_or_mode_fs_storage/models/res_config_settings.py
```diff
@@ -0,0 +1,67 @@
...
+        string="fs storage source payment",
+        related="company_id.fs_storage_source_payment",
+        readonly=False,
+    )
+
+    fs_storage_ids = fields.Many2many("fs.storage", string="Fs Storage allowed")
+
+    @api.model
+    def get_values(self):
+        res = super(ResConfigSettings, self).get_values()
```
acsonefho: More python 3: `super().get_values()`

account_payment_method_or_mode_fs_storage/models/res_config_settings.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+        return res
+
+    @api.model
+    def set_values(self):
+        self.env["ir.config_parameter"].sudo().set_param(
+            f"account_payment_method_or_mode_fs_storage.fs_storage_ids_{self.env.company.id}",
+            self.fs_storage_ids.ids,
+        )
+        return super(ResConfigSettings, self).set_values()
```
acsonefho: More python 3: `super().set_values()`

account_payment_method_or_mode_fs_storage/readme/CONTRIBUTORS.rst
acsonefho, line 1: You're a contributor :muscle:

acsonefho (changes requested):

acsonefho (approved):

acsonefho (approved):

acsonefho (approved):

PR #1280, opened 2024-06-04
[16.0][FIX] account_banking_sepa_direct_debit: don't test payment state

chienandalu:
PR #1276 changes the rules for the move payment state affecting this module's tests. In any case, it isn't something relevant to test for this case.

cc @Tecnativa 

ping @pedrobaeza


pedrobaeza (approved):
Yes, it's a bad habit to test things that your module doesn't perform, as they are fragile in an integration scenario. Some people think that this way, when failing, you detect the exact moment where things start to go bad, but at the end, it's more code to maintain, and it's not clear that including that asserts will assure the result of your test. This example serves to see that it's a non needed check.

Fast-tracking this:

/ocabot merge patch

PR #1282, opened 2024-06-04
[15.0][IMP] account_banking_mandate_contact: Show contact mandate account bank in invoice report

carlosdauden:
@Tecnativa TT49436

### DIFF
--- a/account_banking_mandate_contact/models/account_move.py
+++ b/account_banking_mandate_contact/models/account_move.py
@@ -1,0 +1,19 @@
+# Copyright 2024 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def partner_banks_to_show(self):
+        self.ensure_one()
+        p_shipping_bank_id = self.partner_shipping_id.valid_mandate_id.partner_bank_id
+        if (
+            p_shipping_bank_id
+            and not self.partner_bank_id
+            and self.payment_mode_id.payment_method_id.code == "sepa_direct_debit"
+        ):
+            return p_shipping_bank_id
+        return super().partner_banks_to_show()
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

CarlosRoca13 (approved):

account_banking_mandate_contact/models/account_move.py
```diff
@@ -0,0 +1,19 @@
...
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def partner_banks_to_show(self):
+        self.ensure_one()
+        p_shipping_bank_id = self.partner_shipping_id.valid_mandate_id.partner_bank_id
+        if (
+            p_shipping_bank_id
+            and not self.partner_bank_id
+            and self.payment_mode_id.payment_method_id.code == "sepa_direct_debit"
```
pedrobaeza, line 16: I think that for decoupling this module from SEPA DD, we should:

```suggestion
            and not self.payment_mode_id.show_bank_account_from_journal
```

carlosdauden: With this change, what will happen when a contact has a specific mandate and an invoice is printed where, exceptionally, payment by transfer has been established?

carlosdauden: What is the problem with the existing code?
Establishing a bank mandate at the contact level cannot be too decoupled from SEPA DD.

pedrobaeza: Without having installed `account_banking_sepa_direct_debit`, this patch won't work, so that's the coupling.

I see that the same coupling is right now on `account_payment_partner`:

https://github.com/OCA/bank-payment/blob/87c513ba1142d0b2bc8eb4d0ede05787a2bf892a/account_payment_partner/models/account_move.py#L154

and this was highlighted already in this comment: https://github.com/OCA/bank-payment/pull/1219#discussion_r1490821169

I think it's the duty of the user to configure correctly the payment mode to show or not customer bank accounts. It's not like how they are forced to show it. Users have the control over this.

What do you think about changing this in both parts with a migration script setting to not show the bank account if it's set that way but not `sepa_direct_debit`? Do you want me to do it?

carlosdauden: Ok, do what you think, I just want the client to see what he expects.

PR #1286, opened 2024-06-12
[16.0][FIX] account_payment_mode: default inbound methods

benwillig:
Exclude electronic payment methods as they should be linked only to one journal.


sbidoul:
Apparently an alternative solution has been merged in https://github.com/OCA/bank-payment/pull/1329.

@benwillig can you check?

benwillig:
@sbidoul Yes indeed, I close this one

PR #1296, opened 2024-06-17
[15.0][IMP] account_payment_order: UX improvements

victoralmau:
Changes done:
- [x] Prevent Payment Date from being defined in transactions to avoid confusion.
- [x] Move Payments (from Payment line form view) to a more appropriate (and readable) place.

Please @pedrobaeza and @sergio-teruel and @carlosdauden can you review it?

@Tecnativa

### DIFF
--- a/account_payment_order/views/account_payment_line.xml
+++ b/account_payment_order/views/account_payment_line.xml
@@ -17,7 +17,12 @@
                             domain="[('reconciled','=', False), ('account_id.reconcile', '=', True)] "
                         />
                         <!-- we removed the filter on amount_to_pay, because we want to be able to select refunds -->
-                        <field name="date" />
+                        <field name="state" invisible="1" />
+                        <field
+                            name="date"
+                            attrs="{'readonly': [('state', '!=', 'confirmed')]}"
+                            force_save="1"
+                        />
                         <field name="ml_maturity_date" readonly="1" />
                         <field name="amount_currency" />
                         <field name="currency_id" />
--- a/account_payment_order/views/account_payment_line.xml
+++ b/account_payment_order/views/account_payment_line.xml
@@ -45,10 +45,14 @@
                         />
                         <field name="amount_company_currency" />
                         <field name="company_currency_id" invisible="1" />
-                        <field name="payment_ids" />
                         <field name="payment_type" invisible="1" />
                     </group>
                 </group>
+                <notebook>
+                    <page name="payments" string="Payments">
+                        <field name="payment_ids" nolabel="1" />
+                    </page>
+                </notebook>
             </form>
         </field>
     </record>


account_payment_order/views/account_payment_line.xml
```diff
@@ -17,7 +17,10 @@
                             domain="[('reconciled','=', False), ('account_id.reconcile', '=', True)] "
                         />
                         <!-- we removed the filter on amount_to_pay, because we want to be able to select refunds -->
-                        <field name="date" />
+                        <field
+                            name="date"
+                            attrs="{'readonly': [('ml_maturity_date', '!=', False)]}"
```
pedrobaeza: I think the readonly condition may not be appropriated. I think it should be readonly in any state different from confirmed. In confirmed state, you should have the option to change it.

victoralmau: In the 'Confirm Payments' process (`draft2open`) this date is re-defined https://github.com/OCA/bank-payment/pull/1296, so I think allowing to edit it would be a mistake (plus there would be discrepancy between transactions and payments that have already used that date).

pedrobaeza: No, we precisely allow to change it in confirmed state, if you don't agree with the proposed one. On draft we avoid the confusion, and in the rest of the states is too late, as the file is already generated. That's why I say readonly when state != 'confirmed'.

account_payment_order/views/account_payment_line.xml
```diff
@@ -17,7 +17,11 @@
                             domain="[('reconciled','=', False), ('account_id.reconcile', '=', True)] "
                         />
                         <!-- we removed the filter on amount_to_pay, because we want to be able to select refunds -->
-                        <field name="date" />
+                        <field name="state" invisible="1" />
+                        <field
+                            name="date"
```
pedrobaeza, line 22: Doesn't it require `force_save="1"`?

pedrobaeza (approved):

sergio-teruel (approved):

carlosdauden (approved):

account_payment_order/views/account_payment_line.xml
victoralmau: Checking this again for v16 I see that it is incorrect: the state condition is incorrect (it should be `open`) but also the transactions (`account.payment.line`) of a payment order has `readonly=True, states={"draft": [("readonly", False)]}`, so what should be the correct condition? (or maybe just `readonly=true`)

pedrobaeza: Didn't you check the value of the state? I told you blindly in pseudo-code. Let's keep it readonly always then, as we don't need to change it. It was just for giving the option.

PR #1289, opened 2024-06-12
[15.0][IMP] account_payment_order: Add Default Tranfer Move Date configuration

victoralmau:
Add Default Tranfer Move Date configuration

Options:
- **Payment date**: Set the payment date in the transfer entry (similar to the existing so far).
- **Now + Payment due date**: Set now date the payment date in the transfer entry and t he corresponding due date in journal items.

ejemplo

Please @pedrobaeza  @carlosdauden and @sergio-teruel can you review it?

@Tecnativa TT49582

PR #1297, opened 2024-06-17
[15.0][IMP] account_payment_order: Integrate Default Tranfer Move Date option (now_date_maturity)

victoralmau:
Integrate Default Tranfer Move Date option (now_date_maturity)

Related to https://github.com/OCA/bank-payment/pull/1289

In v13 the existing behavior was defined with today's date the Transfer Move, therefore, we define that behavior by default without configuration since it is the expected behavior.

https://github.com/OCA/bank-payment/pull/805/files

@Tecnativa TT49582

### DIFF
--- a/account_payment_order/models/account_payment.py
+++ b/account_payment_order/models/account_payment.py
@@ -51,13 +51,13 @@ class AccountPayment(models.Model):
 
     def _prepare_move_line_default_vals(self, write_off_line_vals=None):
         """Overwrite date_maturity of the move_lines that are generated when related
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Please check CI

pedrobaeza (approved):
Sorry for the noise. This one should be the default now that transfer moves is always the option.

victoralmau:
Ping @sergio-teruel

carlosdauden (approved):

sergio-teruel (approved):

PR #1298, opened 2024-06-18
[16.0][FIX] account_payment_mode: do not propose an electronic payment method when creating a journal if it has already been assigned

JordiBForgeFlow:
Previously, when you tried to install this module and the module payment_authorize at the same time, an error would raise because payment_authorize creates an electronic payment method, that is then assigned to the cash and bank accounts when the chart of accounts is loaded, resulting in an error.


EmilioPascual:
BP for 16.0 with fix already merged in 17.0 https://github.com/OCA/bank-payment/pull/1329

PR #1304, opened 2024-07-01
[16.0][IMP] account_payment_order: Add Transfer moves dates

victoralmau:
FWP from 15.0: https://github.com/OCA/bank-payment/pull/1289 + https://github.com/OCA/bank-payment/pull/1297

In v13 the existing behavior was defined with today's date the Transfer Move, therefore, we define that behavior by default without configuration since it is the expected behavior.

https://github.com/OCA/bank-payment/pull/805/files

@Tecnativa TT49582


pedrobaeza (approved):

sergio-teruel (approved):

SoniaViciana:
I don't understand this merge, If I want to make a order for future payments (due date), I can't because it takes today's date:

image

image

For me, the expected behavior it's that the payment to my customers corresponds to the expiration date defined in the transaction lines. In line with `date_prefered`.

pedrobaeza:
This follows the logic previous the refactoring into account.payment and depending the transfer move selection. There's anyway some missing bits that are put here: #1315. Please check it to not let half the patch.

For what you states, I think a new option should be set and it can be implemented inside the module.

PR #1308, opened 2024-07-05
[16.0][ADD] account_payment_order_filename: new module

acsonefho:
Add new module account_payment_order_filename to let the user customize the file generated. This behavior use sequence to also have the possibility to generate a unique filename if necessary


account_payment_order_filename/README.rst
```diff
@@ -0,0 +1,86 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/bank-payment-16-0/bank-payment-16-0-account_payment_order_filename
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/bank-payment&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module let the possibility to the user to specify a sequence for the filename generation on the payment order.
```
sbidoul, line 31: Could you add a CONTEXT.rst explaining when this is useful?

A question I'd ask is in which circumstance we need a different number for the files than the payment orders.

gdgellatly:
Why does this module override the specific sepa file and not `generate_payment_file`?

I took a look, it would be much easier, just depend on `account_payment_order` and adjust the super of that with the filename and then everything, not just modules depending on `pain_base` can use. The finalize sepa seems to just return to generate_payment_file anyway. I feel like I must be missing something.

sbidoul:
@acsonefho do you still need this?

acsonefho:
> @acsonefho do you still need this?

@sbidoul Yes and it's used in production

SAnnabelle (commented):
It's ok functionally. It's used in production.

sbidoul:
Good! Can you handle the review comments please?

AnizR:
> Why does this module override the specific sepa file and not `generate_payment_file`?
> 
> I took a look, it would be much easier, just depend on `account_payment_order` and adjust the super of that with the filename and then everything, not just modules depending on `pain_base` can use. The finalize sepa seems to just return to generate_payment_file anyway. I feel like I must be missing something.

I agree. 
Moreover, this module is specific to SEPA files but there isn't any mention anywhere.

I think that we have two options:
1) Rename the module to make it specific to SEPA payments.
2) Make it generic to all payment files and modify the function `generate_payment_file` as specified. 

I prefer the second option

PR #1309, opened 2024-07-08
[15.0][IMP] account_payment_order*: Define the correct date in the files that are generated.

victoralmau:
Define the correct date in the files that are generated.

Compatibility with https://github.com/OCA/bank-payment/pull/1297

The payment (`account.payment`) and the account entry (`account.move`) are defined with today's date, but the "_important_" date is the date of the line, therefore the `payment_line_date` field is created.

The modules `account_banking_sepa_sepa_direct_debit` + `account_banking_sepa_credit_transfer` to use the payment_line_date field.

The modules `l10n_es_payment_order_confirming_aef` + `l10n_es_payment_order_confirming_sabadell` of OCA/l10n-spain do not need these changes because they do not use `payment_ids`.

@Tecnativa TT49988

### DIFF
--- a/account_banking_sepa_credit_transfer/models/account_payment_order.py
+++ b/account_banking_sepa_credit_transfer/models/account_payment_order.py
@@ -83,11 +83,8 @@ class AccountPaymentOrder(models.Model):
             priority = payment_line.priority
             local_instrument = payment_line.local_instrument
             categ_purpose = payment_line.category_purpose
-            # The field line.date is the requested payment date
-            # taking into account the 'date_prefered' setting
-            # cf account_banking_payment_export/models/account_payment.py
-            # in the inherit of action_open()
-            key = (line.date, priority, local_instrument, categ_purpose)
+            # The field line.payment_line_date is the requested payment date
+            key = (line.payment_line_date, priority, local_instrument, categ_purpose)
             if key in lines_per_group:
                 lines_per_group[key].append(line)
             else:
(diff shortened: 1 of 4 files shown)


account_payment_order/models/account_payment.py
```diff
@@ -49,6 +50,14 @@ def _compute_payment_method_line_fields(self):
                 )
         return res
 
+    @api.depends("payment_line_ids", "payment_line_ids.date")
+    def _compute_payment_line_date(self):
+        for item in self:
+            if item.payment_line_ids:
+                item.payment_line_date = item.payment_line_ids[:1].date
+            else:
+                item.payment_line_date = item.payment_line_date
```
pedrobaeza: I think there's no need for the `if`:

```suggestion
        for item in self:
            item.payment_line_date = item.payment_line_ids[:1].date
```

account_banking_sepa_credit_transfer/models/account_payment_order.py
```diff
@@ -87,7 +87,7 @@ def generate_payment_file(self):  # noqa: C901
             # taking into account the 'date_prefered' setting
             # cf account_banking_payment_export/models/account_payment.py
             # in the inherit of action_open()
-            key = (line.date, priority, local_instrument, categ_purpose)
+            key = (line.payment_line_date, priority, local_instrument, categ_purpose)
```
pedrobaeza, line 87: Change the previous comment according this change

pedrobaeza (approved):

pedrobaeza:
@carlosdauden this is the reason why the previous payment date was set to the requested date, but let's have both things through this extra field in `account.payment`.

sergio-teruel (approved):

pedrobaeza:
Let's patch the problem ASAP:

/ocabot merge patch

Issue #1283, opened 2024-06-05
[16.0] account_payment_mode incompatible with payment_stripe and payment_adyen


EmilioPascual:
## Module
account_payment_mode

## Describe the bug
If you has installed module `account_payment_mode` and any of this modules `payment_stripe` or `payment_adyen`, when you install a localisation module such as `l10n_generic_coa` the following error happens
```
odoo.tools.convert.ParseError: while parsing /opt/odoo/auto/addons/l10n_generic_coa/data/l10n_generic_coa_post.xml:79
Some payment methods supposed to be unique already exists somewhere else.
([adyen] Adyen (inbound), [stripe] Stripe (inbound))
View error context:
'-no context-'
```

## To Reproduce
**Affected versions**:

1. In a new db.
2. Install `account_payment_mode`
3. Install `payment_stripe` or `payment_adyen`
4. Try install `l10n_generic_coa`

**Expected behavior**
Having a database with modules installed in that order

**Additional context**
Issue closed: https://github.com/OCA/bank-payment/issues/977
PR https://github.com/OCA/bank-payment/pull/1093 with fix but it didn't solve it

@moduon MT-6161

yajo:
Maybe the fix is to backport https://github.com/OCA/bank-payment/pull/1312?

EmilioPascual:
Yes, I did the backport in https://github.com/OCA/bank-payment/pull/1329 a months ago and I forgot close the issue.

PR #1312, opened 2024-07-08
[17.0][FIX] account_payment_mode: error when payment mode

imlopes:
When installing `account_payment_mode` and a provider together, we have an error linked to a duplicate payment method : 

```
  File "/odoo/src/addons/account/models/account_payment_method.py", line 141, in _ensure_unique_name_for_journal
    self.journal_id._check_payment_method_line_ids_multiplicity()
  File "/odoo/src/addons/account/models/account_journal.py", line 540, in _check_payment_method_line_ids_multiplicity
    raise ValidationError(_(
odoo.exceptions.ValidationError: Some payment methods supposed to be unique already exists somewhere else.
(Stripe)
```

https://github.com/OCA/bank-payment/issues/977

This error is linked to `account_payment_mode` and it was considered fixed here: https://github.com/OCA/bank-payment/pull/1093 but the problem is still there: https://github.com/OCA/bank-payment/issues/1283

But now odoo is checking by `account.payment.method` mode `unique` or `electronic`, so both need to be "unique".

https://github.com/odoo/odoo/commit/d07a6969903705f5676cc3052c3fba78bdf76f45

So here I'm adapting the code to follow this logic.

I've found this PR for v16.0 (https://github.com/OCA/bank-payment/pull/1298) and I really don't know the best solution.

### DIFF
--- a/account_payment_mode/models/account_journal.py
+++ b/account_payment_mode/models/account_journal.py
@@ -19,8 +19,13 @@ class AccountJournal(models.Model):
         method_info = self.env[
             "account.payment.method"
         ]._get_payment_method_information()
+        allowed_modes = ["unique"]
+        if "payment_provider_id" in self.env["account.payment.method.line"]._fields:
+            allowed_modes.append("electronic")
         unique_codes = tuple(
-            code for code, info in method_info.items() if info.get("mode") == "unique"
+            code
+            for code, info in method_info.items()
+            if info.get("mode") in allowed_modes
         )
         all_in = self.env["account.payment.method"].search(
             [


pedrobaeza:
Please give more context.

imlopes:
> Please give more context.

Hello @pedrobaeza 
I've updated the description for this PR.

Thanks

cyrilmanuel (approved):

yankinmax (approved):
LGTM code review

account_payment_mode/models/account_journal.py
```diff
@@ -19,8 +19,15 @@ def _default_inbound_payment_methods(self):
...
+        manage_providers = (
+            "payment_provider_id" in self.env["account.payment.method.line"]._fields
+        )
         unique_codes = tuple(
-            code for code, info in method_info.items() if info.get("mode") == "unique"
+            code
+            for code, info in method_info.items()
+            if info.get("mode") in ("unique", "electronic")
+            and manage_providers
+            or info.get("mode") == "unique"
```
pedrobaeza: Don't you need parenthesis in this second condition?

imlopes: Hello @pedrobaeza 
I don't think so. Can you explain your point here ?

pedrobaeza: OK, I see, but let's make the code more readable:

```python
allowed_modes = ["unique"]
if "payment_provider_id" in self.env["account.payment.method.line"]._fields:
    allowed_modes.append("electronic")
unique_codes = tuple(
    code for code, info in method_info.items() if info.get("mode") in allowed_modes
)
```

pedrobaeza (approved):
/ocabot merge patch

PR #1314, opened 2024-07-09
[16.0][FIX] account_payment_order: add missing option "structured" to payment line reference type

sbejaoui:
The payment line reference type takes its value from the invoice reference type, and both fields must have the same options.
The missing option "structured" causes a bug during payment creation.

### DIFF
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -77,6 +77,7 @@ class TestPaymentOrderInboundBase(AccountTestInvoicingCommon):
                 invoice_line_form.price_unit = 100.0
                 invoice_line_form.account_id = self.invoice_line_account
                 invoice_line_form.tax_ids.clear()
+        invoice_form.reference_type = "structured"
         invoice = invoice_form.save()
         invoice_form = Form(invoice)
         invoice_form.payment_mode_id = self.inbound_mode
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -51,9 +51,7 @@ class AccountMove(models.Model):
         """Retrieve the communication string for this direct item."""
         communication = self.payment_reference or self.ref or self.name
         if self.is_invoice():
-            if (self.reference_type or "none") != "none":
-                communication = self.ref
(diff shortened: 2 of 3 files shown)


flotho:
Hi,
Thanks for this PR.
Could you change the commit and PR name to follwo the guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#git

Regards

ThomasBinsfeld (approved):

rousseldenis:
@sbejaoui

sbejaoui:
> Hi, Thanks for this PR. Could you change the commit and PR name to follwo the guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#git
> 
> Regards

done

rousseldenis (approved):

flotho (approved):
Thnaks

account_payment_order/tests/test_payment_order_inbound.py
```diff
@@ -77,6 +77,7 @@ def _create_customer_invoice(self):
                 invoice_line_form.price_unit = 100.0
                 invoice_line_form.account_id = self.invoice_line_account
                 invoice_line_form.tax_ids.clear()
+        invoice_form.reference_type = "structured"
```
pedrobaeza, line 80: The fact that you need to change this means that there's a breaking change. Can you please avoid it?

pedrobaeza (changes requested):

account_payment_order/tests/test_payment_order_inbound.py
sbejaoui: I made this change in a separate commit to demonstrate how to reproduce the bug, but it is not necessary.

pedrobaeza: OK, understood. Then just squash both together and indicate in the extended message commit that the test is modified to be able to reproduce the problem.

sbejaoui: done

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to 17

PR #1320, opened 2024-07-16
[14.0][FIX] account_payment_order: out of index date

chienandalu:
If we had a payment order with no lines we'll be getting an index error. Better to avoid it having an empty recordset.

cc @Tecnativa 

please check @pedrobaeza


pedrobaeza (approved):
/ocabot merge patch

PR #1323, opened 2024-07-25
[15.0][IMP] account_banking_mandate: Set sequence when creating a mandate with empty mandate reference

victoralmau:
Set sequence when creating a mandate with empty mandate reference

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT50317

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -176,8 +176,7 @@ class AccountBankingMandate(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
-            unique_mandate_reference = vals.get("unique_mandate_reference", "/")
-            if unique_mandate_reference == "/":
+            if (vals.get("unique_mandate_reference") or "/") == "/":
                 vals["unique_mandate_reference"] = (
                     self.env["ir.sequence"].next_by_code("account.banking.mandate")
                     or "New"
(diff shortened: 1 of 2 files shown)


account_banking_mandate/models/account_banking_mandate.py
```diff
@@ -176,8 +176,10 @@ def _check_valid_state(self):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
-            unique_mandate_reference = vals.get("unique_mandate_reference", "/")
-            if unique_mandate_reference == "/":
+            if (
+                not vals.get("unique_mandate_reference")
+                or vals.get("unique_mandate_reference") == "/"
+            ):
```
pedrobaeza:
```suggestion
            if vals.get("unique_mandate_reference", "/") == "/":
```

victoralmau: Sorry, but that's how it was previously and it didn't cover the use case of leaving the reference empty.

pedrobaeza: Ah, OK, that's right. Then it should be:

```suggestion
            if (vals.get("unique_mandate_reference") or "/") == "/":
```

pedrobaeza (approved):

carolinafernandez-tecnativa (approved):
Functional review LGTM :)

PR #1326, opened 2024-07-26
[17.0][FIX] account_payment_purchase: copy the fields to supplier invoices.

Tisho99:
The account_payment_purchase says that the new fields (Bank Account and Payment Mode) are copied from partner to purchase order, but they are not.

This PR fixes it, but respecting the constrains of the compute of account_payment_partner, the payment mode has to require a bank account to copy the bank account. (Related with #1303)

I have added this constraint to the compute of the bank_account_id of the purchase, but it can be manually edited.

I have also adapted the _onchange_purchase_auto_complete to use the new function.

T-6102

### DIFF
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -10,7 +10,7 @@ class PurchaseOrder(models.Model):
 
     supplier_partner_bank_id = fields.Many2one(
         comodel_name="res.partner.bank",
-        compute="_compute_payment_mode",
+        compute="_compute_supplier_partner_bank",
         readonly=False,
         store=True,
         precompute=True,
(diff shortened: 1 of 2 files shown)


Tisho99:
@HaraldPanten @manuelregidor

Tisho99:
Hi @victoralmau 

This PR overrides a function you created last month (#1303), but it respects the account_payment_partner compute and your test. Maybe you are interested

A review would also be appreciated

Thank you!

account_payment_purchase/models/purchase_order.py
```diff
@@ -46,18 +46,36 @@ def _get_default_supplier_partner_bank(self, partner):
...
-        """Leave the bank account empty so that account_payment_partner set the
-        correct value with compute."""
-        invoice_vals = super()._prepare_invoice()
-        invoice_vals.pop("partner_bank_id")
-        return invoice_vals
+        res = super()._prepare_invoice()
+        if self.env.context.get("skip_payment_mode_propagation"):
+            return res
+        payment_mode_id = (
+            self.payment_mode_id or self.partner_id.supplier_payment_mode_id
```
manuelregidor: Is it correct that in case there's no a payment mode set in the purchase order the invoice is created with the payment mode set in the vendor? Shouldn't it be left empty in the invoice too so the information matches?

Tisho99: You are right, changed!

manuelregidor (approved):
Technical review. LGTM

HaraldPanten:
@Tisho99 You commented that this PR is fixing a functionallity of the module. Souldn't it be a FIX, instead of an IMP, then?

HaraldPanten:
@Tisho99 What about the commit message?

Tisho99:
> @Tisho99 What about the commit message?

Updated

HaraldPanten (approved):
@pedrobaeza Doesn't it make sense to you?

THX

account_payment_purchase/models/purchase_order.py
```diff
@@ -46,18 +46,31 @@ def _get_default_supplier_partner_bank(self, partner):
     def _compute_payment_mode(self):
         for order in self:
             if order.partner_id:
+                order = order.with_company(order.company_id)
+                order.payment_mode_id = order.partner_id.supplier_payment_mode_id
+            else:
+                order.payment_mode_id = False
```
pedrobaeza: This part should go out. The payment mode should remain to the existing one.

account_payment_purchase/models/purchase_order.py
```diff
@@ -46,18 +46,31 @@ def _get_default_supplier_partner_bank(self, partner):
...
-        invoice_vals.pop("partner_bank_id")
-        return invoice_vals
+        res = super()._prepare_invoice()
+        vals = {"payment_mode_id": self.payment_mode_id.id}
+        if self.payment_mode_id.payment_method_id.bank_account_required:
+            vals["partner_bank_id"] = self.supplier_partner_bank_id.id
+        else:
+            res.pop("partner_bank_id")
+        res.update(vals)
+        return res
```
pedrobaeza: Please avoid the use of an intermediate dictionary:

```suggestion
        res = super()._prepare_invoice()
        res["payment_mode_id"]: self.payment_mode_id.id
        res.pop("partner_bank_id", False)
        if self.payment_mode_id.payment_method_id.bank_account_required:
            res["partner_bank_id"] = self.supplier_partner_bank_id.id
        return res
```

pedrobaeza (changes requested):

account_payment_purchase/models/purchase_order.py
Tisho99: Okay, the module already sets it to False, but i think you are right. i'll change it

Tisho99:
Hi @pedrobaeza 

I applied your requested changes

Thank you for the review!

pedrobaeza (approved):
/ocabot merge patch

PR #1315, opened 2024-07-10
[16.0][IMP] account_payment_order*: Define the correct date in the files that are generated

victoralmau:
FWP from 15.0: https://github.com/OCA/bank-payment/pull/1309

Define the correct date in the files that are generated.

Compatibility with https://github.com/OCA/bank-payment/pull/1304

The payment (`account.payment`) and the account entry (`account.move`) are defined with today's date, but the "_important_" date is the date of the line, therefore the `payment_line_date` field is created.

The modules `account_banking_sepa_sepa_direct_debit` + `account_banking_sepa_credit_transfer` to use the payment_line_date field.

The modules `l10n_es_payment_order_confirming_aef` + `l10n_es_payment_order_confirming_sabadell` of OCA/l10n-spain do not need these changes because they do not use `payment_ids`.

Please @pedrobaeza can you review it?

@Tecnativa TT49988

PR #1327, opened 2024-07-29
[16.0][FIX] account_payment_order: Fix inconsistency between payment line and account payment date

sbejaoui:
The payment line date contains the scheduled payment execution date, but the account payment is created with the current date as the payment date. This causes the execution date (ReqdColltnDt) in the XML file to be incorrect.

This PR fixes the account payment date by setting it to the scheduled date from the payment line.

image

### DIFF
--- a/account_banking_mandate_contact/tests/test_account_payment_order.py
+++ b/account_banking_mandate_contact/tests/test_account_payment_order.py
@@ -96,7 +96,9 @@ class TestAccountPaymentOrder(TransactionCase):
     def test_invoice_payment_mode(self):
         self.assertEqual(self.invoice.state, "posted")
         self.assertEqual(self.invoice.payment_mode_id, self.payment_core)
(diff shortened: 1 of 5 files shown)


pedrobaeza:
This was done on purpose, as the transfer journal entry must follow the accrual principle, not to put a date in the past or in the future. I see now that you have restricted the change only to past, isn't it? Anyway, I think that if the problem is `ReqdColltnDt` of the XML, then what should be changed is that filling, not the move date.

sbejaoui:
> This was done on purpose, as the transfer journal entry must follow the accrual principle, not to put a date in the past or in the future. I see now that you have restricted the change only to past, isn't it? Anyway, I think that if the problem is `ReqdColltnDt` of the XML, then what should be changed is that filling, not the move date.

Thank you, in that way it was fixed in #1315

I close this

PR #1311, opened 2024-07-08
[14.0][IMP] account_payment_order*: Define the correct date in the files that are generated.

victoralmau:
Backport from 15.0: https://github.com/OCA/bank-payment/pull/1309

Define the correct date in the files that are generated.

Compatibility with https://github.com/OCA/bank-payment/pull/1305

The payment (`account.payment`) and the account entry (`account.move`) are defined with today's date, but the "_important_" date is the date of the line, therefore the `payment_line_date` field is created.

The modules `account_banking_sepa_sepa_direct_debit` + `account_banking_sepa_credit_transfer` to use the payment_line_date field.

The modules `l10n_es_payment_order_confirming_aef` + `l10n_es_payment_order_confirming_sabadell` of OCA/l10n-spain do not need these changes because they do not use `payment_ids`.

Please @pedrobaeza can you review it?

@Tecnativa TT49988

PR #1328, opened 2024-08-05
[14.0][IMP] account_payment_order: Fix dates in payments

Raul-S73:
The date of payments is corrected because the date on which the payment order is confirmed is currently established
This means that payments are not grouped correctly.

Before:
image
image

With this changes:
image
image

### DIFF
--- a/account_payment_order/models/account_payment.py
+++ b/account_payment_order/models/account_payment.py
@@ -50,16 +50,3 @@ class AccountPayment(models.Model):
     def _compute_payment_line_date(self):
         for item in self:
             item.payment_line_date = item.payment_line_ids[:1].date
-
-    def _prepare_move_line_default_vals(self, write_off_line_vals=None):
-        """Overwrite date_maturity of the move_lines that are generated when related
-        to a payment order.
-        """
-        vals_list = super()._prepare_move_line_default_vals(
-            write_off_line_vals=write_off_line_vals
-        )
-        if not self.payment_order_id:
-            return vals_list
-        for vals in vals_list:
-            vals["date_maturity"] = self.payment_line_ids[:1].date
-        return vals_list
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -187,7 +187,7 @@ class AccountPaymentLine(models.Model):
             "destination_account_id": self.move_line_id.account_id.id,
             "company_id": self.order_id.company_id.id,
             "amount": sum(self.mapped("amount_currency")),
-            "date": fields.Date.today(),
+            "date": self[:1].date,
             "currency_id": self.currency_id.id,
             "ref": self.order_id.name,
             "payment_reference": " - ".join([line.communication for line in self]),
--- a/account_payment_order/tests/test_payment_order_inbound.py
+++ b/account_payment_order/tests/test_payment_order_inbound.py
@@ -145,14 +145,11 @@ class TestPaymentOrderInbound(TestPaymentOrderInboundBase):
         self.assertEqual(len(self.payment_order_obj.search(self.domain)), 0)
 
     @freeze_time("2024-04-01")
-    def test_creation_transfer_move_date_01(self):
+    def test_creation_transfer_move_date(self):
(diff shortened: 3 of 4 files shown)


pedrobaeza (changes requested):
This was changed on purpose on #1305 and #1311

Raul-S73:
@pedrobaeza
How are payments going to be grouped if the due date is always the date on which the collection order is confirmed?
If I want to have payments grouped by due date, how do I do it?

pedrobaeza:
You should change the grouping criteria in the other module then.

pedrobaeza:
Anyway, if I don't remember bad, the *_grouped_output module groups everything together without more criteria.

Raul-S73:
@pedrobaeza 

Hello again, 
I have done this PR due to the needs of our clients.
When they place the payment order until now they were grouped by the due date, this way they had everything more under control. Now there is no possibility to group that way. 

With these changes, it returns to grouping by due date, making the groupings in the payment order.

pedrobaeza (changes requested):
It was incorrect before, not following the accrual principle. If your customer wants such grouping, I think you must change it at the grouping module, or in your custom.

gdgellatly:
@pedrobaeza @Raul-S73 This is also a big problem for us. With fixed execution date even, you get the date of confirmation and not the fixed date you proposed. It really is not intuitive, but seeing it won't be changed we will just put in own module.

pedrobaeza:
As said, that way of working is not correct in accounting terms. It's just like Odoo, that worked a lot of time wrongly posting entries directly in the bank account when you register a payment, and now it works correctly.

gdgellatly:
cmon @pedrobaeza I am a chartered accountant, so I don't need telling what is correct in accounting terms. It is absolutely correct to show a payment at the date it is paid. whether that is tomorrow or in 6 months time. It is extremely common and every accountant knows this.

pedrobaeza:
The question is that this is not a payment, it's the promise of a future payment. That's why the journal entry date + due date has been adjusted this way.

gdgellatly:
Just to follow this up by way of example. I have a balance date of 31 December, and a fixed payment going on 1st Jan. By you argument, the financial year it should appear is whether I do it on the 31st or 1st? complete nonsense to say that a confirmation date is an accounting date.

pedrobaeza:
On the 31st of December, you have the balance in the outstanding receipt. On the 1st of January, when coming to your bank, you neutralize the outstanding receipt, and have it on the bank account.

What is not correct is to have all the outstanding receipt balance, its neutralization, and the bank balance on the 1st of January.

gdgellatly:
> On the 31st of December, you have the balance in the outstanding receipt. On the 1st of January, when coming to your bank, you neutralize the outstanding receipt, and have it on the bank account.
> 
> What is not correct is to have all the outstanding receipt balance, its neutralization, and the bank balance on the 1st of January.

It is a January payment not a december payment, anyway I can't be bothered arguing standard accounting practice going back to the days of postdated cheques. You are right, we will just be wrong somewhere else.

pedrobaeza:
Well, I'm trying to understand any use case outside of the common ones to check for inconsistencies. Till now, we haven't found any. I don't get if your last comment is irony, or I'm missing some detail being English my foreign language.

gdgellatly:
It just means it is not a fight worth having when I can add to a custom module in 5 minutes. The use case is simple. It is not a promise of a future payment, it is a plan to pay on a certain date. Now I might understand if we were talking about customer payments, but if you are trying to do any cashflow planning even then I think it is unwise to not record at the correct date, but for suppliers no such promise exists. It is simply a future dated transaction.

pedrobaeza:
Well, I think any cashflow forecasting should be done using the due date, not the accounting date. It's the same case as with invoices: you have the entry date on the day you post the invoice, but the due date specifies when it should paid/debt.

gdgellatly:
As I say, I completely disagree, the issue was raised by internal audit, a team of 4 accountants prior to external audit.  A fixed execution date is the payment date. But not worth the fight, we fixed it where it needs fixing and everyone can be happy.

PR #1329, opened 2024-08-06
[16.0][FIX] account_payment_mode: error when payment mode

EmilioPascual:
Back-port of the fix introduced in https://github.com/OCA/bank-payment/pull/1312

@moduon MT-5800

### DIFF
--- a/account_payment_mode/models/account_journal.py
+++ b/account_payment_mode/models/account_journal.py
@@ -19,8 +19,13 @@ class AccountJournal(models.Model):
         method_info = self.env[
             "account.payment.method"
         ]._get_payment_method_information()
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Maybe the test in #1298 can be reused here if it triggers the problem?

EmilioPascual:
> Maybe the test in #1298 can be reused here if it triggers the problem?

I think this PR not fixes problem in https://github.com/OCA/bank-payment/pull/1298

rafaelbn (approved):
👍🏼 

Thank you! ❤️

PR #1331, opened 2024-08-14
[16.0][FIX/IMP] account_payment_order: use payment_reference if present for out payments

sbidoul:
When a payment reference (field `payment_reference`) is provided on the vendor bill, it should be used in priority over the vendor bill number (field `ref`).

One reason is that the `ref` field is normally different for each invoice of the same supplier (it is used in Odoo's standard duplicate warning), but the payment reference maybe the same for all payments to the same supplier. For instance some suppliers request that the customer id is used on the payment communication and it is the same on all their invoices.

Another reason is that some suppliers use a structured payment communication scheme, and that one only makes sense in the payment_reference field.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -52,7 +52,7 @@ class AccountMove(models.Model):
         communication = self.payment_reference or self.ref or self.name
         if self.is_invoice():
             if self.is_purchase_document():
-                communication = self.ref or self.payment_reference
+                communication = self.payment_reference or self.ref
             else:
                 communication = self.payment_reference or self.name
         return communication or ""
--- a/account_payment_order/tests/test_payment_order_outbound.py
+++ b/account_payment_order/tests/test_payment_order_outbound.py
@@ -304,9 +304,12 @@ class TestPaymentOrderOutbound(TestPaymentOrderOutboundBase):
         )
 
     def test_invoice_communication_02(self):
-        self.invoice.payment_reference = "R1234"
         self.assertEqual(
             "F1242", self.invoice._get_payment_order_communication_direct()
+        )
+        self.invoice.payment_reference = "R1234"
+        self.assertEqual(
+            "R1234", self.invoice._get_payment_order_communication_direct()
         )
 
     def test_invoice_communication_03(self):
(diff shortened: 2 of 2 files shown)


sbidoul:
I don't quite understand why the `payment_reference` field was not used in priority for vendor bills.

So this is a change, as the modified tests show.

@pedrobaeza @etobella @sbejaoui any opinion on this?

account_payment_order/models/account_move.py
```diff
@@ -52,7 +52,11 @@ def _get_payment_order_communication_direct(self):
...
             if self.is_purchase_document():
-                communication = self.ref or self.payment_reference
+                if self.move_type in self.get_outbound_types(include_receipts=False):
+                    # Vendor bill
+                    communication = self.payment_reference or self.ref
+                else:
+                    communication = self.ref or self.payment_reference
             else:
                 communication = self.payment_reference or self.name
         return communication or ""
```
sbejaoui: It makes sense to me that the payment reference should have priority in all cases, so there's no need for a special case for the in_refund type.

```python

def _get_payment_order_communication_direct(self):
    """Retrieve the communication string for this direct item."""
    communication = self.payment_reference
    if self.is_invoice():
        if self.is_purchase_document():
            communication = communication or self.ref
        else:
            communication = communication or self.name
    return communication or ""

```

sbejaoui (changes requested):

account_payment_order/models/account_move.py
sbidoul: I agree and that's what I would prefer, but this breaks another test related to refunds. And since I don't understand why it was done that way I'd like to have opinions of others.

pedrobaeza:
No problem on my side with this change.

account_payment_order/models/account_move.py
etobella: I think we should follow @sbejaoui comment.

I don't know why it was done this way :thinking:

pedrobaeza: Wait, wait, please let the customer part. It was done on purpose, when Odoo decided to add in the payment reference a number suffix to the invoice number, so you got an unknown number. For example, you had as invoice number "FV/2024/0001". When posting, Odoo put as payment reference "FV/2024/0001/24".

Fortunately, this is not the case anymore (or at least it seems so), but that field is used a lot by users to fill other information, like some customer references, so changing it will break current flows.

sbejaoui (approved):

sbidoul:
I have updated the PR to change de communication for supplier invoices and refunds and left the customer part as is.

pedrobaeza:
Please squash both commits together. They are the same change and the same explanation.

sbidoul:
Squashed

pedrobaeza (approved):
/ocabot merge patch

PR #1332, opened 2024-08-14
[16.0] reference_type usability improvements

sbidoul:
This PR makes the Reference Type
- modifiable, included on posted invoices, because the standard payment reference field is modifiable on posted invoices too
- visible on all bill types, because we want to be able to use structured communications on vendor bills (same visibility as the standard payment_reference field)

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -25,6 +25,7 @@ class AccountMove(models.Model):
         selection=[("none", "Free Reference"), ("structured", "Structured Reference")],
         states={"draft": [("readonly", False)]},
         default="none",
+        string="Payment Reference Type",
     )
     payment_line_count = fields.Integer(compute="_compute_payment_line_count")
 
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -23,7 +23,6 @@ class AccountMove(models.Model):
     # payment mode or company level
     reference_type = fields.Selection(
         selection=[("none", "Free Reference"), ("structured", "Structured Reference")],
-        readonly=True,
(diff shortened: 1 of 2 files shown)


sbejaoui:
for the vendor bill case, the communication is provided by the supplier (structured or not), and the user only needs to fill in the given reference in the payment_reference field. There's no need for them to specify its type.

account_payment_order/models/account_move.py
```diff
@@ -23,9 +23,9 @@ class AccountMove(models.Model):
     # payment mode or company level
     reference_type = fields.Selection(
         selection=[("none", "Free Reference"), ("structured", "Structured Reference")],
-        readonly=True,
```
sbejaoui, line 26: Consider adding `reference_type` to the `_compute_payment_reference` dependencies. We didn't need it before because the field was readonly after the invoice was posted. Now, since it can change for posted invoices, we need to trigger the payment reference recompute.

sbejaoui (changes requested):

sbidoul:
> for the vendor bill case, the communication is provided by the supplier (structured or not), and the user only needs to fill in the given reference in the payment_reference field. There's no need for them to specify its type.

Hmm, but in Belgium (at least, but in the SEPA credit transfer format too), there is a concept of structured communication for outbound payments. So don't we need a way to specify that the communication is in a structured format that needs to be validated? This relates to https://github.com/OCA/l10n-belgium/pull/204

alexis-via:
@sbidoul I would love to discuss more about the business needs for structured communication for supplier invoices and customer invoices during the next OCA code sprint in Liège. I don't have a clear understanding of the details because it is not used in France. I'm sure we can both improve the usability and improve/simplify/cleanup the code!

PR #1333, opened 2024-08-16
[IMP] account_payment_order: add button to view business document

sbidoul:
Quick access to business document on payment lines.

### DIFF
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -244,3 +244,8 @@ class AccountPaymentLine(models.Model):
         if transfer_journal:
             vals["journal_id"] = transfer_journal.id
         return vals
+
+    def action_open_business_doc(self):
+        if not self.move_line_id:
+            return False
+        return self.move_line_id.action_open_business_doc()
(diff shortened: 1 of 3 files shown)


sbidoul:
I've been wanting to do this for a long time.

Very useful for people who need to validate payment orders that have been prepared by others.

image

pedrobaeza (approved):

sbidoul:
@SAnnabelle what do you think of this one?

AnizR (approved):
Code LGTM, great new feature!

SAnnabelle:
Functionally, I think it would be better to have a smart button to go to all the related business documents. 

Because if people want to check all the invoices, with the current development, they have to click on view and then go back to the payment order to see the next invoice. 

A smart button would make it possible to have all the invoices and simply navigate from one invoice to another using the arrows. The user could open this smart button on one screen and work on the payment order on another screen. This would make it easy to make any changes to the payment order.

adrienpeiffer:
On the other hand, it's great to be able to open the document linked to a payment line directly without searching in many invoices. The best thing would be to combine the two solutions to answer different use cases. Add a smartbutton in addition to what's set up here. Let's merge this one as is for the moment.

adrienpeiffer (approved):

alexis-via:
FYI, the smart button is implemented in OCA/bank-payment-alternative v18

PR #1338, opened 2024-08-26
[14.0][ADD] account_internal_transfer: add new module

WesleyOliveira98:
New module for internal transfers compatible with payment orders

Set up a payable and a receivable account in settings
image

- Create or use a general journal for transfers
- Create or use bank journals with their bank accounts filled
- Fill in the amount, date and due date fields
image

The Journal Entry was created
image

And now, can be used in a payment order
image

cc: @marcelsavegnago @kaynnan @douglascstd


rvalyi:
Great @WesleyOliveira98 ! 

I cannot look at the PR in detail now, but at the same time I would like to make sure you guys know something that might be a bit similar that has been done in Odoo v18 so may be it gives you idea how to get it right:

- v18 PR to **share an account between companies**: https://github.com/odoo/odoo/pull/171079
- the demo https://x.com/niyasraphy/status/1824136486956024047

I don't know if you are also trying here to help transfer money between different companies during the reconciliation process (like something can be bought for company A or its branch B using the same bank account which is common in Brazil), but this is indeed a big deal in Brazil where branches use to share some accounts...

cc @renatonlima @mbcosta @alexis-via @sebastienbeau @antoniospneto @felipemotter @pedrobaeza @etobella @douglascstd

WesleyOliveira98:
> but at the same time I would like to make sure you guys know something that might be a bit similar that has been done in Odoo v18 so may be it gives you idea how to get it right:

Pretty cool this new functionality, but isn't similar to what we want to do here

WesleyOliveira98:
> I don't know if you are also trying here to help transfer money between different companies during the reconciliation process

The idea of the module is transfer money between two bank accounts of the same company.

WesleyOliveira98:
ping @OCA/banking-maintainers

kaynnan (approved):

kaynnan:
ping @OCA/banking-maintainers

PR #1346, opened 2024-09-11
[16.0][FIX] account_payment_order: Assure partner bank account in payments

pedrobaeza:
On certain v16 instances, although the partner bank account that is being pre-set in the account.payment creation values dictionary:

https://github.com/OCA/bank-payment/blob/fc7783669f40d85/account_payment_order/models/account_payment_line.py#L206

it's being marked as dirty (it's a computed writable field), and thus, being recomputed before being saved, getting an invalid value (on direct debits, the company's bank account).

The triggers that lead to this situation are unknown due to the low level where it's happening and the tons of interactions being taken in place, so the best way to deal with this is to override the compute method that computes this value, so even if the field is recomputed, it gets the correct value.

Let's be pragmatic...

@Tecnativa TT50804

### DIFF
--- a/account_payment_order/models/account_payment.py
+++ b/account_payment_order/models/account_payment.py
@@ -47,6 +47,15 @@ class AccountPayment(models.Model):
         for item in self:
             item.payment_line_date = item.payment_line_ids[:1].date
 
+    @api.depends("payment_line_ids")
+    def _compute_partner_bank_id(self):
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

victoralmau (approved):

aritzolea:
Just tried it and this solves the issue.

PR #1159, opened 2023-10-16
[15.0][IMP] bank-payment : test performance improvement

josep-tecnativa:
CC @Tecnativa TT45458

PR #1350, opened 2024-09-19
[16.0][FW] [15.0][IMP] bank-payment : test performance improvement

sebalix:
Port of #1159 from 15.0 to 16.0.

Commits related to `account_payment_order_return` are not included as this module is not ported on 17.0.


sebalix:
cc @josep-tecnativa @pedrobaeza

pedrobaeza (approved):
In this versions, it's possible to inherit from a base test class including that context, but let's fast-track it as is as a good base:

/ocabot merge patch

PR #1360, opened 2024-09-26
[16.0][IMP] account_payment_mode: change type of payment_mode.note field

clementelabore:
Change type of payment_mode.note from Text to Html
This to improve display on Invoice document

### DIFF
--- a/account_payment_mode/migrations/16.0.2.0.0/post-migrate.py
+++ b/account_payment_mode/migrations/16.0.2.0.0/post-migrate.py
@@ -1,0 +1,11 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Migrate note field from Text to Html"""
+    openupgrade.convert_field_to_html(
+        env.cr, "account_payment_mode", "note", "note", False, True
+    )
--- a/account_payment_mode/models/account_payment_mode.py
+++ b/account_payment_mode/models/account_payment_mode.py
@@ -67,7 +67,7 @@ class AccountPaymentMode(models.Model):
         related="payment_method_id.code", readonly=True, store=True
     )
     active = fields.Boolean(default=True)
-    note = fields.Text(translate=True)
+    note = fields.Html(translate=True)
     sequence = fields.Integer(default=10)
 
     @api.onchange("company_id")


pedrobaeza:
I see this improvement as something good, but you should provide a migration script for converting previous content to HTML. There's an openupgradelib method for that.

clementelabore:
I hope it was the right way to do it ;)

pedrobaeza:
I'm afraid not. That's a hook, not a migration script. A sample (not with the same change though): https://github.com/OCA/contract/tree/16.0/contract/migrations

clementelabore:
it should be ok when https://github.com/OCA/openupgradelib/pull/384 will be merged

pedrobaeza:
Please squash commits into one

pedrobaeza (approved):
Thanks for all the changes.

This must be merged with `major` switch in the ocabot merge command

pedrobaeza:
Rescuing this:

/ocabot merge major

pedrobaeza:
@AnizR can you please check the problem with your module `account_payment_method_fs_storage`?

https://github.com/OCA/bank-payment/actions/runs/13748693138/job/38446868447?pr=1360#step:8:438

`ValueError: External ID not found in the system: fs_storage.default_fs_storage`

I think it should be a change in the upstream dependency `fs_storage`.

AnizR:
> @AnizR can you please check the problem with your module `account_payment_method_fs_storage`?
> 
> https://github.com/OCA/bank-payment/actions/runs/13748693138/job/38446868447?pr=1360#step:8:438
> 
> `ValueError: External ID not found in the system: fs_storage.default_fs_storage`
> 
> I think it should be a change in the upstream dependency `fs_storage`.

I will investigate the issue with the `account_payment_method_fs_storage` module. However, it seems unusual that the demo data from `fs_storage `is not accessible during the test execution. Although `fs_storage` is listed as a dependency in my module, the test runs in the post-install phase, which previously worked without any issues.

pedrobaeza:
It's better to not use demo data in tests. Demo data may have also any problem being loaded.

AnizR:
> It's better to not use demo data in tests. Demo data may have also any problem being loaded.

I have done the changes: https://github.com/OCA/bank-payment/pull/1430

pedrobaeza:
Can you please fw-port it to 17? In this case, the script should check if the column is not JSONB for calling the method.

PR #1369, opened 2024-10-17
[16.0][IMP]account_banking_sepa_credit_transfer: add inheritance point to set charge bearer tag

AnizR:
This PR doesn't change the behavior of the module but adds an inheritance point.
Moreover, it makes the code a little bit less monolithic.

In some cases, I would like to have control on the tag 'charge bearer' even if it is a sepa payment.

### DIFF
--- a/account_banking_sepa_credit_transfer/models/account_payment_order.py
+++ b/account_banking_sepa_credit_transfer/models/account_payment_order.py
@@ -121,11 +121,7 @@ class AccountPaymentOrder(models.Model):
                 payment_info, "Dbtr", "B", self.company_partner_bank_id, gen_args
             )
             charge_bearer = etree.SubElement(payment_info, "ChrgBr")
-            if self.sepa:
-                charge_bearer_text = "SLEV"
-            else:
-                charge_bearer_text = self.charge_bearer
-            charge_bearer.text = charge_bearer_text
+            charge_bearer.text = self._get_charge_bearer_text()
             transactions_count_b = 0
             amount_control_sum_b = 0.0
             for line in lines:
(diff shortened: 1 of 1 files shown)


adrienpeiffer (approved):

account_banking_sepa_credit_transfer/models/account_payment_order.py
```diff
@@ -205,3 +201,9 @@ def generate_payment_file(self):  # noqa: C901
             nb_of_transactions_a.text = str(transactions_count_a)
             control_sum_a.text = "%.2f" % amount_control_sum_a
         return self.finalize_sepa_file_creation(xml_root, gen_args)
+
+    def _get_charge_bearer_text(self):
+        if self.sepa:
+            return "SLEV"
+        else:
+            return self.charge_bearer
```
pedrobaeza:
```suggestion
        return "SLEV" if self.sepa else self.charge_bearer
```

pedrobaeza (commented):
Please don't put the Odoo version in the commit message (but do on the PR title, but reversed, first the version, and then the tag).

A curiosity question: what is the use case for this?

account_banking_sepa_credit_transfer/models/account_payment_order.py
AnizR: Indeed, more pythonic :+1:

AnizR:
> Please don't put the Odoo version in the commit message (but do on the PR title, but reversed, first the version, and then the tag).

Done

> A curiosity question: what is the use case for this?

It is really 'specific' but we are creating SEPA transfers between entities of the same group of companies and therefore, they some time don't want this tag to be set or they want it to be set to "SHAR" (shared).

Thanks for your review!

pedrobaeza:
But that's something allowed by SEPA? If so, I think we should unlock the choice at this module level.

AnizR:
> But that's something allowed by SEPA? If so, I think we should unlock the choice at this module level.

As far as I know, it is not allowed by SEPA (Only the value 'SLEV' seems to be commonly accepted).

pedrobaeza (approved):

acsonefho (approved):

PR #1371, opened 2024-10-25
[14.0][FIX/IMP] account_payment_order: use payment_reference if present for out payments

FrankC013:
Supersedes #1331

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -48,7 +48,7 @@ class AccountMove(models.Model):
             if (self.reference_type or "none") != "none":
                 communication = self.ref
             elif self.is_purchase_document():
-                communication = self.ref or self.payment_reference
+                communication = self.payment_reference or self.ref or ""
             else:
                 communication = self.payment_reference or self.name
         # If we have credit note(s) - reversal_move_id is a one2many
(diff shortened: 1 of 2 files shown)


eantones (changes requested):
@FrankC013 This is 14.0 and it supersedes an already merged 16.0 PR???? It cannot supersedes if it's merged and it's another version. This seems a backport. Please check it out.

account_payment_order/models/account_move.py
```diff
@@ -48,7 +48,7 @@ def _get_payment_order_communication(self):
             if (self.reference_type or "none") != "none":
                 communication = self.ref
             elif self.is_purchase_document():
-                communication = self.ref or self.payment_reference
+                communication = self.payment_reference or self.ref or ""
```
eantones, line 51: Do you really need to add `or ""`?

eantones (changes requested):

account_payment_order/models/account_move.py
FrankC013: I added `or ""` because, in version 16, the **__get_payment_order_communication_** function is divided into two separate methods: _**_get_payment_order_communication_direct**_ and _**_get_payment_order_communication_full**_.

If I remove the `or ""`, the line `communication += " " + " ".join(references)` throws the error `unsupported operand type(s) for +=: 'bool' and 'str'`, as self.ref can be False. To avoid this issue, we can either split the function as in version 16 or handle it with `or ""` to ensure communication is always a string.

FrankC013:
Superseded by #1372

account_payment_order/models/account_move.py
eantones: Undestood, thanks @FrankC013

PR #1374, opened 2024-11-12
[18.0][MIG] account_invoice_select_for_payment: Migration to 18.0

Kimkhoi3010:
### Change in 18.0:

- I changed the widget of the selected_for_payment field from toggle_button to boolean_toggle because, starting from version 16, the toggle_button widget is no longer supported
- Since Odoo 14.0 the 'action_register_payment' method has been moved from the 'account.payment' model to 'account.move'


sebalix (commented):
I know there was none, but a test is welcome, especially to check if the payment removes the flag from the invoices.

grindtildeath (approved):
Thanks for the mig.

Can you please add a small test as suggested by @sebalix ?

Kimkhoi3010:
Hello @grindtildeath, @sebalix, I updated by your comments

account_invoice_select_for_payment/models/account_payment.py
```diff
@@ -13,5 +13,4 @@ def action_register_payment(self):
             invoices = self.env["account.move"].search(
                 [("id", "in", active_ids), ("selected_for_payment", "=", True)]
             )
-            invoices.write({"selected_for_payment": False})
-        return super().action_register_payment()
+        return invoices.write({"selected_for_payment": False})
```
TDu: Why loose the call to super and could the `invoices` variable be unassigned ?

Kimkhoi3010: Hi @TDu, I checked again and found out that this function has been moved to the `account.move` module since `Odoo 14.0`. I updated the PR. Thank for your comment.

account_invoice_select_for_payment/models/account_move.py
```diff
@@ -16,3 +16,12 @@ def action_toggle_select_for_payment(self):
             selected.write({"selected_for_payment": False})
         if unselected:
             unselected.write({"selected_for_payment": True})
+
+    def action_register_payment(self):
+        active_ids = self.env.context.get("active_ids")
+        if active_ids:
```
TDu: Now that it has been moved to `account_move.py `self` is an `account.move` recordset so the logic should be changed...
Is this covered by unit tests ?

TDu: It is covered by tests but they rely on the context, which will not work in live code...

Kimkhoi3010: Hi @TDu, 
I updated the PR. Thank for your comment.

account_invoice_select_for_payment/models/account_move.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def action_toggle_select_for_payment(self):
+        selected = self.filtered(lambda rec: rec.selected_for_payment)
+        unselected = self - selected
+        if selected:
+            selected.write({"selected_for_payment": False})
+        if unselected:
+            unselected.write({"selected_for_payment": True})
+
+    def action_register_payment(self, invoice_ids=None):
```
TDu: Does not look correct, the base declaration of this function is different...
https://github.com/odoo/odoo/blob/ca9a1157dd47927ae7750e6669b7efac1fdcd629/addons/account/models/account_move.py#L5100

sebalix (approved):
LG, thank you

florian-dacosta (approved):
I approve since it seems to match the scope of the module in v14.
But I'd like to improve this module, please check the PR here : https://github.com/Kimkhoi3010/bank-payment/pull/2
It does not block the migration of this one, if merged, I'll redirect the PR toward the OCA repo.

Improvement is about make the module compatible with other way to register payment than the `action_register_payment` wizard and some minor improvements about views.
Please have a look @sebalix @grindtildeath @TDu

PR #1375, opened 2024-11-13
[15.0][IMP] account_banking_mandate: Allow to change the bank mandate on invoices if they are not paid.

sergio-teruel:
cc @Tecnativa TT51723

Use case:
You already have invoices created and your customer wants change the account bank, so you need change the mandate in all pending invoices.

ping @carlosdauden @victoralmau


carlosdauden (approved):

victoralmau (approved):
IMO it makes sense.

pedrobaeza (approved):
Please forward port it to upper versions taking into account also payment in progress state

/ocabot merge minor

PR #1380, opened 2024-11-27
[17.0][FIX] account_payment_order: fix ``TypeError``

SilvioC2C:
Method ``_get_reconciled_invoices_partials()`` returns 2 lists:
- the first one contains triplets defined as (account.partial.reconcile record, amount, account.move.line record)
- the second one contains account.move IDs

Merging the 2 lists and cycling over them with
```
for (_x, _y, payment_move_line) in list1 + list2:
    ...
```
resulted in ``TypeError: cannot unpack non-iterable int object``.

This commit fixes this bug.

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -78,19 +78,17 @@ class AccountMove(models.Model):
             reference_moves |= self.reversal_move_id
         # Retrieve partial payments - e.g.: manual credit notes
         (
+            # List of triplets
+            # (account.partial.reconcile record, amount, account.move.line record)
             invoice_partials,
-            exchange_diff_moves,
+            # List of account.move IDs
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Maybe you can add a regression test so this doesn't happen again in another migration (or upstream change)? Being a very low method, the chances to break again are high.

SilvioC2C:
@pedrobaeza alas, I'm on a tight schedule at the moment and I don't have enough functional knowledge to replicate the issue; I'll try to do it asap though

pedrobaeza:
OK, thanks!

yankinmax (approved):

vvrossem (approved):

pedrobaeza:
Let's merge for now to fix the problem. The test may come later (but please don't forget it):

/ocabot merge patch

PR #1391, opened 2024-12-23
[IMP] account*: Remove dependencies in demo data on tests

etobella:
Right now, it depends a lot on demo data to execute tests. It is a good practice to remove all of these dependancies.

@victoralmau @rousseldenis @JordiBForgeFlow

PR #1395, opened 2025-01-07
[17.0][IMP] account_banking_sepa_credit_transfer: Use BaseCommon in tests

victoralmau:
Remove partner demo data on tests

@Tecnativa TT52526

### DIFF
--- a/account_banking_sepa_credit_transfer/tests/test_sct.py
+++ b/account_banking_sepa_credit_transfer/tests/test_sct.py
@@ -9,16 +9,14 @@ import time
 from lxml import etree
 
 from odoo.exceptions import UserError
-from odoo.tests.common import TransactionCase
 
-from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestSCT(TransactionCase):
+class TestSCT(BaseCommon):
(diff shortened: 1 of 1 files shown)


victoralmau:
Ping @pedrobaeza

pedrobaeza:
I don't get the relation of the commit message with the code.

victoralmau:
Excuse me, but what do you want explained? I think it has been explained many times why you should not use demo data, even less if they are demo data from odoo core, using them will have side effects of behaviors that you do not want.

pedrobaeza (commented):
Yeah, it's the word `project` which confused me, as I don't see anything related to project module.

And isn't there other data like bank partner accounts? @etobella did a cleaning on 16.0 that may be forward-ported.

pedrobaeza:
OK, I see that #1391 was already applied. Thus, the commit message should be changed to reflect the real change.

victoralmau:
Ok, it's already changed.

pedrobaeza:
The commit message is not reflecting the code.

victoralmau:
Ping @pedrobaeza

etobella (commented):
The message of the commit is different to the applied changes. Is it correct?

pedrobaeza (approved):
Finally adjusted.

/ocabot merge patch

PR #1401, opened 2025-01-20
[18.0][ADD] account_payment_base_oca: New base module for OCA/bank-payment

alexis-via:
Work in progress. Expect datamodel changes.


pedrobaeza:
Why this? Why didn't you answer to my comment in the issue? I don't want data model changes if they are not justified.

alexis-via:
I'm trying to adopt the new datamodel. I'm not saying it's the best choice, but I want to try and see what difficulties I face.
I now have the full stack working: account_payment_base_oca + account_payment_order + account_banking_pain_base + account_banking_sepa_credit_transfer + account_banking_mandate + account_banking_sepa_direct_debit. So I am able to make full scenarios and see the real problems I face and try to address them.
At the moment, it's too early to list them, because I'm currently performing the tests.

pedrobaeza:
Then we are going to have 2 forks. I don't want that "datamodel", as it's very bad as explained in the migration issue.

alexis-via:
I just pushed a new version of account_payment_base_oca. In this new version, the active field of account.payment.method.line (added by account_payment_base_oca) is set to False by default. So, when a account.payment.method is added, all the account.payment.method.line automatically created by Odoo are inactive. You can active the ones you need and leave the others inactive.
You still have the option bank_account_link = "variable" as in Odoo v17-. When "variable" is selected, a  M2M field alternative_journal_ids is displayed ; the possible journals are journal_id + alternative_journal_ids. journal_id acts as the default journal.
When you confirm a payment order that has a payment_method_line_id configured as "variable", Odoo will search the account.payment.method.line with the journal selected on the payment order and the same payment method and use that on the account.payment.
My first tests are positive.
I'll push the code of the other modules in a few days.

pedrobaeza:
I sincerely see it as a nonsense to have to look for "similar" things one way and the reverse because of a failed datamodel putting the data at journal level. You can't for example group by this field if you select different lines according the journal but representing the same payment mode. Let's keep our `account.payment.mode` model, and we can link the `account.payment.method.line` on the generated payment as we were doing previously.

alexis-via:
@pedrobaeza It's also a big temptation for me to keep account.payment.mode. In fact, I'm the one who introduced account.payment.mode in v9 during the Sorrento code sprint (to replace payment.mode that was dropped by Odoo SA with the removal of the module account_payment from the official addons), so it's kind of my baby and I like it.

But, in the long list of wired/sub-optimal datamodel of Odoo SA that I would love to avoid/change, this one is not a big deal. I think I found a solution that is acceptable and the balance between the advantage of being compatible with the native datamodel and the drawbacks of using the trick I propose to make the new datamodel work with bank_account_link = 'variable'  tend towards the adoption of the new datamodel I think. I'll continue my tests in the coming days to make sure I don't discover any other problems and I'll publish the upper-level modules.

account_payment_base_oca/models/account_payment_method_line.py
```diff
@@ -0,0 +1,195 @@
...
+    # generates an account.payment.method.line for each bank journal and there is no
+    # prepare method to inherit the values of the generated method lines.
+    # With default=False, Odoo will auto-generate inactive account.payment.method.line
+    # and we will only enable the one we want to use. If a payment order has a method
+    # line with bank_account_link='variable' and an alternative journal is selected,
+    # Odoo will not use the method line of the payment order to generate
+    # the account.payment but it will select the (inactive) method line
+    # linked to the chosen journal
+    # TODO default=False causes problems in tests of the account module
+    active = fields.Boolean(default=False)
```
sbidoul: Would it make sense to set active=False on create only when bank_account_link is variable?

sbidoul:
This sounds promising. I also prefer using the native `account.payment.method.line` if that is possible.

pedrobaeza:
I insist: how it would be better to assign at partner level the payment method line linked to a specific journal, seeing all of them in the dropdown (and this active=False trick is something conflicting with the way Odoo expects to work), instead of having an agnostic payment mode?

The payment method lines are not new. They are there since ages (v14). The only new thing is that there's a many2one at partner level, but that doesn't change anything. Please don't go with this, or we will have 2 `account_payment_order` modules.

pedrobaeza:
Another reason spotted now for not switching from `account.payment.mode`: the missing logic about the mode on refunding invoices.

pedrobaeza:
Another reason spotted now for not switching from `account.payment.mode`: the missing logic about the mode on refunding invoices.

pedrobaeza:
Another reason spotted now for not switching from `account.payment.mode`: the missing logic about the mode on refunding invoices.

sbidoul:
@pedrobaeza I think Alexis is doing a lot of very valuable work here. Let's wait a bit to see where this goes.

pedrobaeza:
No, if the foundations are incorrect, no more work should be put in here. I announce my intention on continuing with the old stack if you switch, and the model names should be preserved for the existing things according current establishment.

sbidoul:
Sorry I don't understand your arguments, Pedro. Odoo is evolving and it seems normal to me to explore and see if we can adapt.

pedrobaeza:
The arguments are clear of the cons of switching, so I'm just stating that I would push for the previous approach and this will conflict in the model names, and it's clear that these cons can't be bypassed without doing ugly hacks and standard code overrides.

pedrobaeza:
Note that I'm the first one embracing upstream changes when they are suitable for the goal of the modules. I did the refactoring to `account.payment` in v14, and thanks to that, we have enjoyed 5 versions without having to touch continuously the journal entry generation, delegating that to core mechanisms.

But in this case, it has no benefit to go with a simple wrong m2o extra field that they have added in the partner, and that is so wrong in the concept. And I insist that the `account.payment.method.line` is not something new. It's there since ages, and we set it correctly on the generated `account.payment`. The only change is about the partner preferred payment mode.

njeudy:
Hello all, I try to understand this part :) Is there a document that list user case for payment_mode use. This can  be a good way to check If new or old datamodel can be used and if there is a problem with ?

I think à create a document but if you have one, I will try not to start with empty blank sheet :)

pedrobaeza:
Basically, the payment mode tries to model the payment "instrument" that the customer (or you for vendors) is going to use. Let's say "Wire transfer", "Check", "SEPA Direct Debit", etc.

The new Odoo field in standard is just adding a sublevel, like SEPA DD for journal (bank) X, so the side effects are:

- You will see N "Sepa Direct Debit", as many as journals you have declared for using that payment method, being very confusing for the user.
- You may want to change to other bank for doing the operation in any moment.

That's way the other data model avoids this problem doing an aggregation before, and only "unfolding" to the corresponding bank (journal) on the final end when you do the payment/debit order (although you can constraint to one or several banks from the beginning if you want).

This PR tries to "bypass" the side effects doing that you select one of the many "SEPA Direct Debit", and it's changed "on the fly" when doing the payment order.

But keeping this so long stable (and simple) extra data model avoids all these problems, and we have also provided other useful tools around it like dynamic text shown in the invoice (for saying in which customer bank account is going to be done the DD, own bank account for wire transfer, etc), that will be lost (or redone) in case of the switch. We can hide totally this other m2o field in the partner view for not having any confusion, at that's all.

For me is a total nonsense to even consider doing this switch.

account_payment_base_oca/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    bank_account_required = fields.Boolean(
+        related="preferred_payment_method_line_id.payment_method_id.bank_account_required",
+    )
+    payment_method_code = fields.Char(
```
leemannd, line 14: This would need a computation script as the table account.move can be heavy on some projects

pedrobaeza:
We need to unblock this for having a complete 18.0, and unless other strong opinions against arise, the natural trend is to keep what we have in previous versions and discuss later any refactoring (if any). We will resume the migration work next week doing the raw migration fine-tuning the UX.

alexis-via:
Here is a video to illustrate this pull request, in particular the fact that is supports the link to bank journal = variable like in OCA/bank-payment v17-

https://youtu.be/fW8pH7niJI4

pedrobaeza:
Closing as this won't be merged here in this repository.

alexis-via:
@pedrobaeza I don't think you can decide this by yourself. You cannot decide alone Pedro ; you're not the dictator of OCA :)
It seems to me that quite a number of OCA users want to adopt the native datamodel for payment modes introduced in v18. Let's finish the debate and take a decision.

pedrobaeza:
I didn't decide. It was you [in this message](https://odoo-community.org/groups/contributors-15/contributors-2038857?mode=thread&date_begin=&date_end=) saying that the convergence is not going to happen, so with that path, this module doesn't fit here.

account_payment_base_oca/models/account_payment_method_line.py
```diff
@@ -0,0 +1,195 @@
...
+        "variable_journal_ids",
+        "payment_method_id",
+    )
+    def _check_payment_method_line(self):
+        for line in self:
+            # I cannot check line.journal_id.bank_account_id when
+            # line.bank_account_link == "fixed" because method lines are
+            # auto-created on bank journal creation and the bank account
+            # but not be configured yet.
+            if line.bank_account_link == "fixed" and not line.journal_id:
```
jbaudoux, line 120: When I enable an online payment provider, it creates multiple payment method lines. Only the last one is used and has a journal. The others seem orphan. I could not find any link to them. But as they don't have a journal, this raises this ValidationError.
I worked around by deleting the orphan lines.

alexis-via:
PR moved to https://github.com/OCA/bank-payment-alternative/pull/3

PR #1403, opened 2025-01-28
[18.0][MIG] account_payment_order

alexis-via:
There is still work to do... but you can play with it.


sbidoul:
@alexis-via do you think you could add the unmerged dependencies in test-requirements.txt so this can be tested on runboat? The syntax is explained here: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

pedrobaeza:
Merging #1440 instead

alexis-via:
PR moved to https://github.com/OCA/bank-payment-alternative/pull/5

PR #1410, opened 2025-01-30
[16.0][IMP] account_payment_purchase: Propagate PO payment mode when invoicing

aritzolea:
Comes from https://github.com/OCA/bank-payment/pull/1225

Fixes payment mode propagation from purchase order to invoice when supplier's payment mode is different to the one on the purchase order.

### DIFF
--- a/account_payment_purchase/models/purchase_order.py
+++ b/account_payment_purchase/models/purchase_order.py
@@ -56,7 +56,10 @@ class PurchaseOrder(models.Model):
 
     def _prepare_invoice(self):
         """Leave the bank account empty so that account_payment_partner set the
-        correct value with compute."""
+        correct value with compute and propagate payment mode to invoice."""
         invoice_vals = super()._prepare_invoice()
         invoice_vals.pop("partner_bank_id")
+        assign_payment_mode = self._context.get("pay_mode_propagate", True)
+        if assign_payment_mode:
+            invoice_vals["payment_mode_id"] = self.payment_mode_id.id
         return invoice_vals
--- a/account_payment_purchase/models/account_move.py
+++ b/account_payment_purchase/models/account_move.py
@@ -13,6 +13,7 @@ class AccountMove(models.Model):
     def _onchange_purchase_auto_complete(self):
         new_mode = self.purchase_id.payment_mode_id.id or False
         new_bank = self.purchase_id.supplier_partner_bank_id.id or False
+        self = self.with_context(pay_mode_propagate=False)
         res = super()._onchange_purchase_auto_complete() or {}
         if self.payment_mode_id and new_mode and self.payment_mode_id.id != new_mode:
             res["warning"] = {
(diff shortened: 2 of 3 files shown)


ljsalvatierra-factorlibre (approved):

aritzolea:
Could you check this @pedrobaeza? Thanks in advance

account_payment_purchase/models/purchase_order.py
```diff
@@ -56,7 +56,10 @@ def _compute_payment_mode(self):
 
     def _prepare_invoice(self):
         """Leave the bank account empty so that account_payment_partner set the
-        correct value with compute."""
+        correct value with compute and propagate payment mode to invoice."""
         invoice_vals = super()._prepare_invoice()
         invoice_vals.pop("partner_bank_id")
+        avoid_assign_pay_mode = self._context.get("avoid_pay_mode_propagate", False)
+        if not avoid_assign_pay_mode:
```
pedrobaeza: Instead of having `not` and a "negative" variable, better to invert it (and the same for the rest):

```suggestion
        if assign_payment_mode:
```

pedrobaeza (changes requested):
And why it was not propagated the payment mode? Please document properly the commit message and the PR description.

account_payment_purchase/models/purchase_order.py
aritzolea: Reverted the variable, now it is possitive.

aritzolea:
> And why it was not propagated the payment mode? Please document properly the commit message and the PR description.

I don't know if it was a bug or there was a specific reason for not propagating the payment mode, I could edit the commit message explaning mode is now propagated, but I don't know if that's what you mean.

aritzolea:
> > And why it was not propagated the payment mode? Please document properly the commit message and the PR description.
> 
> I don't know if it was a bug or there was a specific reason for not propagating the payment mode, I could edit the commit message explaning mode is now propagated, but I don't know if that's what you mean.

I added this commit message: 

> When invoicing a purchase order the payment mode was not being propagated prior
to this commit.

Do you think it's now OK for merge or would you add something else on commit message @pedrobaeza ?

pedrobaeza:
3 commits is too much for such little code and being about the same problem, but I still don't find in your PR description nor the commit messages the steps to reproduce the problem.

aritzolea:
> 3 commits is too much for such little code and being about the same problem, but I still don't find in your PR description nor the commit messages the steps to reproduce the problem.

Squashed commits and added co-author.

Explained on PR description and commit message the case in which this bug is fixed.

pedrobaeza:
I have checked the code, and I still don't get the problem, or at least, it shouldn't be fixed there. We have this code for that:

https://github.com/OCA/bank-payment/pull/1410/files#diff-9e60db0f5d8d1708de9fd29c9e6c32f0622a159c541173ab6c3d51ce1d8e3bcbR18-R33

Please specify complete steps to reproduce the problem.

aritzolea:
> I have checked the code, and I still don't get the problem, or at least, it shouldn't be fixed there. We have this code for that:
> 
> https://github.com/OCA/bank-payment/pull/1410/files#diff-9e60db0f5d8d1708de9fd29c9e6c32f0622a159c541173ab6c3d51ce1d8e3bcbR18-R33
> 
> Please specify complete steps to reproduce the problem.

New test code would fail without the changes in the PR

Issue #1408, opened 2025-01-30
[16.0] account_payment_partner: Don't void the bank on account move if no payment mode


rousseldenis:
If no payment mode is set on the account move, the partner bank is voided if already filled in.

https://github.com/OCA/bank-payment/pull/1094/files#diff-560b5e454acd270e25223d56b85967eaec36fe9a08eac999124990650aa32f6cR114

I think this assertion is true for computed fields not stored... :

```
Now that partner_bank_id is a computed field, we shouldn't set its value
to False if the move is not a vendor bill or doesn't have a payment mode.

```

So, I will do a fix for that

@grindtildeath

rousseldenis:
This in order to fix incompatibility with overrides on bank computation on account moves:

https://github.com/OCA/account-invoicing/actions/runs/13049589264/job/36406666815?pr=1736#step:8:211

jbaudoux:
It's the occasion to improve the tests to include the refunds:
`move_type == "out_invoice"`
should be
`move_type in ("out_invoice", "in_refund")`

same for the other direction

rousseldenis:
@yvaucher

grindtildeath:
@rousseldenis Answering here for your question on https://github.com/OCA/bank-payment/pull/1183#discussion_r1935609151

It's been some time since I did https://github.com/OCA/bank-payment/pull/1094 but IIRC the idea was not to set `partner_bank_id` if we don't have a payment mode.

But I agree if it's set manually, then the computation shouldn't remove it.

rousseldenis:
> but IIRC the idea was not to set `partner_bank_id` if we don't have a payment mode.

The problem is some other modules that don't depend on this would fill in the field.

We can agree that we should not modify the field if no payment mode is not filled in.

pedrobaeza:
Handled in #1555

PR #1412, opened 2025-01-31
[FIX] account_payment_partner: Respect the filled in value if no payment mode

rousseldenis:
Fixes: https://github.com/OCA/bank-payment/issues/1408

If no payment mode is set on invoices, we should respect what is computed.

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -95,33 +95,33 @@ class AccountMove(models.Model):
                             partner.supplier_payment_mode_id.refund_payment_mode_id
                         )
 
-    @api.depends("bank_partner_id", "payment_mode_id")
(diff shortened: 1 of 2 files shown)


grindtildeath:
@rousseldenis Should we favour #1418 first? Since there are a lot of changes in your PR, I'd prefer to go with @sbidoul 's fix first, and we can discuss another potential fix you're providing here afterwards.

rousseldenis:
> @rousseldenis Should we favour #1418 first? Since there are a lot of changes in your PR, I'd prefer to go with @sbidoul 's fix first, and we can discuss another potential fix you're providing here afterwards.

Yes of course.

cyrilmanuel:
hello @rousseldenis, many thing repose on this PR, could we make some change or fix to allow this one to be merged ?

PR #1417, opened 2025-02-10
[16.0][IMP] Allow to use the storage either from the payment mode or from the payment method

AnizR:
I reopened: #1279 that aims to push payment file to a distant storage by @lmarion-source.

With a small twist:

IMO, pushing to the storage should be done as late as possible because we may not have control over it.
Therefore, if an error occurs after sending the file we may not be able to remove the file from the storage.

I had a case where my client had an error when changing the state of a payment order to 'generated'. He tried multiple times and sent duplicate of a payment file because the error was thrown after the file's sending.


account_payment_method_or_mode_fs_storage/tests/test_account_payment_method_or_mode_fs_storage.py
```diff
@@ -103,6 +103,8 @@ def test_payment_method_fs_storage(self):
         order.draft2open()
         with self.with_custom_method():
             action = order.open2generated()
+            # commit need to be performed in order to perform the 'postcommit'
+            self.env.cr.commit()  # pylint: disable=E8102
```
acsonefho: Why not doing your own sub-savepoint?

AnizR: As discussed, adding a savepoint wasn't possible since I was commiting (savepoints are destroyed once the transaction is commited).
But, I could avoid to commit and manualy run the 'postcommit' of the cursor. 
Thanks for the suggestion!

acsonefho (approved):

Issue #1290, opened 2024-06-13
account_payment_partner: partner_bank_id should not be set to False if payment_mode is not set


PNurm:
## Module
account_payment_partner

## Describe the bug
This module as of 16.0 overrides the default behavior, and previous behavior of this module in 14.0. 
- Invoices without payment_mode_id are created without a partner_bank_id

- When creating a sales invoice through sales order without a payment_mode_id, it will create invoice without **partner_bank_id**.  This module overrides  _compute_partner_bank_id and changes the behavior so if payment_mode_id is not set, it will not assign partner_bank_id. 

- From users point of view and from configuration standpoint, users should not be forced to configure a otherwise useless payment_mode_id and set that by hand on sales order. For a lot of our customers, payment_mode_id is never used for sales, and mostly used for bills and payment orders. For sales it is unnecessary step to fill out just to get bank account to compute properly. 
 
## To Reproduce
**Affected versions**:
16.0

1. Create sales order without setting payment_mode_id
2. Create invoice from sales order, it will not have bank account, therefore printing invoice will not have necessary payment related information.

**Expected behavior**
- partner_bank_id should still be computed regardless if payment_mode_id is undefined.

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

remi-filament:
Thanks @PNurm for reporting this. I also find it weird that we reset the bank account on invoice when this module is installed for move without payment_mode_id.

There is probably a good reason for this but I cannot find which one ?

Using account_edi module with Factur-x/XRechnung CII 2.2.0 enabled, if you do not have a partner_bank_id on invoice, you get an banner stating that bank_partner_id is required for this EDI format.
Also, since https://github.com/odoo/odoo/commit/7c3e5e95e5f97a8cd3f5ef43d54bf40bb07f217b on v16 (same applies to v15+), partner_bank_id is used on invoice report to show bank account on which payment is expected.

Resetting the field to False in most cases as it is currently done in this module forces the users to configure the proper account on each invoice, when in most of the cases there is only one (either the one from company / partner if no payment_mode_id selected, or the one from account_journal if payment_mode is selected).

Maybe @alexis-via @pedrobaeza @victoralmau or @grindtildeath can clarify why it is needed to reset partner_bank_id to False ?

remi-filament:
Maybe it is expected that default Manual payment mode is created and used everywhere instead of None ?

pedrobaeza:
[...]

PR #1418, opened 2025-02-11
[FIX] account_payment_partner: keep the default value for partner_bank_id

sbidoul:
The Odoo default computation for partner_bank_id on customer invoices sets the first bank account of the company.

The override in account_payment_partner resets it to false if there is no payment mode defined.

Since for customer invoices we usually don't define a payment mode (unless when using direct debit), this partner_bank_id field is therefore emptied.

This is problematic for instance when generating UBL invoices with the official Odoo module, because the bank account is mandatory to create a valid Peppol document.

Therefore I propose to keep the default behaviour if there is no payment mode on the invoice.

### DIFF
--- a/account_payment_partner/models/account_move.py
+++ b/account_payment_partner/models/account_move.py
@@ -120,8 +120,6 @@ class AccountMove(models.Model):
                             continue
                     else:
                         move.partner_bank_id = False
-            else:
-                move.partner_bank_id = False
         return res
 
     @api.depends("line_ids.matched_credit_ids", "line_ids.matched_debit_ids")


sbidoul:
This is a more focused fix for #1408

grindtildeath (approved):

grindtildeath:
@sbidoul I forward ported this change to v17.0 and fixed the test in #1423

pedrobaeza:
The idea behind the else part is for avoiding Odoo assigning a partner bank automatically, which does harm, as the payment/debit order auto-selection mechanism won't select the more suitable one, and you can see by the red CI that this is something intended.

Tisho99:
@grindtildeath 

Yes, my module is based on the feature that an invoice does not have to have a bank account if it does not have a payment mode.

If this is going to be merged, my test and my module will need a little refactor to edit the default bank account if there is not a payment mode.

However, i think the best is waiting for the original v16 PR to be merged  and resolve Pedro's comments

grindtildeath:
> The idea behind the else part is for avoiding Odoo assigning a partner bank automatically, which does harm, as the payment/debit order auto-selection mechanism won't select the more suitable one, and you can see by the red CI that this is something intended.

grindtildeath:
@pedrobaeza 
This is actually breaking a core behaviour of Odoo. Although this might not suit your needs, it's breaking the swiss localization and we have to fix that.

IMO if someone wants to deactivate this assignation done by Odoo, which I can understand, I would suggest to add a setting (eg on the journal) so that the computation can rely on this setting to remove the automatically assigned value.

Please consider payment mode can be installed on an instance where it's not used by all the companies, in that case, it shouldn't break default behaviours as is the case with this compute.

@Tisho99 what do you think about adding such a setting?

pedrobaeza:
Please specify which is the issue in the Swiss localization for determining if there are options to solve it.

Tisho99:
@grindtildeath 

At first glance, I think that adding a configuration option is a good idea. If the default operation is left as it is now (setting the bank account to false if there is no payment method), it will not be necessary to refectorize other modules, just activate the option with Swiss accounting.

sbidoul:
@pedrobaeza this is not only Swiss accounting. See the description of this issue. The default Odoo behaviour of selecting the company bank account on customer invoices is disabled, so various standard Odoo behaviours that depend on having the bank account populated do not work anymore. Peppol invoice generation is an example. Adding QR code to the invoices is another.

My question is *why* it is necessary to disable the default Odoo behaviour when there is no payment mode. I could not determine that by reading the code and the tests.

pedrobaeza:
The reasoning for emptying the bank account is to keep the "non deterministic" state until the payment/debit order is done. Check this use case (which is more usual than you think):

- You issue an invoice with 60 days due date.
- When the due is over, you make the debit order.
- Meanwhile, the customer has changed its bank account.

Having the bank account fixed on invoice creation, provokes returns due to the invalid bank account number.

And a lot of users don't have decided yet on invoice issuing the payment mode, so they don't set it.

That's why we try to always keep it empty. In fact, we never mark the `bank_account_required` flag in the payment method in any of our installations.

I also remember a glitch that puts the bank account without this part on the invoice creation, but this may changed over the versions.

Removing this part, it will affect installations relying on that behavior.

Something we can do is to add a company flag `bank_account_required` for modeling both behaviors according it. Other option is to dynamically set the bank account if not set (this is already done for the invoice PDF if selected) on the PEPPOL file, but I suppose this is not under your control.

grindtildeath:
@pedrobaeza @sbidoul How can we unblock this PR? It's a shame having to integrate this in all our 16.0+17.0 instances when account_payment_partner is installed

pedrobaeza:
Nobody has answered my comment at https://github.com/OCA/bank-payment/pull/1418#issuecomment-2692140380, and anyway, this CI is red.

remi-filament:
Hi @pedrobaeza I think that the issue you describe in https://github.com/OCA/bank-payment/pull/1418#issuecomment-2692140380 is valid for bills and out payments, but is not so relevant for customer invoices and in payment where what we want to add on account.move is our own bank account, not the one from the customer which can be calculated (this was already my proposal here : https://github.com/OCA/bank-payment/issues/1290#issuecomment-2561110404)

pedrobaeza:
@remi-filament having only one field for the bank account, if you add your own bank account in there, the payment order will be issued incorrectly. Why is not OK to have that data dynamically populated in the invoice report? That's already the case. Do you know it?

remi-filament:
You mean here : https://github.com/odoo/odoo/blob/9ab9e254cb0658e38dc0b9c4fe54c317e184bb04/addons/account/views/report_invoice.xml#L207 ?

The point is that if you do not use any payment mode on customer invoices, the field is not filled anymore, so you need to fill it manually for each customer invoice you issue (even worse when using contracts where invoices are auto-generated and you do not expect to have to go to each to set your bank account)

cyrilmanuel:
hi @sbidoul  is it possible to fix this pr or rebase this one ? this PR block others prs, perhaps we can go with this solutions no ?

sbidoul:
@cyrilmanuel I rebased, but this PR is blocked by https://github.com/OCA/bank-payment/pull/1418#issuecomment-2692140380. I still think this PR provides a better behaviour that respects how standard Odoo works, though.

gurneyalex:
@sbidoul 

Pedro wrote:
> Something we can do is to add a company flag bank_account_required for modeling both behaviors according it. 

Is this something that would be acceptable? Could we contribute this to your pull request?

sbidoul:
> Is this something that would be acceptable? Could we contribute this to your pull request?

@gurneyalex  Feel free to pick up this PR, of course.

pedrobaeza:
Merging #1555 that includes this, but through a configurable option.

PR #1420, opened 2025-02-15
[16.0][IMP] account_*: Remove demo data dependancy for tests

etobella:


kluna1998 (approved):

luisDIXMIT (approved):
LGTM!

pedrobaeza:
Thanks for decoupling the tests from demo data. That's great. However, the first 2 commits are note related to them and it's changing business logic. Can you please remove them and propose them in separate PRs with the reasoning behind?

pedrobaeza (approved):
There's still one change in business logic, but I think it makes sense and it's self explanatory, so let's unblock this:

/ocabot merge patch

PR #1409, opened 2025-01-30
[17.0][FIX] account_payment_sale: Company error validation when user changes company in sale order

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden @victoralmau

PR #1421, opened 2025-02-17
[16.0][FIX] account_payment_sale: Sale Default Payment Mode need to belong to the same Company as the Order

mbcosta:
Sale Default Payment Mode need to belong to the same Company as the Order, simple PR and should not affect the actual behavior, but with this avoid error in some CI Multi-Company cases, for example:

```bash
    raise UserError("\n".join(lines))
odoo.exceptions.UserError: Incompatible companies on records:
- 'SN l10n_br_sale - Produtos' belongs to company 'Empresa Simples Nacional' and 'Payment Mode' (payment_mode_id: 'Inbound Credit Trf Société Générale') belongs to another company.
```

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -20,10 +20,9 @@ class SaleOrder(models.Model):
     @api.depends("partner_id")
     def _compute_payment_mode(self):
         for order in self:
-            if order.partner_id:
-                order.payment_mode_id = order.partner_id.customer_payment_mode_id
-            else:
-                order.payment_mode_id = False
+            order.payment_mode_id = order.partner_id.with_company(
+                order.company_id
+            ).customer_payment_mode_id
 
     def _get_payment_mode_vals(self, vals):
         if self.payment_mode_id:


rvalyi (approved):

pedrobaeza (changes requested):
What you have to do is:

```python
order.payment_mode_id = order.partner_id.with_company(order.company_id).customer_payment_mode_id
```

And this is `[FIX]`, not `[REF]`.

pedrobaeza:
And you can remove the `if`, as the result is the same.

pedrobaeza (commented):
Remove the `if` and put just the new code, as if not partner, it will be False and you reduce code.

pedrobaeza (approved):
/ocabot merge patch

Please check if needed in 17 (I remember a similar patch).

mbcosta:
Thank you @rvalyi @pedrobaeza for review, yes Pedro in v17 has a similar change but kept the IF

PR https://github.com/OCA/bank-payment/pull/1409

image

Should I make the Port Foward?

PR #1422, opened 2025-02-18
[17.0][REF] account_payment_sale: Simplify Default method, FWD 1421

mbcosta:
Simplify Default method, Foward Port https://github.com/OCA/bank-payment/pull/1421


pedrobaeza (approved):
Yes, less code to maintain for the same result and with no performance penalty (the if is similar to the assignment).

Fast-tracking this:

/ocabot merge patch

PR #1425, opened 2025-02-24
[16.0][FIX] account_*: Post-install test + fallback to load CoA

pedrobaeza:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with errors like:

```
odoo.exceptions.RedirectWarning: ('Cannot find a chart of accounts for this company, You should configure it. \nPlease go to Account Configuration.', 246, 'Go to the configuration panel', None)
```

or

```
odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale
```

or

```
psycopg2.errors.CheckViolation: new row for relation "account_move_line" violates check constraint "account_move_line_check_accountable_required_fields"
```

All of them provoked by the lack of a CoA installed.

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA, and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/account_banking_mandate/tests/test_invoice_mandate.py
+++ b/account_banking_mandate/tests/test_invoice_mandate.py
@@ -5,12 +5,13 @@ from unittest.mock import patch
 
 from odoo import fields
 from odoo.exceptions import UserError
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import TransactionCase, tagged
 
 from odoo.addons.account.models.account_payment_method import AccountPaymentMethod
 from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
 
 
+@tagged("post_install", "-at_install")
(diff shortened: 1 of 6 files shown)


pedrobaeza:
A failing example can be found in https://github.com/OCA/l10n-spain/actions/runs/13498660769/job/37711546321?pr=4040

HaraldPanten (approved):

pedrobaeza:
Tested on integration tests and working:

/ocabot merge patch

PR #1427, opened 2025-02-26
[16.0][FIX] account_payment_partner: Fallback to load CoA

victoralmau:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/account_payment_partner/tests/test_account_payment_partner.py
+++ b/account_payment_partner/tests/test_account_payment_partner.py
@@ -16,7 +16,15 @@ class TestAccountPaymentPartner(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
-
+        if not cls.env.company.chart_template_id:
+            # Load a CoA if there's none in current company
+            coa = cls.env.ref("l10n_generic_coa.configurable_chart_template", False)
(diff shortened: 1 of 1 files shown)


luisDIXMIT (approved):
LGTM!

pedrobaeza:
The code is already in https://github.com/OCA/bank-payment/pull/1427/files#diff-226621acbc1a2075b0fb41ad42eb49f2556e3033091746026536239acd6789e1R41

Why do you need it again?

victoralmau:
Right now I'm not sure, I think it's because main_company is checked instead of cls.env.company (in that case I would change the previously existing code instead of repeating it), I'll leave the PR in draft and I'll comment with the conclusion.

pedrobaeza:
Can this be closed?

victoralmau:
Yes, finally this is not necessary.

PR #1430, opened 2025-03-13
[16.0][FIX] account_payment_method_fs_storage: improve test by removing usage of datas

AnizR:
No modification in the behavior, simply a correction in the test suite.

The usage of data within `account_payment_method_fs_storage`'tests seems to create an error:
https://github.com/OCA/bank-payment/pull/1360

I 'manually' created the same data in my tests.


AnizR:
@pedrobaeza as promised, here is the correction of my tests :smile:

pedrobaeza (approved):
Thanks!

/ocabot merge patch

PR #1431, opened 2025-03-14
[IMP][account_payment_order] Add Transmit Invoices filter

vehi-invitu:
cc @cvinh

### DIFF
--- a/account_payment_order/views/account_invoice_view.xml
+++ b/account_payment_order/views/account_invoice_view.xml
@@ -95,4 +95,25 @@
             </button>
         </field>
 </record>
+    <record id="view_account_move_search_inherit" model="ir.ui.view">
+        <field name="name">view.account.move.search.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_account_invoice_filter" />
+        <field name="priority" eval="100" />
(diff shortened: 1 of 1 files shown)


cvinh (changes requested):
Thanks for your PR.
IMHO, it's better to filter invoices that has NOT been part of a payment order instead of invoices that has been paid.

account_payment_order/views/account_move_views.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
+    <record id="view_account_move_search_inherit" model="ir.ui.view">
+        <field name="name">view.account.move.search.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_account_invoice_filter" />
+        <field name="priority" eval="10" />
```
cvinh:
```suggestion
        <field name="priority" eval="100" />
```

cvinh (changes requested):
I think it should be transmitted invoices instead of transmit invoices
Please check it

cvinh (approved):
Thanks

cvinh:
@vehi-invitu can you rebase please ? runboat is stuck

PR #1434, opened 2025-03-17
[16.0][OU-FIX] account_payment_mode: Ensure that all lines have translations

etobella:
It is a weird case, but I found Databases where we couldn't find a translation and the field was left empty. then the transformation failed.


pedrobaeza:
This is not needed with https://github.com/OCA/openupgradelib/pull/406.

PR #1435, opened 2025-03-24
[17.0][FIX/IMP] account_payment_order: use payment_reference if present for out payments

MaxyMoos:
Forward-porting of #1331 (16.0)


sbidoul (changes requested):
Could you cherry-pick the original commit in order to preserve the commit message and authorship?

MaxyMoos:
Well now i'm feeling a bit stupid because it was way simpler but hadn't even crossed my mind! Thanks for the suggestion, it's done.

sbidoul (approved):
Thanks!

PR #1438, opened 2025-03-26
[18.0][MIG] account_payment_mode: Migration to version 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/bank-payment/pull/1367
TT54087
@Tecnativa @pedrobaeza @carlosdauden @juancarlosonate-tecnativa could you please review this?

PR #1415, opened 2025-02-07
[18.0][MIG] account_payment_partner: Migration to 18.0

absal-smile:

PR #1439, opened 2025-03-26
[18.0][MIG] account_payment_partner: Migration to version 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/bank-payment/pull/1415

Depends on:

- [x] https://github.com/OCA/bank-payment/pull/1438

TT51707
@Tecnativa @pedrobaeza @victoralmau @etobella @JordiMForgeFlow @carlosdauden could you please review this?


account_payment_partner/security/account_payment_partner_security.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="group_show_account_payment_method" model="res.groups">
+        <field name="name">Show the default payment method in contacts</field>
```
carlosdauden: We should specify something else so that this group does not cause confusion when viewing the option from the user's permissions.
If I read it as it is, I'm sure I'll mark it. :wink: 
```suggestion
        <field name="name">Show the default payment method line in contacts</field>
```

carlos-lopez-tecnativa: I updated the name, but if you have a better name for the new group, please let me know. Thanks.

JordiMForgeFlow (approved):
LGTM 👍🏼

account_payment_partner/views/res_partner_view.xml
```diff
@@ -21,6 +21,19 @@
...
+                >account_payment_partner.group_show_account_payment_method</attribute>
+            </field>
+            <field
+                name="property_outbound_payment_method_line_id"
+                position="attributes"
+            >
+                <attribute
+                    name="groups"
+                >account_payment_partner.group_show_account_payment_method</attribute>
+            </field>
```
sbidoul, line 36: This is the drawback of not using the standard Odoo fields. If for any reason someone needs to use the standard fields, they end up with a very confusing user interface with duplicate Payment Method/Mode fields with almost identical names and meanings on partners and invoices. 

If we want to continue using payment modes, we need a better solution to that problem.

pedrobaeza: The solution is already setup: a group for not showing the payment method lines. What do you want instead?

gdgellatly: Could we not do something like add a field to Odoo Payment Mode which is a many2one to OCA Payment Mode (in practice operates as One2one)? Then we can use the field both ways, no hiding.

Sure there is probably a little adaptation to a couple functions.

pedrobaeza: Not really. They aren't equivalent: from one OCA payment mode you can have multiple account.payment.method.line, one per journal. They are set in the `account.payment` resulting record depending on which journal you select on the payment order.

The solution for this rare case is to have 2 security groups for showing/hiding what you want.

sbidoul: > The solution for this rare case is to have 2 security groups for showing/hiding what you want.

I don't think it is a rare case.

Imagine situations where you want to have some invoices paid with OCA SEPA direct debit, and other invoices paid with PayPal with the standard Odoo payment provider.

You would be obliged to show both the OCA payment mode and the Odoo Payment Method Line, and this would be super confusing for users.

> Could we not do something like add a field to Odoo Payment Mode which is a many2one to OCA Payment Mode (in practice operates as One2one)? Then we can use the field both ways, no hiding.

We could maybe do that but at this point, Alexis' approach of extending `acount.payment.method.line` with the additional features we need feels simpler.

pedrobaeza: Well, rare or not, what you are describing is not happening, as there's so few code associated with the standard field:

https://github.com/search?q=repo%3Aodoo%2Fodoo%20property_inbound_payment_method_line_id&type=code

And the preferred method line dumped to the invoice is only being used as suggestion for the register payment wizard, basically for the journal:

https://github.com/search?q=repo%3Aodoo%2Fodoo+preferred_payment_method_line_id&type=code

so no logic associated (for now) with online payments and so on. And if in the future there's some, then we can consider switching, or relabel fields, or the contrary: associate that logic to the payment modes, which is already done for deducing the payment method line on the `account.payment` standard record.

What I see that is deforming more the standard is to artificially limit the selection values in that m2o, and to add a lot of logic, fields, etc to them, as the refactoring is doing.

sergioM-S73:
Shouldn't the dependency be added to the module that creates "mandate_id" ?
I'm getting this error when printing an invoice:

image

The module that creates the field in v17 is "account_banking_mandate"

pedrobaeza:
@sergioM-S73 yes, that part should be extracted outside this one. We will do it after merging `account_payment_partner`. Anyway, this is not something exclusive of this migration, but dragged from previous ones.

pedrobaeza (approved):
As for now, there seems to be a "fork", we don't dig more on the fields UI, just the current visibility group over core fields. If this changes in the future, we can open improvement PRs.

/ocabot merge nobump

PR #1414, opened 2025-02-06
[18.0][MIG]  account_payment_order: Migration to 18.0

absal-smile:

PR #1440, opened 2025-03-26
[18.0][MIG] account_payment_order: Migration to version 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/bank-payment/pull/1414
Depends on:

- [x] https://github.com/OCA/bank-payment/pull/1438
- [x] https://github.com/OCA/bank-payment/pull/1439

TODO:

- [x] Check the implications related to this PR https://github.com/odoo/odoo/pull/180813 when the payment does not generate account.move (the test failures are related).

TT54088

@Tecnativa


carlos-lopez-tecnativa:
@pedrobaeza @victoralmau @etobella @JordiMForgeFlow @carlosdauden could you please review this?

pedrobaeza:
The 2 extra separated commits are part of the migration (at least the blocked one), as it's something of the process of making this module compatible with the new version.

account_payment_order/migrations/18.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2025 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env, "account_payment_order", "18.0.1.0.0/noupdate_changes.xml"
```
Reyes4711-S73:
```suggestion
        env, "account_payment_order", "migrations/18.0.1.0.0/noupdate_changes.xml"
```
When I tried to update the module, the migration script fails due to the fact that the path written in the script can't find the file

carlos-lopez-tecnativa: Solved, thanks!

JordiMForgeFlow (approved):
code + functional

etobella (approved):

pedrobaeza (approved):
I have squashed a bit the final commit history, and join the migration commits, as the initial attempt of the other contributor was adding incorrect things (like translating a domain!) that later was needed to be reversed again, so that was dirtying the history. I have also removed the reference assignation to `memo`, as it was not correct.

/ocabot merge nobump

PR #1441, opened 2025-04-02
[17.0][FIX] account_payment_order: Fix chatter HTML link formatting on invoice payment

MaxyMoos:
Fixes the chatter messages posted on invoices when clicking on the `Add to Payment Order` button, which were not escaped.

Before:
before

After:
after

### DIFF
--- a/account_payment_order/i18n/account_payment_order.pot
+++ b/account_payment_order/i18n/account_payment_order.pot
@@ -18,8 +18,8 @@ msgstr ""
 #: code:addons/account_payment_order/models/account_move.py:0
 #, python-format
 msgid ""
-"%(count)d payment lines added to the existing draft payment order <a href=# "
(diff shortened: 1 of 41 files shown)


pedrobaeza (commented):
Thanks for the fix. Please modify the .pot/.po for accommodating to this change and not having to make the full translation.

MaxyMoos:
@pedrobaeza Thanks for looking at the PR.
Unsure of whether only the .pot or all .po files had to be modified, I added a commit that adapted all files to the new string format, adapting the few actual translations too. I'll squash this new commit with the previous if you confirm that was the right approach.

pedrobaeza (approved):
Thank you for the effort!

I think we can fast-track this, as it's an obvious patch

/ocabot merge patch

PR #1442, opened 2025-04-07
[16.0][FIX] account_banking_pain_base : fix DbtrAcct node

luc-demeyer:
The DbtrAcct node should contain the currency of the bank account for outgoing payments.

It's not a required element in the pain XSD but not specifying it can have bad consequences when paying from multi-currency bank accounts.

I encountered this issue twice in the last year with German and Balgian banks.

Assume you have a bank account number with two financial journals for EUR and USD.

If you initiate a USD payment from your EUR bank account than the bank will try to remove the money from your USD bank account (which is a problem if you don't have enough dollars on that account).

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -519,6 +519,11 @@ class AccountPaymentOrder(models.Model):
             party_account_other = etree.SubElement(party_account_id, "Othr")
             party_account_other_id = etree.SubElement(party_account_other, "Id")
             party_account_other_id.text = partner_bank.sanitized_acc_number
+        if self.payment_type == "outbound" and party_type == "Dbtr":
+            currency = etree.SubElement(party_account, "Ccy")
+            currency.text = (
+                self.journal_id.currency_id.name or self.company_currency_id.name
+            )
         return True
 
     @api.model


account_banking_pain_base/models/account_payment_order.py
```diff
@@ -519,6 +519,15 @@ def generate_party_acc_number(
             party_account_other = etree.SubElement(party_account_id, "Othr")
             party_account_other_id = etree.SubElement(party_account_other, "Id")
             party_account_other_id.text = partner_bank.sanitized_acc_number
+
+        if self.payment_type == "outbound" and party_type == "Dbtr":
+            currency = etree.SubElement(party_account, "Ccy")
+            currency.text = (
+                self.journal_id.currency_id
+                and self.journal_id.currency_id.name
+                or self.company_currency_id.name
```
pedrobaeza:
```suggestion
                self.journal_id.currency_id.name
                or self.company_currency_id.name
```

Or it should be the currency of each move line?

fabian-spantech (approved):

BenjaHe (approved):
OK

account_banking_pain_base/models/account_payment_order.py
luc-demeyer: @pedrobaeza 
PR adapted with your suggestion.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -519,6 +519,13 @@ def generate_party_acc_number(
             party_account_other = etree.SubElement(party_account_id, "Othr")
             party_account_other_id = etree.SubElement(party_account_other, "Id")
             party_account_other_id.text = partner_bank.sanitized_acc_number
+
```
pedrobaeza: Please don't add empty lines inside a method.

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions

PR #1445, opened 2025-04-14
[17.0][IMP] account_payment_mode: change type of payment_mode.note field from Text to Html

pilarvargas-tecnativa:
fw-port of https://github.com/OCA/bank-payment/pull/1360

@Tecnativa TT56010

@pedrobaeza @sergio-teruel please review

### DIFF
--- a/account_payment_mode/models/account_payment_mode.py
+++ b/account_payment_mode/models/account_payment_mode.py
@@ -67,7 +67,7 @@ class AccountPaymentMode(models.Model):
         related="payment_method_id.code", readonly=True, store=True
     )
     active = fields.Boolean(default=True)
-    note = fields.Text(translate=True)
+    note = fields.Html(translate=True)
     sequence = fields.Integer(default=10)
 
     @api.onchange("company_id")


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge minor

etobella:
@pedrobaeza @pilarvargas-tecnativa Shouldn't this need a migration script?

pilarvargas-tecnativa:
> @pedrobaeza @pilarvargas-tecnativa Shouldn't this need a migration script?

Hi @etobella, this change was made in v16 with his script, when migrating the module to v17 this commit had not been included so the bd's that have already been migrated to v17 have the field as text and contain something like `<p> Text </p>` and it is displayed like this literally in the field and in the reports. When testing by running the script on the migrated field in v17 with html tags in the text, it is not interpreted and the text is stored and displayed like this: `<p>&lt;p&gt;Text&lt;/p&gt;</p>`
For that reason the script already running in v16 has been removed.

etobella:
:+1: 

I didn't notice that it comes from a Forward Port. Thanks for the clarification

etobella:
Actually, After giving 5 seconds of thought the Scripts are necessary. Imagine that you started on 17. Then the field was text, and after this change, it might be broken.

pilarvargas-tecnativa:
> Actually, After giving 5 seconds of thought the Scripts are necessary. Imagine that you started on 17. Then the field was text, and after this change, it might be broken.

Right, different scenarios: 

- Start in v17 with the text field (in this case the script would work).
- Coming from v16 and having in v17 the html field as text with the tags (the script would break the data).
- Be still in v16 and move to v17 (in this case it would not be necessary).

Within these scenarios, if you add the script otherwise, you have to redo by hand in one of the first two. How do you decide which is the priority?

etobella:
Well, there are options for this. For example:

1- Check the migration version (not the best option, but it might work in 90% of the cases)
2- Check if the field has html tags

pedrobaeza:
The second is not very feasible, and the first one may have several variables. Please go ahead with the script if you have it clear.

etobella:
Using the following query might work (tested with other fields and tables)

```sql
SELECT id from account_payment_mode where note::text ~* '<[^>]+>' LIMIT 1
```

If it is OK for you, I can do the migration script.

pedrobaeza:
OK, go ahead with it, which seems good enough

PR #1452, opened 2025-04-23
[17.0][FIX] account_banking_pain_base : fix DbtrAcct node

luc-demeyer:
The DbtrAcct node should contain the currency of the bank account for outgoing payments.

It's not a required element in the pain XSD but not specifying it can have bad consequences when paying from multi-currency bank accounts.

I encountered this issue twice in the last year with German and Balgian banks.

Assume you have a bank account number with two financial journals for EUR and USD.

If you initiate a USD payment from your EUR bank account than the bank will try to remove the money from your USD bank account (which is a problem if you don't have enough dollars on that account).

### DIFF
--- a/account_banking_pain_base/models/account_payment_order.py
+++ b/account_banking_pain_base/models/account_payment_order.py
@@ -515,6 +515,11 @@ class AccountPaymentOrder(models.Model):
             party_account_other = etree.SubElement(party_account_id, "Othr")
             party_account_other_id = etree.SubElement(party_account_other, "Id")
             party_account_other_id.text = partner_bank.sanitized_acc_number
+        if self.payment_type == "outbound" and party_type == "Dbtr":
+            currency = etree.SubElement(party_account, "Ccy")
+            currency.text = (
+                self.journal_id.currency_id.name or self.company_currency_id.name
+            )
         return True
 
     @api.model


luc-demeyer:
cherry-pick of https://github.com/OCA/bank-payment/pull/1442

pedrobaeza:
Please remove `[16.0]` from the commit message.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -515,6 +515,11 @@ def generate_party_acc_number(
             party_account_other = etree.SubElement(party_account_id, "Othr")
             party_account_other_id = etree.SubElement(party_account_other, "Id")
             party_account_other_id.text = partner_bank.sanitized_acc_number
+        if self.payment_type == "outbound" and party_type == "Dbtr":
+            currency = etree.SubElement(party_account, "Ccy")
+            currency.text = (
+                self.journal_id.currency_id.name or self.company_currency_id.name
+            )
```
patrickt-oforce, line 522: In Italy with this node the payment file is rejected from bank, because the standard not specificy the currency node for debtor account. Actually I don't know if every SEPA debit / credit pain specifiy this node in DbtrAcct; but if it is only for Sepa Credit Transfer, suggest to override the method in the module `account_banking_sepa_credit_transfer` and add the node there conditioned to PAIN flavour

luc-demeyer: @patrickt-oforce 
I can hardly believe that the presence of the Ccy tag is the reason of the rejection.
I also checked the pain.001.001.09.xsd and also in this newer standard (becoming mandatory by november 2025 !) the Ccy tag remains optional.

Can you send me 
- the failing pain file
- the error message from the bank

Since a payment file is confidential, please send me the data offline (luc.demeyer@noviat.com). I can send an NDA if required.

patrickt-oforce: hi @luc-demeyer this the bank error about Ccy attribute
image
You can find the xsd for the italian sct on https://github.com/OCA/l10n-italy/blob/16.0/l10n_it_sct_cbi/data/standards/CBIPaymentRequest.00.04.01.xsd#L45
If you check the type CBICashAccount2 there is no mention about Ccy node

The sct file is modified to hide some information, but the nodes are untouched, the extension is txt because Github not allow to load xml file, need to change the fiel extension
[sct_PAY0318_OCA.txt](https://github.com/user-attachments/files/20406109/sct_PAY0318_OCA.txt)

Add to this, if you want you can check this [PR#4751 in OCA/l10n-italy](https://github.com/OCA/l10n-italy/pull/4751)

PR #1317, opened 2024-07-12
[14.0][IMP] add compute in account_invoice_select_for_payment in case of paid

HekkiMelody:
Reviving #1194 

> 1- Make possible to choice selected_for_payment in case of patially paid invoices
> 2- Make a compute in a field when invoices are paid
> 3- Make possible to choice in a view form through a green button To pay
> 4-Once the invoice is to pay, add a ribbon to inform users at this subject

PR #1453, opened 2025-04-24
[18][account_invoice_select_for_payment] Remove the to pay mark on an invoice when it is fully paid or in payment

florian-dacosta:
Presently, the module only works with the register payment button, which seems quite problematic, as it is normally not very used when the full accounting is done in Odoo.
Indeed, if you pay your invoice by wire transfer, the payment will come from your bank statement or from an account.payment.order for instance.

So the main change here is that the selected_for_payment is computed so it is removed when invoice is paid or in payment.

This PR includes other small improvement like :

    possibility to mark the invoice as to pay from the form invoice view.
    Make it possible to hide the field on tree view, since it is probably only for some users. Also, this allow to hide it on the customer invoice view, since it mainly concerns the supplier invoices.

Last thing, I think it is very weird to remove the selected_for_payment mark if one user click on the register payment button but does not proceed to the payment (cancel it).
I would remove the override of action_register_payment and only rely on the computed field to remove it...
Still I did not do it in this PR because I am not sure it is ok for everybody, and I do not really have a solution for the partial payment case.

An equivalent PR actually was already proposed/approved in v14, but was never merged see : https://github.com/OCA/bank-payment/pull/1317

FYI @aleuffre @renda-dev
@Kimkhoi3010 Can you re-do your review over here?
@sebalix @TDu @grindtildeath as reviewer of the migration PR it may be of your interest

### DIFF
--- a/account_invoice_select_for_payment/models/account_move.py
+++ b/account_invoice_select_for_payment/models/account_move.py
@@ -1,13 +1,19 @@
 # Copyright 2020 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    selected_for_payment = fields.Boolean("To Pay")
+    selected_for_payment = fields.Boolean(
+        string="To Pay",
+        compute="_compute_selected_for_payment",
+        readonly=False,
+        store=True,
+        tracking=True,
+    )
 
     def action_toggle_select_for_payment(self):
         selected = self.filtered(lambda rec: rec.selected_for_payment)
@@ -16,6 +22,12 @@ class AccountMove(models.Model):
             selected.write({"selected_for_payment": False})
         if unselected:
             unselected.write({"selected_for_payment": True})
+
+    @api.depends("payment_state")
+    def _compute_selected_for_payment(self):
+        for rec in self:
+            if rec.payment_state in ("in_payment", "paid", "reversed"):
+                rec.selected_for_payment = False
 
     def action_register_payment(self):
         invoices = self.filtered(lambda inv: inv.selected_for_payment)
--- a/account_invoice_select_for_payment/tests/test_account_invoice_select_for_payment.py
+++ b/account_invoice_select_for_payment/tests/test_account_invoice_select_for_payment.py
@@ -1,3 +1,4 @@
+from odoo import fields
 from odoo.tests import TransactionCase
 
 
@@ -26,7 +27,8 @@ class TestAccountMoveAndPayment(TransactionCase):
     def _create_invoice(cls, partner_id, selected_for_payment):
         invoice = cls.env["account.move"].create(
             {
-                "move_type": "out_invoice",
+                "invoice_date": fields.Date.today(),
+                "move_type": "in_invoice",
                 "partner_id": partner_id,
                 "selected_for_payment": selected_for_payment,
             }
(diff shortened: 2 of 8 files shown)


florian-dacosta:
@aleuffre @renda-dev @Kimkhoi3010 @sebalix @TDu @grindtildeath

I think it is an important improvement, a review would be appreciated.
@LESTRAT21

florian-dacosta:
Trying once more :)
@aleuffre @renda-dev @Kimkhoi3010 @sebalix @TDu @grindtildeath @LESTRAT21

account_invoice_select_for_payment/views/account_move.xml
```diff
@@ -1,14 +1,51 @@
...
+                <button
+                    name="action_toggle_select_for_payment"
+                    string="Not To Pay"
+                    type="object"
+                    invisible="selected_for_payment == False or payment_state not in ('not_paid', 'partial') or move_type not in ('in_invoice', 'out_refund')"
+                    class="btn-primary"
+                    style="background-color:grey"
+                />
+            </button>
+            <xpath expr="//sheet/widget[5]" position="after">
```
HekkiMelody, line 25: **question**: would it be possible to choose a field in a more deterministic way?

e.g. `//sheet/widget[@name='some_widget']` or even `//sheet/widget[last()]` ?

HekkiMelody (approved):
I don't really know the module, so I can't say functionally, but aside from a small note, Code Review LGTM.

account_invoice_select_for_payment/models/account_move.py
```diff
@@ -1,13 +1,19 @@
...
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    selected_for_payment = fields.Boolean("To Pay")
+    selected_for_payment = fields.Boolean(
+        string="To Pay",
+        compute="_compute_selected_for_payment",
+        readonly=False,
+        store=True,
+        tracking=True,
```
grindtildeath, line 15: IMO this will involve lots of unneeded messages

account_invoice_select_for_payment/tests/test_account_invoice_select_for_payment.py
```diff
@@ -26,7 +27,8 @@ def setUpClass(cls):
     def _create_invoice(cls, partner_id, selected_for_payment):
         invoice = cls.env["account.move"].create(
             {
-                "move_type": "out_invoice",
+                "invoice_date": fields.Date.today(),
+                "move_type": "in_invoice",
```
grindtildeath, line 31: Why do you need to change the move_type in a test? 🤔 

(IMO it's better to add a param with default value so that you can change in a new test what it's doing)

account_invoice_select_for_payment/views/account_move.xml
```diff
@@ -1,14 +1,51 @@
...
+                <widget
+                    name="web_ribbon"
+                    title="To Pay"
+                    bg_color="bg-info"
+                    invisible="selected_for_payment == False or payment_state != 'not_paid' or move_type not in ('in_invoice', 'out_refund')"
+                />
+            </xpath>
+            <field name="partner_bank_id" position="after">
+                <field name="selected_for_payment" invisible="1" />
```
grindtildeath, line 34: IIRC that shouldn't be needed anymore

account_invoice_select_for_payment/views/account_move.xml
```diff
@@ -1,14 +1,51 @@
...
+
     <record id="view_invoice_tree" model="ir.ui.view">
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_invoice_tree" />
         <field name="arch" type="xml">
             <field name="name" position="before">
                 <field
                     name="selected_for_payment"
-                    invisible="payment_state not in ('not_paid', 'partial')"
+                    invisible="payment_state not in ('not_paid', 'partial') or move_type not in ('in_invoice', 'out_refund')"
```
grindtildeath, line 46: I'm concerned this might not suit well with existing installations

account_invoice_select_for_payment/models/account_move.py
```diff
@@ -17,6 +23,12 @@ def action_toggle_select_for_payment(self):
         if unselected:
             unselected.write({"selected_for_payment": True})
 
+    @api.depends("payment_state")
+    def _compute_selected_for_payment(self):
+        for rec in self:
+            if rec.payment_state in ("in_payment", "paid", "reversed"):
+                rec.selected_for_payment = False
```
grindtildeath, line 30: This will be very heavy if the module is installed on an existing DB with lots of invoices 😬

account_invoice_select_for_payment/views/account_move.xml
```diff
@@ -1,14 +1,51 @@
...
+                    style="background-color:grey"
+                />
+            </button>
+            <xpath expr="//sheet/widget[5]" position="after">
+                <widget
+                    name="web_ribbon"
+                    title="To Pay"
+                    bg_color="bg-info"
+                    invisible="selected_for_payment == False or payment_state != 'not_paid' or move_type not in ('in_invoice', 'out_refund')"
+                />
```
grindtildeath, line 31: I'm wondering if a pill wouldn't be less intrusive than a ribbon for this. (since having it selected doesn't mean as much as having it paid or not)

grindtildeath (commented):
Thanks for the contribution. That could be a useful change but IMO there are a few more things to pay attention to.

Also would be nice to have split commits between the change to a computed field and the extra views improvements

grindtildeath:
ping @gurneyalex since you were IIRC the original author

PR #1457, opened 2025-04-30
[16.0][ADD] account_payment_order_filename: new module 2.0

AnizR:
It improves PR #1308.

It keeps the original purpose: 
_Add new module account_payment_order_filename to let the user customize the file generated. This behavior use sequence to also have the possibility to generate a unique filename if necessary_

But, it is more generic.
Filename of every type of payment order (not only SEPA payment) can be changed.


AnizR:
ping @acsonefho @lmarion-source

alexis-via:
FYI, this feature is present in https://github.com/OCA/bank-payment/pull/1174 so it will be part of the base module of OCA/bank-payment-alternative

PR #1458, opened 2025-05-05
[18.0][IMP] account_payment_*: Add support for using payment orders on receipts

celm1990:
When a company uses receipts, for example, to import initial balances (instead of using journal entries), and later needs to pay those receipts, the previous process was complicated. Before this commit, using a payment order to pay these items was not straightforward. After this commit, users can generate a payment order for receipts in the same way as for invoices.

@pedrobaeza @victoralmau @etobella @AaronHForgeFlow @JordiMForgeFlow what do you think about this feature?

### DIFF
--- a/account_payment_order/models/account_move_line.py
+++ b/account_payment_order/models/account_move_line.py
@@ -29,7 +29,7 @@ class AccountMoveLine(models.Model):
     def _compute_partner_bank_id(self):
         for ml in self:
             if (
-                ml.move_id.move_type in ("in_invoice", "in_refund")
+                ml.move_id.move_type in ("in_invoice", "in_receipt", "in_refund")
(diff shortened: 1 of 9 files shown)


celm1990:
@pedrobaeza @victoralmau @etobella @AaronHForgeFlow @JordiMForgeFlow what do you think about this feature?

pedrobaeza (commented):
Seems OK, although it's weird to use receipts.

pedrobaeza (approved):

celm1990:
ping @victoralmau @etobella @AaronHForgeFlow @JordiMForgeFlow

victoralmau (approved):

etobella (approved):
/ocabot merge patch

PR #1470, opened 2025-05-19
[18.0][IMP] account_banking_mandate: add mandate field on account payment

JordiMForgeFlow:
In V18 account.payment does not inherit from account.move model anymore, as payments may not create a journal entry. To handle this new behavior we add a computed mandate field in account.payment that takes it either from its related move or from the payment lines.

### DIFF
--- a/account_banking_mandate/models/account_payment.py
+++ b/account_banking_mandate/models/account_payment.py
@@ -1,0 +1,23 @@
+# Copyright 2025 ForgeFlow
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import api, fields, models
+
+
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    mandate_id = fields.Many2one(
+        "account.banking.mandate", compute="_compute_mandate_id"
+    )
+
+    @api.depends(
+        "payment_line_ids.mandate_id",
+        "move_id.mandate_id",
+    )
+    def _compute_mandate_id(self):
+        for payment in self:
+            payment.mandate_id = (
+                payment.move_id.mandate_id or payment.payment_line_ids[:1].mandate_id
+            )
--- a/account_banking_mandate/models/__init__.py
+++ b/account_banking_mandate/models/__init__.py
@@ -5,3 +5,4 @@ from . import res_partner_bank
 from . import res_partner
 from . import account_payment_line
 from . import account_move_line
+from . import account_payment


JordiMForgeFlow:
@xaviedoanhduy in https://github.com/xaviedoanhduy/l10n-usa/pull/2 you are just filtering out the payments without related move_id to check the mandate. However, the mandate is linked to the account payment lines of the payment order, and for payments without entries we could check it from there. WDYT?

account_banking_mandate/models/account_payment.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "account.payment"
+
+    mandate_id = fields.Many2one(
+        "account.banking.mandate", compute="_compute_mandate_id"
+    )
+
+    @api.depends("payment_line_ids", "payment_line_ids.mandate_id")
+    def _compute_mandate_id(self):
+        for item in self:
+            item.mandate_id = item.payment_line_ids[:1].mandate_id
```
xaviedoanhduy: Hello, thanks for your work on improving this issue!

I'm not entirely convinced by the current approach. From what I understand, Odoo's architecture maintains a strong link between `account.payment` and `account.move` via the `move_id` field, which is considered a trusted source of accounting information.
For example, in this change: 
https://github.com/OCA/bank-payment/pull/1440/commits/43c010f906a92705ff997bb870697e85198e161f#diff-c2ba7d4de44953fb8d64ab39e502b50d72853f2aff2dac0d97488d3c29a4b8b1R86-R101 
you can see that the logic relies on `move_id.payment_order_id`, which suggests that integrity between both models is expected and preserved.

Given this, wouldn't it be safer to fallback to `payment_line_ids` *only* when `move_id` is not yet available?

Here's a suggestion that reflects this logic:

```suggestion
    @api.depends( "move_id.mandate_id", "payment_line_ids.mandate_id")
    def _compute_mandate_id(self):
        for payment in self:
            if payment.move_id:
                payment.mandate_id = payment.move_id.mandate_id
            elif payment.payment_line_ids:
                 payment.mandate_id = payment.payment_line_ids[:1].mandate_id
```

This way, we continue to respect the trust model already present in Odoo's design.

What do you think about this suggestion?

JordiMForgeFlow: @xaviedoanhduy makes sense, could you check now?

account_banking_mandate/models/account_payment.py
```diff
@@ -0,0 +1,26 @@
...
+
+    mandate_id = fields.Many2one(
+        "account.banking.mandate", compute="_compute_mandate_id"
+    )
+
+    @api.depends(
+        "payment_line_ids",
+        "payment_line_ids.mandate_id",
+        "move_id",
+        "move_id.mandate_id",
```
pedrobaeza:
```suggestion
        "payment_line_ids.mandate_id", "move_id.mandate_id",
```

account_banking_mandate/models/account_payment.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    def _compute_mandate_id(self):
+        for payment in self:
+            if payment.move_id:
+                payment.mandate_id = payment.move_id.mandate_id
+            else:
+                payment.mandate_id = payment.payment_line_ids[:1].mandate_id
```
pedrobaeza:
```suggestion
            payment.mandate_id = payment.move_id.mandate_id or payment.payment_line_ids[:1].mandate_id
```

account_banking_mandate/models/account_payment.py
xaviedoanhduy: I think it should just depend on `move_id.mandate_id` and `payment_line_ids.mandate_id`

and IMO is it better to use `elif` to branch the condition checking `payment_line_ids` if `move_id` doesn't exist? using `else` seems a bit risky if `payment` has no value in both fields above

or for better optimization we can use `filered` and combine with Pedro's suggestion below

```
@api.depends( "move_id.mandate_id", "payment_line_ids.mandate_id")
def _compute_mandate_id(self):
    for payment in self.filtered(
        lambda p: p.move_id.mandate_id or p.payment_line_ids.mandate_id
    ):
        payment.mandate_id = payment.move_id.mandate_id or payment.payment_line_ids[:1].mandate_id
```

JordiMForgeFlow: evaluating payment.payment_line_ids[:1] will not give an error if there are no payment_line_ids so I think it is fine to go with Pedro's suggestion

pedrobaeza (approved):

xaviedoanhduy (approved):

xaviedoanhduy:
Hi @pedrobaeza, could you merge it?

PR #1472, opened 2025-05-19
[18.0][FIX] account_banking_mandate: _rec_names_search

eugenios73:
Solution to warning issued when searching for mandates to assign to a sale order from the "sale.order" form view.

`odoo.models: Cannot search on display_name, no _rec_name or _rec_names_search defined on account.banking.mandate`

### DIFF
--- a/account_banking_mandate/models/account_banking_mandate.py
+++ b/account_banking_mandate/models/account_banking_mandate.py
@@ -19,6 +19,7 @@ class AccountBankingMandate(models.Model):
     _inherit = ["mail.thread", "mail.activity.mixin"]
     _order = "signature_date desc"
     _check_company_auto = True
+    _rec_names_search = ["unique_mandate_reference", "partner_bank_id.acc_number"]
 
     def _get_default_partner_bank_id_domain(self):
         if "default_partner_id" in self.env.context:


account_banking_mandate/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     "name": "Account Banking Mandate",
     "summary": "Banking mandates",
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.0.2",
```
pedrobaeza: Please don't change manually the version. It will be done by bot.

xaviedoanhduy (approved):
except for the above change mentioned by another reviewer it looks fine to me. 
thanks for fixing it

account_banking_mandate/__manifest__.py
pedrobaeza: and remove `[18.0]` from the commit message: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

eugenios73: @pedrobaeza thanks for the review. I solved your comments.

pedrobaeza (approved):
/ocabot merge patch

PR #1475, opened 2025-06-05
[18.0][FIX] account_banking_sepa_direct_debit: generation of 'Instruction Identification' and 'End to End Identification'

alexis-via:
BEFORE
sdd_bug

AFTER
fix_sdd

In Odoo v18, a draft account.payment don't have a move_id. So payment.move_id.id is False when generating the payment file (the account.payment are confirmed when going from 'generated' to 'uploaded'). So we can't use payment.move_id.id to generate 'Instruction Identification' and 'End to End Identification' in the generation of the SEPA XML file.

BTW, I don't think using the ID of the move or the ID of the account.payment in the identification tags of the SEPA XML file is a good practise... but that's another story. This change was introduced by the move to account.payment.

I found this bug while working on OCA/bank-payment-alternative... I'll fix it there too.

### DIFF
--- a/account_banking_sepa_direct_debit/models/account_payment_order.py
+++ b/account_banking_sepa_direct_debit/models/account_payment_order.py
@@ -164,7 +164,7 @@ class AccountPaymentOrder(models.Model):
                 )
                 instruction_identification.text = self._prepare_field(
                     "Instruction Identification",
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #1487, opened 2025-07-23
[18.0] [IMP] Payment Order: due on field in payment line wizard

baimont:
This is a migration of https://github.com/OCA/bank-payment/pull/539 with a few improvements

### DIFF
--- a/account_banking_mandate_contact/tests/test_account_payment_order.py
+++ b/account_banking_mandate_contact/tests/test_account_payment_order.py
@@ -97,7 +97,7 @@ class TestAccountPaymentOrder(BaseCommon):
             )
         )
         line_create_form.date_type = "due"
-        line_create_form.due_date = fields.Date.today()
+        line_create_form.filter_date = fields.Date.today()
         line_create = line_create_form.save()
         line_create.populate()
         line_create.create_payment_lines()
@@ -115,7 +115,7 @@ class TestAccountPaymentOrder(BaseCommon):
             )
         )
         line_create_form.date_type = "due"
-        line_create_form.due_date = fields.Date.today()
+        line_create_form.filter_date = fields.Date.today()
         line_create = line_create_form.save()
         line_create.populate()
         line_create.create_payment_lines()
--- a/account_payment_order/tests/test_payment_order_outbound.py
+++ b/account_payment_order/tests/test_payment_order_outbound.py
@@ -185,7 +185,7 @@ class TestPaymentOrderOutbound(TestPaymentOrderOutboundBase):
             self.env["account.payment.line.create"]
             .with_context(active_model="account.payment.order", active_id=order.id)
             .create(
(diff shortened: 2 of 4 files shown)


pedrobaeza (commented):
Although I still think it's dangerous, let's include it as this doesn't change the defaults, but please perform the following actions:

1. Squash both commits into one, being for the same feature.
2. The commit message should contain the module name: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

pedrobaeza (commented):
Thanks for the changes. As we are here, I don't get why having `due_date` and `move_date`. I think it's better to have only one field, and change in screen the label (or put a neutral one like simply "Date"). What do you think about simplifying that part?

baimont:
> Thanks for the changes. As we are here, I don't get why having `due_date` and `move_date`. I think it's better to have only one field, and change in screen the label (or put a neutral one like simply "Date"). What do you think about simplifying that part?

date_type, due_date and move_date are as old as v9: https://github.com/acsone/bank-payment/commit/6ae2903ee5279c8bd642b1c846e00a3e5ae10b3c

If we do that, I believe it should be done in an other pull request, don't you think?

pedrobaeza:
I don't think so, as in this one you are expanding these fields setting a new `_end` fields, so doubling them, and then un-doubling is not very practical. And furthermore, this is a TransientModel, so no need of migration scripts and so. What you can do is to put a commit before your changes doing that "merge", and then add your change on top just adding one `_end` field.

What do you think?

baimont:
Ok I get your point and I agree, it should be done here. Let me just immerge myself in `due_date` and `move_date` since I'm not familiar with these fields.

pedrobaeza:
Basically, both fields are for putting a date, but one or another is shown (and used in code), depending on the previous selection value.

baimont:
Ok, here https://github.com/OCA/bank-payment/blob/722985a65b4431b2daa036f757e1761a4b255735/account_payment_order/wizard/account_payment_line_create.py#L96 and here https://github.com/OCA/bank-payment/blob/722985a65b4431b2daa036f757e1761a4b255735/account_payment_order/wizard/account_payment_line_create.py#L100 we search on different fields based on the fact that we decided to use due_date or move_date. What should we do there?

pedrobaeza:
IMO, the actions should be:

- Have a generic field `filter_date`.
- On the search, use that same field for both.
- On UI, put the same field 2 times according visibility conditions, but adding a `string` attribute with the proper label, or if you prefer, just let the generic label "Filter date".

pedrobaeza (approved):
Thank you very much for the changes.

baimont:
> Thank you very much for the changes.

Of course, thanks for your review!

qgroulard (approved):
Nice improvement, LGTM :+1:

PR #1358, opened 2024-09-20
[16.0][MIG] account_payment_mode_default_account: Migration to 16.0

LauraCForgeFlow:

PR #1488, opened 2025-07-25
[18.0][MIG] account_payment_mode_default_account

RicardCForgeFlow:
Migration of https://github.com/OCA/bank-payment/pull/1358


account_payment_mode_default_account/tests/test_account_payment_mode_default_account.py
```diff
@@ -3,26 +3,118 @@
...
+            }
+        ),
+        "default_tax_sale": company.account_sale_tax_id,
+        "default_tax_purchase": company.account_purchase_tax_id,
+    }
+
+
+account_common.AccountTestInvoicingCommon.collect_company_accounting_data = classmethod(
+    _patched_collect_company_accounting_data
+)
```
JordiMForgeFlow: could you explain why is this needed?

account_payment_mode_default_account/tests/test_account_payment_mode_default_account.py
```diff
@@ -3,26 +3,118 @@
...
 
 
 class TestAccountPaymentModeDefaultAccount(TransactionCase):
```
JordiMForgeFlow, line 8: you can try using BaseCommon class to inherit for the test instead of TransactionCase

account_payment_mode_default_account/models/chart_template.py
```diff
@@ -4,13 +4,14 @@
 from odoo import models
 
 
-class AccountChartTemplate(models.Model):
+class AccountChartTemplate(models.AbstractModel):
     _inherit = "account.chart.template"
 
     def generate_properties(self, acc_template_ref, company):
```
JordiMForgeFlow: this method does not exist anymore in v18, could you check if it needs to be changed to something else? It may be _get_property_accounts()

account_payment_mode_default_account/hooks.py
```diff
@@ -1,39 +0,0 @@
...
-            SET name = %s,
-            fields_id = %s
-            WHERE fields_id = %s;
-        """
-        cr.execute(sql, (new_fname, new_model_field.id, orig_model_field.id))
-
-
-def uninstall_hook(cr, registry):
-    env = api.Environment(cr, SUPERUSER_ID, {})
-    fields_mapping = [
```
JordiMForgeFlow: odoo is not using ir.property anymore, but the company dependent fields are stored in the DB as jsonb to keep the values for the different companies. Therefore, you still need the hooks, but simply they should now be managing the filling of data directly in the field of the res.partner table

account_payment_mode_default_account/tests/test_account_payment_mode_default_account.py
RicardCForgeFlow: The problem is here: https://github.com/OCA/bank-payment/blob/0e85f61034f2e9b8686535514a4abd445ed35ff8/account_payment_order/tests/test_payment_order_inbound.py#L20

One option could be to do a PR to `account_payment_order` to use BaseCommon there. What do you think is cleaner, a patch in this module or changing `account_payment_order`

account_payment_mode_default_account/models/fields.py
```diff
@@ -0,0 +1,18 @@
+from odoo import fields
+
+
+def monkey_patch(cls):
```
JordiMForgeFlow, line 4: could you add a comment to explain that you are basing this on the module base_sparse_field where this is also used?

account_payment_mode_default_account/models/fields.py
```diff
@@ -0,0 +1,18 @@
...
+        name = func.__name__
+        func.super = getattr(cls, name, None)
+        setattr(cls, name, func)
+        return func
+
+    return decorate
+
+
+@monkey_patch(fields.Field)
+def get_company_dependent_fallback(self, records):
```
JordiMForgeFlow, line 17: could you add a comment saying that this is necessary because odoo makes an 'assert self.company_dependent'

PR #1496, opened 2025-09-10
[18.0][IMP] account_payment_order: make part of generated2uploaded inheritable

baimont:
we would like to jobify that part as we could with generate_move() in v12

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -415,7 +415,7 @@ class AccountPaymentOrder(models.Model):
         )
         return action
 
-    def generated2uploaded(self):
+    def post_and_reconcile(self):
         self.payment_ids.action_post()
         # Perform the reconciliation of payments and source journal items
         for payment in self.payment_ids:
@@ -425,6 +425,9 @@ class AccountPaymentOrder(models.Model):
                     lambda x, p=payment: x.account_id == p.destination_account_id
                 )
             ).reconcile()
+
+    def generated2uploaded(self):
+        self.post_and_reconcile()
         self.write(
             {"state": "uploaded", "date_uploaded": fields.Date.context_today(self)}
         )


pedrobaeza (commented):
Don't you think it should go everything together in the same method and enqueue the whole method? That way, you control which orders remain without perform that part.

baimont:
> Don't you think it should go everything together in the same method and enqueue the whole method? That way, you control which orders remain without perform that part.

We already have enqueued the whole generated2uploaded method. But we have a lot of payments per order and therefore we need to go even further. This allows me to create a job per payment.

pedrobaeza:
But the only way to get that is totally overriding the method, and the reconciliation is global AFAIK. I don't really see how you want to do that.

yvaucher:
@pedrobaeza Checking on our end we do rewrite `action_post_and_reconcile` in the context of jobs and each payment makes a reconcile calls is put in a batch of jobs with a finalization job.

Here is the idea:

```
    def action_post_and_reconcile(self):
        """
        Trigger reconciliation trough jobs.
        """
        if self.env.context.get("job_uuid"):
            # Perform the reconciliation of payments and source journal items
            batch_name = _(
                "Payment Order: Generate Move %s", fields.Date.context_today(self)
            )
            batch = self.env["queue.job.batch"].get_new_batch(batch_name)
            for payment in self.payment_ids:
                payment.with_context(job_batch=batch).with_delay(
                    priority=10
                ).action_post_and_reconcile()
            self.with_context(job_batch=batch).with_delay(
                priority=20
            )._finalize_generated2uploaded()
            return True
        return super().action_post_and_reconcile()
```

Each account.payment is posted and reconciled individually. 
The finalize method just free the lock on payment order for async task.

IMO here @baimont is making the simplest change to only target the lines to replace.

This PR changes nothing to the current state of the module but allows for more tweaking.

account_payment_order/models/account_payment_order.py
```diff
@@ -415,7 +415,7 @@ def open2generated(self):
         )
         return action
 
-    def generated2uploaded(self):
+    def action_post_and_reconcile(self):
```
yvaucher: The prefix `action_` suggests that we have a view action, please remove the prefix.

lmarion-source (approved):

pedrobaeza:
Please change the commit message. `[ADD]` is for adding modules, and follow the proper syntax:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

baimont:
> Please change the commit message. `[ADD]` is for adding modules, and follow the proper syntax:
> 
> https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

done

benwillig (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1497, opened 2025-09-17
[15.0][FIX]: account_payment_order: payment line domain add uploaded state

eduezerouali-tecnativa:
Add uploaded due to if you unreconcile a payment, you can add to payment order again and if you set the old payment order to draft, you can pay twice the bill(the old one + new one).

cc @Tecnativa TT58037
ping @carlosdauden @sergio-teruel

### DIFF
--- a/account_payment_order/models/account_move.py
+++ b/account_payment_order/models/account_move.py
@@ -128,7 +128,7 @@ class AccountMove(models.Model):
                     % move.name
                 )
             payment_lines = applicable_lines.payment_line_ids.filtered(
-                lambda l: l.state in ("draft", "open", "generated")
+                lambda l: l.state in ("draft", "open", "generated", "uploaded")
             )
             if payment_lines:
                 raise UserError(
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

carlosdauden:
Check if FWPort to next versions is necessary

sergio-teruel (approved):
Great @eduezerouali-tecnativa !!!

Good job!!

pedrobaeza (changes requested):
That was done on purpose because the payment may be returned, so using `account_payment_return`, the original move line is unreconciled again, and you can re-issue it in another debit order. With this, you are forbidding this. Am I wrong? cc @carlosdauden being the maintainer of the return modules, can you confirm?

carlosdauden:
Improvements should be made so that this method is easily extendable in the account_payment_return module, and these modifications should be made in that module so that an invoice included in a payment order but marked as returned can be included.

For now, we can close the PR, and if it is done, it will be from 18.0 onwards.

pedrobaeza:
OK, let's revisit this in the latest version.

PR #1499, opened 2025-09-19
[18.0][IMP] account_payment_order: add smart button to access related payments

Ricardo-MC:
This commit adds a smart button in the payment order form view to access the related payments easily. It also provides an action to display those payments and a counter field.

This improvement facilitates navigation and batch payment creation directly from the payment order.

### DIFF
--- a/account_payment_order/models/account_payment_order.py
+++ b/account_payment_order/models/account_payment_order.py
@@ -451,3 +451,16 @@ class AccountPaymentOrder(models.Model):
         ctx.update({"search_default_misc_filter": 0})
         action["context"] = ctx
         return action
+
+    def action_view_payments(self):
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "account.action_account_payments"
+        )
+        action.update(
+            {
+                "domain": [("payment_order_id", "=", self.id)],
+                "context": {"default_payment_order_id": self.id},
+            }
+        )
+        return action
--- a/account_payment_order/views/account_payment_order.xml
+++ b/account_payment_order/views/account_payment_order.xml
@@ -72,6 +72,19 @@
                                 widget="statinfo"
                             />
                         </button>
+                        <button
+                            class="oe_stat_button"
+                            name="action_view_payments"
+                            invisible="payment_count == 0"
+                            type="object"
+                            icon="fa-bars"
+                        >
+                            <field
+                                string="Payments"
+                                name="payment_count"
+                                widget="statinfo"
+                            />
+                        </button>
                     </div>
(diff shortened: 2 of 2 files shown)


account_payment_order/models/account_payment_order.py
```diff
@@ -451,3 +451,14 @@ def action_move_journal_line(self):
...
+    def action_view_payments(self):
+        self.ensure_one()
+        return {
+            "name": _("Payments"),
+            "type": "ir.actions.act_window",
+            "res_model": "account.payment",
+            "view_mode": "list,form",
+            "domain": [("payment_order_id", "=", self.id)],
+            "context": {"default_payment_order_id": self.id},
+        }
```
celm1990: There could be a better way to do this, without losing the views, since payments can be displayed in `kanban, graph, and activities`, not only in list and form views.

image

```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id(
            "account.action_account_payments"
        )
        action.update(
            "domain": [("payment_order_id", "=", self.id)],
            "context": {"default_payment_order_id": self.id},
        })
        return action
```

celm1990 (changes requested):
@Ricardo-MC

Ricardo-MC:
@celm1990,
Thank you very much for the correction.
Changes above.

celm1990 (approved):
LGTM

@pedrobaeza could you please review this?

pedrobaeza (commented):
It should be only shown when the order is confirmed. Isn't this already present in previous versions? If so, you should cherry-pick the commit instead.

account_payment_order/views/account_payment_order.xml
```diff
@@ -72,6 +72,18 @@
                                 widget="statinfo"
                             />
                         </button>
+                        <button
+                            class="oe_stat_button"
+                            name="action_view_payments"
+                            type="object"
+                            icon="fa-bars"
+                        >
```
celm1990, line 81:
```suggestion
                        <button
                            class="oe_stat_button"
                            name="action_view_payments"
                            invisible="payment_count == 0"
                            type="object"
                            icon="fa-bars"
                        >
```

celm1990 (changes requested):
> It should be only shown when the order is confirmed. Isn't this already present in previous versions? If so, you should cherry-pick the commit instead.

I don’t know. I reviewed other versions, but this feature is not present. If you remember it, please share the link with me.

celm1990 (approved):

celm1990:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge minor

PR #1504, opened 2025-10-01
[19.0][MIG] account_payment_mode: Migration to 19.0 and Merged account_payment_partner into account_payment_mode.

bizzappdev:
Merged `account_payment_partner` into `account_payment_mode`.


etobella (approved):
Functional review OK.

@OCA/banking-maintainers Wouldn't have sense to minimize the so many modules and leave it more compact? For example, in this case we have:

- account_payment_mode
- account_payment_partner

All modules depend on partner, not in mode, so, IMO, we could merge both modules in a single one. WDYT?

I leave open the decision of account_payment_order, as we are adding a strong dependency on other modules, but at the end, if you use account_payment_mode, you will be using account_payment_order, isn't it?

HaraldPanten:
Merging account_payment_mode and account_payment_partner could make sense. But account_payment_order and related addons would add a lot of dependencies that should be in difierent modules, IMO.

etobella:
I was opening the idea just for order, not sepa modules.

However, I agree that partner and mode is enough.

@bizzappdev can you handle this?

grindtildeath:
I agree with merging account_payment_mode and account_payment_partner :+1:

Alexgars73 (approved):
Functional Review OK!

bizzappdev:
> I was opening the idea just for order, not sepa modules.
> 
> However, I agree that partner and mode is enough.
> 
> @bizzappdev can you handle this?

After analysing, I also agree. Now the question is, what should the module be called? Integrate everything into account_payment_partner and deprecate the account_payment_mode module, or create a new module and deprecate both modules?
@etobella @HaraldPanten

etobella:
I would suggest to merge everything on account_payment_mode. For me specifies the goal of the module (create the payment mode) and the merge with partner is just a direct evolution of it.

HaraldPanten:
> I would suggest to merge everything on account_payment_mode. For me specifies the goal of the module (create the payment mode) and the merge with partner is just a direct evolution of it.

I agree merging both modules in account_payment_mode.

account_payment_mode/readme/CONTRIBUTORS.md
```diff
@@ -1,6 +1,18 @@
...
+- Stéphane Bidoul \<<stephane.bidoul@acsone.eu>\>
+- Danimar Ribeiro
+- Angel Moya \<<angel.moya@domatix.com>\>
+- [Tecnativa](https://www.tecnativa.com):
+  - Pedro M. Baeza
+  - Carlos Dauden
+  - Víctor Martínez
+- [DynApps](https://www.dynapps.be):
+  - Raf Ven \<<raf.ven@dynapps.be>\>
+- Marçal Isern \<<marsal.isern@qubiq.es>Bert Van Groenendael\>
```
etobella: Marçal an Bert are two different people :wink:

etobella (approved):
LGTM, just a minor remark. We should add the merge in OpenUpgrade when we have it.

miguel-S73 (approved):

Reyes4711-S73 (approved):
LGTM

Reyes4711-S73:
@etobella @OCA/banking-maintainers Can you merge this PR?

etobella (approved):

MiquelRForgeFlow (approved):
> We should add the merge in OpenUpgrade when we have it.

I take note.

HaraldPanten (approved):
Merging, considering it has been validated by several users from different companies and another PSC

/ocabot merge nobump

pedrobaeza:
Not everything was removed on `account_payment_partner`: https://github.com/OCA/bank-payment/tree/19.0/account_payment_partner and I'm also suspecting that the commit history will be a mesh on the next migration. Someone checked it before merging?

bizzappdev:
@pedrobaeza Indeed, the folder structure was in place. The commit history gave the illusion that the module had been removed. We are working on the fix MR for the same. So it will not mess up. Might be, the merge commit needs to be squashed with a fixed commit. We are checking on the solution.

PR #1508, opened 2025-10-10
[17.0][ADD] ``account_banking_international_credit_transfer``

Camille0907:
Add a new module to manage international credit transfers.
This adds the possibility to add an Intermediary bank on a `res.partner.bank`. When generating the pain file for an International Credit Transfer payment method, the intermediary bank account info will be added in `CstmrCdtTrfInitn > PmtInf > CdtTrfTxInf > IntrmyAgt1 > FinInstnId`
For now the module only supports pain.001.001.03

A lot of the code is very close from the module ``account_banking_sepa_credit_transfer``: since a lot of copy/paste have been done, we've kept the Copyright of the original module

Also rename `finalize_sepa_file_creation into finalize_pain_file_creation`. Indeed, this method can also be used for non-sepa pain files


yankinmax (approved):
LGTM

ivantodorovich (approved):

vvrossem (approved):
LG

grindtildeath (commented):
Since a lot of code comes from account_banking_sepa_credit_transfer, I would recommend to add a note in the roadmap, to either:
- Extract common parts in a dedicated module
- Add proper hooks in account_banking_sepa_credit_transfer and depend on that module here

Also I'd suggest to set dev status to alpha to allow such changes in the future before we merge this.

grindtildeath (approved):

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -283,12 +283,12 @@ def _validate_xml(self, xml_string, gen_args):
             ) from None
         return True
 
-    def finalize_sepa_file_creation(self, xml_root, gen_args):
+    def finalize_pain_file_creation(self, xml_root, gen_args):
```
pedrobaeza, line 286: This renaming may be conflicting for those overriding it in third-party modules. Please add at least the other method for compatibility, and call to this one from the other.

pedrobaeza (commented):
Wasn't another way to do this instead of verbatim copy of the code?

PR #1513, opened 2025-10-15
[16.0][IMP] account_banking_mandate_contact: Make field company-dependent

pedrobaeza:

@Tecnativa TT58428

### DIFF
--- a/account_banking_mandate_contact/__manifest__.py
+++ b/account_banking_mandate_contact/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Account Banking Mandate Contact",
     "summary": "Assign specific banking mandates in contact level",
-    "version": "16.0.1.0.4",
+    "version": "16.0.1.1.0",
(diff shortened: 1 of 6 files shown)


carlos-lopez-tecnativa (changes requested):
I think it's necessary to update the code in the module [account_banking_mandate_sale_contact](https://github.com/OCA/bank-payment/blob/badf03a5f47a84013222ceef2f597c8fde20ab0e/account_banking_mandate_sale_contact/models/sale_order.py#L28-L36) to ensure that the field is taken from the company of the sale order, just like Odoo does https://github.com/odoo/odoo/blob/f86ebd3a1772dfd9b5b04062df2d58a1fcf1af79/addons/sale/models/sale_order.py#L369

pedrobaeza:
@carlos-lopez-tecnativa done

carlosdauden (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #1515, opened 2025-10-17
[18.0][FIX] account_banking_mandate: Make partner bank domain to work

pedrobaeza:
The old mechanism with the field domain and the lambda is not working anymore (it was hackish anyway), so let's use the modern approach of binary computed domain and apply such domain on the view.

@Tecnativa TT58529


carlos-lopez-tecnativa (approved):
LGTM

victoralmau (approved):

Issue #1507, opened 2025-10-07
Error when migrating to Odoo18 from Odoo17


Darumait:
When migrating from Odoo 17 to Odoo18 I'm getting the following error. I'm not quite sure it's an error of the migration script since i'm not an expert on them but it seems like it.

## Module
account_payment_partner

## Describe the bug
I'm not quite sure what it is so I will just add the logs, as I said it seems like there's something wrong in the migration script but since i'm not an expert on it i'm opening an issue:
2025-10-07 13:53:57,447 48 DEBUG darumait-test-24379418 OpenUpgrade: -1 rows affected after 0:00:00.000145 running ALTER TABLE res_partner ADD COLUMN IF NOT EXISTS supplier_payment_mode_id jsonb
2025-10-07 13:53:57,447 48 ERROR darumait-test-24379418 odoo.sql_db: bad query: b'\n        UPDATE res_partner SET supplier_payment_mode_id=ir_property_by_company.value\n        FROM (\n            SELECT\n            SPLIT_PART(res_id, \',\', 2)::integer res_id,\n            JSON_OBJECT_AGG(company_id, SPLIT_PART(value_reference, \',\', 2)::integer) AS "value"\n            FROM ir_property\n            WHERE\n            fields_id=17319 AND res_id IS NOT NULL\n            AND company_id IS NOT NULL\n            GROUP BY res_id\n        ) ir_property_by_company\n        WHERE res_partner.id=ir_property_by_company.res_id\n        '
ERROR: relation "ir_property" does not exist
LINE 7:             FROM ir_property
                         ^
2025-10-07 13:53:57,449 48 ERROR darumait-test-24379418 OpenUpgrade: Error after 0:00:00.001731 running 
        UPDATE res_partner SET supplier_payment_mode_id=ir_property_by_company.value
        FROM (
            SELECT
            SPLIT_PART(res_id, ',', 2)::integer res_id,
            JSON_OBJECT_AGG(company_id, SPLIT_PART(value_reference, ',', 2)::integer) AS "value"
            FROM ir_property
            WHERE
            fields_id=17319 AND res_id IS NOT NULL
            AND company_id IS NOT NULL
            GROUP BY res_id
        ) ir_property_by_company
        WHERE res_partner.id=ir_property_by_company.res_id
[...]

PR #1492, opened 2025-08-14
[18.0][FIX] account_payment_partner: missing migration scripts

remi-filament:
The fields customer_payment_mode_id and supplier_payment_mode_id on res_partner are company_dependent and therefore need to be migrated in v18.0 in case you come from a previous Odoo version.

This PR converts fields from ir_property.

It should be transparent if properties do not exist.

PR #1516, opened 2025-10-22
[IMP] account_payment_partner: Handle ir_property table rename in migrations

desdelinux:
In Odoo 18.0, the ir_property table is renamed to _ir_property when using the Enterprise Migration Service. This adds a pre-migration to create a temporary copy for compatibility with openupgrade_180.convert_company_dependent(), and cleans it up in post-migration.

Related: 

https://github.com/OCA/bank-payment/issues/1507
https://github.com/OCA/bank-payment/issues/1498

https://github.com/OCA/bank-payment/pull/1492

### DIFF
--- a/account_payment_partner/migrations/18.0.1.0.0/post-migration.py
+++ b/account_payment_partner/migrations/18.0.1.0.0/post-migration.py
@@ -12,3 +12,18 @@ def migrate(env, version):
     openupgrade_180.convert_company_dependent(
         env, "res.partner", "customer_payment_mode_id"
     )
+    # Drop temporary ir_property table created in pre-migration
+    cr = env.cr
+    cr.execute(
+        """
+        SELECT EXISTS (
+            SELECT FROM pg_tables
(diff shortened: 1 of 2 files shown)


desdelinux:
Hi @pedrobaeza 

What do you think of this solution?

cc @Darumait @matteoopenf @remi-filament

pedrobaeza:
Sorry, but no. We are not going to handle enterprise specific code that dirties everything and make this less readable (or even to know if it's going to work). And it also has impact in the performance terms. Compare your solution with the current one: https://github.com/OCA/bank-payment/blob/18.0/account_payment_partner/migrations/18.0.1.0.0/post-migration.py

Put that scripts in your side.

pedrobaeza:
I think your best bet is to rename the table `_ir_property` to `ir_property` when Odoo SA delivers you the DB, because this will be recurrent, like in https://github.com/OCA/sale-workflow/blob/18.0/sale_order_type/migrations/18.0.1.0.0/post-migration.py.

PR #1518, opened 2025-10-30
[17.0][IMP] account_banking_mandate_contact: Make field company-dependent

pedrobaeza:
Forward-port of #1513 

@Tecnativa TT58428


pedrobaeza:
Let's proceed with the merge for finishing the fw-port chain:

/ocabot merge patch

PR #1519, opened 2025-10-30
[17.0][FIX] account_banking_mandate: Make partner bank domain to work

pedrobaeza:
Backport of #1515 

The old mechanism with the field domain and the lambda is not working anymore (it was hackish anyway), so let's use the modern approach of binary computed domain and apply such domain on the view.

@Tecnativa TT58529


pedrobaeza:
Merging this one as well for making it work in this version the same as the patch in 18.

/ocabot merge patch

PR #1520, opened 2025-10-30
[18.0][IMP] account_banking_mandate_contact: Make field company-dependent

pedrobaeza:
Forward-port of #1513 and #1518

@Tecnativa TT58428

### DIFF
--- a/account_banking_mandate_contact/__manifest__.py
+++ b/account_banking_mandate_contact/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Account Banking Mandate Contact",
     "summary": "Assign specific banking mandates in contact level",
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.2.0",
     "development_status": "Production/Stable",
     "category": "Banking addons",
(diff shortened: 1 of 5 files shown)


Tisho99:
Hello @pedrobaeza 

I think this PR might have some problems related to the migration of the contact_mandate_id field

In a testing enviroment with the last version of the module, the update of account_banking_mandate_contact causes me an error. If i go back to the "18.0.1.1.0" version, it works as normally.

The error is "psycopg2.errors.CannotCoerce: cannot cast type integer to jsonb", and it seems to be related with the new way to store company_dependent_fields in v18, storing it as a jsonb field instead of doing it in another table. https://www.odoo.com/documentation/saas-18.2/developer/reference/backend/orm.html#odoo.fields.Field

I'll try to find a solution if no one does it first, but it will take me quite a while.

Thank you for reading

EDIT: Uninstalling and installing the module worked for me

pedrobaeza:
OK, I see... you have to rename the column in pre, and re-rename it in post and do the conversion. Can you please change the migration scripts according this?

PR #1528, opened 2025-11-11
[UPD] account_banking_pain_base: sepa hybrid mode

ntsirintanis:


thomaspaulb:
So to discuss about the methodology (source: [this doc](https://www.europeanpaymentscouncil.eu/sites/default/files/kb/file/2025-10/EPC153-22%20v2.1%20EPC%20guidance%20document%20-%20Provision%20of%20Addresses%20under%20the%20EPC%20Payment%20Schemes.pdf)):

1. What the current version 16.0 module does, before this PR, is to create an **unstructured address block** with no `TwnNm` and just a `Ctry` and two `AdrLine` blocks. I understand that this has become forbidden.
2. In the new situation we can choose the **structured address block**, however, unless OCA's `base_address_extended` is installed and all fields of the partner duly filled, it's pretty hard to get this right
3. So, it is more realistic to aim to generate the **hybrid address block**, where `PstCd` is optional, `TwnNm` and `Ctry` are mandatory and the address is complemented by two 70-character-max `AdrLine` blocks.

What we have implemented now is that when the "enforce hybrid mode" checkbox is checked, it adds TwnNm and Ctry and does not include any AdrLine blocks.

Perhaps what's even better is that we always assume hybrid mode and:

- Always add TwnNm, eg if partner.city is empty, raise an error or fully omit the address block
- Always try adding PstCd from the "zip" field in Odoo
- Still add AdrLine blocks, but don't repeat any information in it eg don't put post code and city there

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -529,51 +529,89 @@ def generate_party_acc_number(
...
 
+    @api.model
+    def _get_bank_record(self):
+        """Retrieve bank"""
+        bank = None
+        bank_id = self.env.context.get("export_bank_id")
+        if bank_id:
+            bank = self.env["res.bank"].browse(bank_id)
+        else:
+            po_id = self.env.context.get("payment_order_id")
```
Kiplangatdan: @ntsirintanis I can't find anywhere where the `payment_order_id` is passed to the context. 

In anycase, `self` should get you the current payment order. 
So `bank = po.journal_id.bank_account_id.bank_id` should work also.

Kiplangatdan (changes requested):

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -529,51 +529,89 @@ def generate_party_acc_number(
...
+    def generate_payment_file(self):
+        """Inject bank_id in context"""
+        self.ensure_one()
+        bank = (
+            self.company_partner_bank_id.bank_id
+            or self.journal_id.bank_account_id.bank_id
+        )
+        if bank:
+            self = self.with_context(export_bank_id=bank.id)
+        return super().generate_payment_file()
```
Kiplangatdan: @ntsirintanis This is not reached when using the [account_banking_sepa_credit_transfer](https://github.com/OCA/bank-payment/blob/16.0/account_banking_sepa_credit_transfer/models/account_payment_order.py#L17) or the [account_banking_sepa_direct_debit](https://github.com/OCA/bank-payment/blob/16.0/account_banking_sepa_direct_debit/models/account_payment_order.py#L17).

It would be better to inject the `bank_id` to `open2generated()` instead.

ntsirintanis: Note that this is an api.model so self may or may not be an empty recordset

mihien:
@ntsirintanis Is this PR doing the same as #1550 but in 16.0, and is it still active? I'm guessing it's not a port since it pre-dates the other PR

ntsirintanis:
> @ntsirintanis Is this PR doing the same as #1550 but in 16.0, and is it still active? I'm guessing it's not a port since it pre-dates the other PR

@mihien Yeah, this was developed in parallel for versions 14, 16, and 18 (https://github.com/OCA/bank-payment/pull/1553)

marcos-mendez (changes requested):
**Automated Review -- Tests Failed**

### 1. **Root Cause**
The test failure is due to a missing Python dependency `unidecode`, required by the `account_banking_pain_base` module, which prevents the module from being installed and loaded during the test run.

---

### 2. **Suggested Fix**
Add `unidecode` to the `external_dependencies` list in the `__manifest__.py` of the `account_banking_pain_base` module:

```python
# In account_banking_pain_base/__manifest__.py
"external_dependencies": {
    "python": ["unidecode"],
},
```

This ensures that the dependency is checked and installed before the module is loaded.

---

### 3. **Additional Code Issues**
- **In `generate_address_block`**: The condition `if not partner.country_id: return True` is a valid early return, but the logic for handling `pain_flavor` starts with `pain.001.001.` and `pain.008.001.` — ensure this logic is consistent across all SEPA pain flavors. Consider adding a helper method or a more robust check for supported pain flavors.
- **In `open2generated`**: The method `self_with_context_bank` is constructed using `with_context`, but `self` is not reassigned, which could be confusing. It's better to assign the result back to `self` or use a clearer variable name like `self_contextualized`.
  
  ```python
  # Instead of:
  self_with_context_bank = self.with_context(export_bank_id=bank.id) if bank else self
  return super(AccountPaymentOrder, self_with_context_bank).open2generated()

  # Prefer:
  self_contextualized = self.with_context(export_bank_id=bank.id) if bank else self
  return super(AccountPaymentOrder, self_contextualized).open2generated()
  ```

---

### 4. **Test Improvements**
- Add tests for different `pain_flavor` values (e.g., `pain.001.003.03`, `pain.008.001.02`) to ensure address block generation behaves correctly.
- Add a test case to verify that `enforce_sepa_hybrid_mode = False` does **not** alter the address block structure (i.e., full address should be preserved).
- Use `SavepointCase` instead of `TransactionCase` if tests are expected to be isolated and run in parallel.
- Add a test to assert that `export_bank_id` is **not** passed in context when no bank is present in the payment order.

```python
# Example test improvement:
def test_open2generated_no_bank(self):
    """Test that open2generated does not inject export_bank_id if no bank is set."""
    self.po.journal_id.bank_account_id.bank_id = False
    self.po.open2generated()
    # Assert that no bank context was set
    self.assertNotIn("export_bank_id", self.po.env.context)
```

---

### Summary
The main issue is a missing dependency. Fix by adding `unidecode` to `external_dependencies`. Also, improve code clarity in `open2generated` and enhance tests to cover hybrid mode and different pain flavors.

---
### ⏰ PR Aging Alert

This PR by @ntsirintanis has been open for **122 days** (4 months).

> _Every ignored PR is a contributor who might not come back. Review time matters. ([OCA Aging Report](https://git.pop.coop/pop/oca-reviewer))_

---
### Reciprocal Review Request



Hi everyone! I found some test failures on this PR and left detailed feedback above. I am happy to discuss or help debug. In the meantime, if any of you get a chance, I would appreciate a look at my open PR(s):

**My open PRs across OCA:**
- [hr-attendance#262 [16.0][ADD] Hr_attendance_idsecure: iDSecure (ControliD) attendance integration](https://github.com/OCA/hr-attendance/pull/262)
- [stock-logistics-workflow#2276 [16.0][ADD] stock_move_line_devaluation](https://github.com/OCA/stock-logistics-workflow/pull/2276)
- [stock-logistics-workflow#2275 [16.0][ADD] Stock move line analytic account](https://github.com/OCA/stock-logistics-workflow/pull/2275)
- [stock-logistics-workflow#2268 [16.0][ADD] stock_move_line_picking_partner](https://github.com/OCA/stock-logistics-workflow/pull/2268)
- [purchase-workflow#2694 [16.0][IMP]Purchase workflow added to review state & exception fix](https://github.com/OCA/purchase-workflow/pull/2694)

Reviewing each other's work helps the whole community move forward. Thank you!

---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

thomaspaulb:
@ntsirintanis We can close this meanwhile

PR #1530, opened 2025-11-13
[19.0][FIX] account_payment_partner: Removed account_payment_partner

bizzappdev:


pedrobaeza:
Has the translations been integrated in the other module?

bizzappdev:
@pedrobaeza  No, we haven’t integrated the translations into the other module.
We’re not entirely sure about the process for handling or updating the .pot files in OCA modules, as this seems to be managed by the OCABOT itself. Could you please point us to the proper procedure or direct us to the relevant documentation on how to generate and integrate the .pot/.po files for this case?

pedrobaeza:
This is a special case not covered by the bots, as merging two modules without "merging" their .po leads to all the missing translations coming from `account_payment_partner`. There's a tool for merging .po called [pomerge](https://pypi.org/project/pomerge/). You can use it for not losing the existing translations.

PR #1532, opened 2025-11-17
[18.0][IMP] account_payment_mode:  Add codeview option to the field note to allow direct HTML edit…

samirGuesmi:
Add codeview option to the field note to allow direct HTML editing.

### DIFF
--- a/account_payment_mode/views/account_payment_mode.xml
+++ b/account_payment_mode/views/account_payment_mode.xml
@@ -35,7 +35,12 @@
                         />
                     </group>
                     <group string="Note" name="note">
-                        <field name="note" nolabel="1" colspan="2" />
+                        <field
+                            name="note"
+                            options="{'codeview': true}"
+                            nolabel="1"
+                            colspan="2"
+                        />
                     </group>
                 </sheet>
             </form>


pedrobaeza:
If you want that, you should enable debug mode and do HTML debugging, but not enabling the code by default.

samirGuesmi:
hello @pedrobaeza thanks for your quick help.
 
I'm referring to the code view toggle button in the editor toolbar, which allows switching between WYSIWYG and HTML source views. I think this might be different from the debug mode you mentioned?

image

pedrobaeza:
Uhm, that was what I was referring to... Is it now activated that way? And everyone can see it? Isn't that counter-productive?

samirGuesmi:
Yes It seems that now its activated by adding {'codeview': true} + debug mode. So the code view toggle button won’t be visible in normal mode.

https://github.com/odoo/odoo/blob/7272b538fc348eb94758da071cd2ed2f005fbf54/addons/mail/views/mail_template_views.xml#L53

pedrobaeza (approved):
OK, thanks for the insights.

qgroulard (approved):

pedrobaeza:
Please fw-port it to 19

PR #1534, opened 2025-11-27
[18.0][FIX] account_payment_sale: We need to set the value if we want to use it as key in grouping

etobella:
That might happen if we use certain python versions, where None cannot be compared with integers (some are defined and some not)

@pedrobaeza

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -25,8 +25,9 @@ class SaleOrder(models.Model):
             ).customer_payment_mode_id
 
     def _get_payment_mode_vals(self, vals):
+        # We need to set it if we want to use it as invoice key
+        vals["payment_mode_id"] = self.payment_mode_id.id
         if self.payment_mode_id:
-            vals["payment_mode_id"] = self.payment_mode_id.id
             if (
                 self.payment_mode_id.bank_account_link == "fixed"
                 and self.payment_mode_id.payment_method_id.code == "manual"
--- a/account_payment_sale/tests/test_sale_order.py
+++ b/account_payment_sale/tests/test_sale_order.py
@@ -131,3 +131,12 @@ class TestSaleOrder(CommonTestCase):
         other_company = self.env["res.company"].create({"name": "other company"})
         order.company_id = other_company
         self.assertFalse(order.payment_mode_id)
+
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Which Python version? And False can be compared always?

etobella:
At least it happens with Python 3.12, False is considered 0 and works correctly.

I am 100% sure if it is a problem of python version, but the issue is that int(None) --> Type Error and int(False) --> 0. Maybe it is a bug that has existed for several years and noone noticed. For this reason I thought  it was on the python version, as we detected the issue after an upgrade.

If it affects other versions, it might affect several odoo versions.

etobella:
After checking with some customers, the error does not depend on python version... 

The error only appears if we have 2 records with the same company, customer, same fiscal position and one with payment_mode and one without it. This happens because we are comparing tuples and comparison is done in order:

```
(1,2,3,1) < (2,2,3,None) # Different company
(1,2,3,1) < (1,3,3,None) # Different partner
(1,2,3,1) < (1,2,4,None) # Different fiscal position
(1,2,3,1) < (1,2,3,None)  --> Type error because int(None) cannot be defined.
```

This is a rare scenario, but might happen

pedrobaeza:
OK, thanks for clarifying. Please document that on the commit message and the test should reproduce that scenario, although I would say it already does.

account_payment_sale/models/sale_order.py
```diff
@@ -25,8 +25,9 @@ def _compute_payment_mode(self):
             ).customer_payment_mode_id
 
     def _get_payment_mode_vals(self, vals):
+        # We need to set it if we want to use it as invoice key
+        vals["payment_mode_id"] = self.payment_mode_id.id
```
luisDIXMIT, line 29: Can this raise an exception such as ‘NoneType object has no attribute id’?

etobella: no, because bools and integers can be compared and self.payment_mode_id.id returns an integer or False if it is false.

luisDIXMIT (approved):
LGTM!

etobella:
@pedrobaeza Actually, this error probably happens in all versions :cry:

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
To be fw-ported to 19.

PR #1536, opened 2025-12-18
[19.0] [FIX] account_payment_sale: We need to set the value if we want to use it as key in grouping

luisDIXMIT:
Forward-port of #1534


pedrobaeza (approved):
Fast-tracking, as it's the same as 18, and has a regression test.

/ocabot merge patch

PR #1537, opened 2025-12-23
[17.0][FIX] account_payment_order: Don't try to go to draft if already there

pedrobaeza:
When cancelling a payment order in state "Confirmed", the linked payments are yet in draft, so cancelling in that moment the payment order gives an error that only posted moves can be reset to draft.

The solution is to filter which payments to move to draft.

@Tecnativa TT59926


AaronHForgeFlow (approved):
It makes sense

sergio-teruel (approved):

PR #1550, opened 2026-01-28
[14.0][IMP] account_banking_pain_base: Support PAIN.001.001.09

ntsirintanis:
Added support for PAIN.001.001.09 address handling.

- [x] Bare mode (only CtryId/TwnNm)
- [x] Hybrid mode (CtryId/TwnNm/AdrLine/AdrLine)
- [x] ~~Structured mode (CtryId/TwnNm/....all sorts of details like Street, Housenumber, but not AdrLine)~~ Handled in followup PR

Needed to satisfy new [EPC 2.1](https://www.europeanpaymentscouncil.eu/document-library/guidance-documents/epc-guidance-document-provision-addresses-under-epc-payment) rules that forbid using the unstructured address format.


thomaspaulb:
@ntsirintanis see review comments in https://github.com/Therp/bank-payment/pull/3

account_banking_pain_base/models/res_bank.py
```diff
@@ -12,6 +12,12 @@
 class ResBank(models.Model):
     _inherit = "res.bank"
 
+    enforce_sepa_hybrid_mode = fields.Boolean(
+        string="Enforce SEPA Hybrid Mode",
+        help="Only applies to PAIN 09 exports. When enabled, the exporter will "
+        "generate a hybrid postal address (PstCd/TwnNm/Ctry + optional AdrLine).",
+    )
+
```
thomaspaulb: copying here my before remark that this makes more sense on payment method.

Provided that payment method is something you can select per bank, so that if you require to have two banks treated differently, you can duplicate the payment method, differentiate, and select one payment method on one bank and one on another.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -544,33 +612,60 @@ def generate_address_block(self, parent_node, partner, gen_args):
                     70,
                     gen_args=gen_args,
                 )
-            if (
-                gen_args.get("pain_flavor").startswith("pain.001.001.")
-                or gen_args.get("pain_flavor").startswith("pain.008.001.")
-            ) and (partner.zip or partner.city):
+            if partner.street2:
```
thomaspaulb: During testing we noticed that our bank, while on paper they should accept `AdrLine` entries in the hybrid address, they actually don't. So maybe we can add a setting, default False, that says "Include address", and that's only visible when Hybrid mode is switched on. Then we only include street and street2 if that setting is on, and we have it off by default. The resulting CtryId/TwnNm combination and nothing else seems to me like the safest default. That's technically not even strictly hybrid mode, it's the common denominator between structured and hybrid.

thomaspaulb (changes requested):

account_banking_pain_base/models/account_payment_method.py
thomaspaulb: It's slightly annoying that it's not possible to duplicate a SEPA payment method because if this gets copied, the unique constraint triggers. Does `copy=False` help?

account_banking_pain_base/models/account_payment_method.py
```diff
@@ -17,6 +17,15 @@ class AccountPaymentMethod(models.Model):
...
     warn_not_sepa = fields.Boolean(string="Warn If Not SEPA")
+    enforce_sepa_hybrid_mode = fields.Boolean(
+        default=True,
+        help="Generate hybrid address blocks for PAIN.001.001.09 ",
+    )
+    sepa_hybrid_include_address = fields.Boolean(
+        default=False,
+        string="Include Address",
+        help="",
+    )
```
thomaspaulb, line 36: Maybe this is nitpicking, but I now think that maybe this would have to be switched around.

First there is "Include Address", which is unchecked by default, which results in Ctry and TwnNm.

Then once people do check it, there are two options to include it:

1. Hybrid mode, so with the AdrLines
2. Structured mode, so in the way that Coopiteasy did in the other PR

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -525,8 +529,46 @@ def generate_party_acc_number(
     @api.model
     def generate_address_block(self, parent_node, partner, gen_args):
         """Generate the piece of the XML corresponding to PstlAdr"""
-        if partner.country_id:
+        if not partner.country_id:
+            return True
```
thomaspaulb, line 533: So a missing country is silently waved away but a missing city throws a UserError? Hmm not ideal, also not blocking

gjotten (approved):
Functional test; this file's being accepted by our bank compared to the previous version which is being rejected.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -525,8 +529,46 @@ def generate_party_acc_number(
...
+        pain_flavor = (gen_args.get("pain_flavor", "")).strip()
+        payment_method = self.payment_mode_id.payment_method_id
+        if pain_flavor == "pain.001.001.09" and payment_method.enforce_sepa_hybrid_mode:
```
thomaspaulb: Actually `and payment_method.enforce_sepa_hybrid_mode` makes no sense.

If people select 09 as a format, but the checkbox is unchecked, it would emit the unstructured address block, which we know to be invalid.

Maybe 09 should always emit Ctry/TwnNm and then we use the hybrid mode checkbox to differentiate between adding AdrLine blocks or... well, or not, and then we leave it to the coopiteasy people to add the code for the "or not" - all the possible details there are to add in Structured mode.

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -525,8 +529,46 @@ def generate_party_acc_number(
...
+                        "Partner missing City: %(partner)s. "
+                        "Please set a City or choose an older PAIN format."
+                    )
+                    % {
+                        "flavor": pain_flavor,
+                        "partner": partner.display_name,
+                    }
+                )
             postal_address = etree.SubElement(parent_node, "PstlAdr")
+            # Order matters for pain.001.001.09
```
thomaspaulb: order matters for any pain format

account_banking_sepa_credit_transfer/models/account_payment_method.py
```diff
@@ -10,16 +10,21 @@ class AccountPaymentMethod(models.Model):
...
         selection_add=[
             ("pain.001.001.02", "pain.001.001.02"),
-            ("pain.001.001.03", "pain.001.001.03 (recommended for credit transfer)"),
+            (
+                "pain.001.001.03",
+                "pain.001.001.03 (recommended for Germany credit transfer)",
+            ),
             ("pain.001.001.04", "pain.001.001.04"),
             ("pain.001.001.05", "pain.001.001.05"),
+            ("pain.001.001.09", "pain.001.001.09 (recommended for credit transfer)"),
```
thomaspaulb, line 20: We would maybe have to add a README line under.. well configuration? To note that the new version requires the format to be changed.

thomaspaulb:
I would think it's ready to come out of Draft.

thomaspaulb (approved):

account_banking_pain_base/models/account_payment_method.py
```diff
@@ -17,6 +17,21 @@ class AccountPaymentMethod(models.Model):
         "characters are used in the generated PAIN file.",
     )
     warn_not_sepa = fields.Boolean(string="Warn If Not SEPA")
+    # get rid of booleans, add selection
+    sepa_pain09_address_mode = fields.Selection(
```
pedrobaeza, line 22: Shouldn't this be required?

account_banking_pain_base/tests/test_context_propagation.py
pedrobaeza, line 1: Why this weird file name?

account_banking_sepa_credit_transfer/models/account_payment_method.py
```diff
@@ -10,16 +10,21 @@ class AccountPaymentMethod(models.Model):
     pain_version = fields.Selection(
         selection_add=[
             ("pain.001.001.02", "pain.001.001.02"),
-            ("pain.001.001.03", "pain.001.001.03 (recommended for credit transfer)"),
+            (
+                "pain.001.001.03",
+                "pain.001.001.03 (recommended for Germany credit transfer)",
```
pedrobaeza: Germany is still with the old one?

account_banking_sepa_credit_transfer/models/account_payment_method.py
pedrobaeza, line 20: Shouldn't migration scripts should be added for switching?

account_banking_sepa_credit_transfer/models/account_payment_method.py
thomaspaulb: We had assumed that from the other PR, @mihien do you know more about this?

account_banking_sepa_credit_transfer/models/account_payment_method.py
thomaspaulb: I'm in doubt a bit about the migration script actually. I don't really have data on a lot of banks in Europe. What I know is that:

- Some banks rigidly don't accept the old format anymore
- Some banks, while doing so, misimplement the standard
- Rumor has it that at least in some countries the implementation is delayed

In this situation I wouldn't be able to predict if migrating all current payment methods at "03" to "09" will be helpful (eg people now have broken SEPA and this fixes it) or if it will actually break things (their bank/country was not ready yet and the migration to the new format has broken it)

Do you have any more info about the situation in Spain?

thomaspaulb: Sources that I can find say that there is a transition period between Nov2025-Nov2026 where both formats should still be accepted, though in practice some banks already started rejecting the unstructured address format. From Nov2026 onwards, the old format will not be accepted anymore.

If this is indeed the situation for all countries involved then a migration script makes sense.

account_banking_sepa_credit_transfer/models/account_payment_method.py
thomaspaulb: A quick search online does not reveal a different picture for Germany than for other countries. Nov2025-Nov2026 as the transition period until by Nov2026 everyone is expected to have done the switch and the old format will cease to be accepted. [Deutche Bank](https://corporates.db.com/files/documents/in-focus/focus-topics/iso20022/Deutsche-Bank-ISO-20022-Changes-in-Payments-and-Account-Statements.pdf) [Commerzbank](https://www.commerzbank.de/service/informationen-zur-abschaltung-von-sepa-altformaten-bei-ebics/) 

"Migration to pain.001.001.09 is required by November 2026, when unstructured address formats will no longer be supported."

account_banking_sepa_credit_transfer/models/account_payment_method.py
ntsirintanis: For now I propose no automatic migration: existing payment methods keep their configured pain versions, and users can opt-in by switching to 09 (which we mark as recommended anyway). I think it would be better to just add a note under Configuration explaining all these, and when users should switch.

account_banking_pain_base/tests/test_context_propagation.py
thomaspaulb: @ntsirintanis I see you added one with a more proper name but shouldn't the old one then be removed?

account_banking_sepa_credit_transfer/models/account_payment_method.py
thomaspaulb: @ntsirintanis For now let's remove the comment about Germany and just leave "pain.001.001.03". @mihien told me this came from the old Alexis' PR so maybe it is just outdated info that was valid at the time that PR was made

ntsirintanis:
@thomaspaulb there we go

thomaspaulb (approved):

account_banking_pain_base/models/account_payment_order.py
```diff
@@ -524,9 +528,39 @@ def generate_party_acc_number(
...
-        """Generate the piece of the XML corresponding to PstlAdr"""
-        if partner.country_id:
+        if not partner.country_id:
+            return True
+        pain_flavor = (gen_args.get("pain_flavor", "")).strip()
+        payment_method = self.payment_mode_id.payment_method_id
+        address_mode = payment_method.sepa_pain09_address_mode
+        if pain_flavor == "pain.001.001.09":
+            if not partner.city:
+                raise UserError(
```
jguenat, line 539: Thanks for the hard work :)
Is it possible to use startswith("pain.001.001.09")?
So that other localisation like pain.001.001.09.ch.03 can benefit from the address implementation?

ntsirintanis: Done @jguenat , thanks for spotting this

account_banking_sepa_credit_transfer/models/account_payment_method.py
pedrobaeza: I think we should do it automatically for 2 reasons:

- Most of the users are not aware of these internals. They will only see the error in the bank when they reject the file.
- Even for us, it reduces the maintenance to make for doing the change in all the instances of our customers.

It's easier to handle errors because a bank don't support structured address than the contrary.

thomaspaulb: Migration script was added - just out of curiosity, let's say we merge this as `minor`, will the bot be smart enough to update the version in the migration script?

pedrobaeza: Nop. That's something to be synced manually.

account_banking_pain_base/models/account_payment_method.py
```diff
@@ -8,7 +9,7 @@
 class AccountPaymentMethod(models.Model):
     _inherit = "account.payment.method"
 
-    pain_version = fields.Selection([], string="PAIN Version")
+    pain_version = fields.Selection([], string="PAIN Version", copy=False)
```
pedrobaeza, line 12: I don't think it's bad to copy this data.

account_banking_pain_base/readme/CONFIGURE.rst
```diff
@@ -10,3 +10,18 @@ If your country requires several identifiers (like Spain), you must:
...
    and "Initiating Party Issuer".
+
+#. When configuring a SEPA Credit Transfer payment method, you can choose the
+   PAIN format version to use.
+
+   Starting from this version, **PAIN.001.001.09** is available and is
+   *recommended for credit transfers*, as it complies with the newer EPC
+   requirements regarding postal address structure.
+
+   Existing payment methods are **not migrated automatically** to
```
pedrobaeza: This text is not correct now.

pedrobaeza (commented):
Does this only apply to SCT, not SDD?

account_banking_pain_base/models/account_payment_method.py
thomaspaulb: What was really annoying about this is that it's a unique-constraint field so if you want to duplicate it you cannot so you have to create and copy paste.

thomaspaulb:
Yes, not tested anything Sepa Direct Debit related.

account_banking_pain_base/models/account_payment_method.py
pedrobaeza: Ouch, then totally right.

ntsirintanis:
alright @pedrobaeza , please take another look

pedrobaeza:
Thanks for the changes. `[UPD]` is not a valid tag. It should be `[IMP]`: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

About SDD, then we are not handling it?

thomaspaulb:
SDD: This is in another module, `account_banking_sepa_direct_debit`. None of our customers uses this. Do they need to be done all at once or can that be the subject of a new PR?

pedrobaeza:
OK, as this touches the main one (`account_banking_pain_base`), it may be direct. You can add the ROADMAP to the module.

thomaspaulb (approved):

pedrobaeza (approved):
Thanks for working on this. Merging with nobump, as the version change is already included.

/ocabot merge nobump

PR #1552, opened 2026-02-04
[FIX] account_payment_order: wrong date set on payment.line

micheledic:
### Before this PR 
Clicking on "generate payments" , all payment lines have today as date 
### After this PR 
Clicking on "generate payments" , all payment lines have the date accordingly to date_prefered field


micheledic:
@OCA/banking-maintainers ping

micheledic:
@OCA/banking-maintainers ping

StefanRijnhart (commented):
What problem does this solve and how is the date 'wrong'? The date on the payment line is not leading for the requested date of the payment order segment. The logic to determine the requested date further down in the payment process already follows the logic you are substituting here (https://github.com/OCA/bank-payment/blob/17.0/account_payment_order/models/account_payment_order.py#L299-L305).

The field description of account.payment.line date "Payment Date" is unfortunately ambiguous, but might well refer to the date that the payment (line) was created, in which case using today's date is actually correct.

PR #1553, opened 2026-02-06
[18.0][MIG] account_banking_pain_base sepa hybrid extended for 09

ntsirintanis:
related: https://github.com/OCA/bank-payment/pull/1550


ntsirintanis:
@thomaspaulb

thomaspaulb:
@pedrobaeza Is the cherry-pick OK or do we actually need to jump through the hoop of the "OCA technical migration method" now that the 14.0 PR has become merged?

pedrobaeza (commented):
I think both commits can be squashed, and more when there are changes in the migration script that doesn't belong to migration itself (like `strip` thing, that sound like a refinement found during the migration).

thomaspaulb:
Ok, but do we need to rebase on 14.0 and follow the technical migration method?

pedrobaeza:
No, because this is not a full module, but a new feature. In fact, this shouldn't be qualified as migration, but a forward-port of a `[IMP]`, being itself `[IMP]`. Remove `[14.0]` from the commit message.

thomaspaulb:
Ah of course yes, don't know where my brain was with that question.

pedrobaeza:
Please squash both commits into one.

thomaspaulb (approved):

thomaspaulb:
@pedrobaeza Can you approve as well?

pedrobaeza (approved):

pedrobaeza:
@ntsirintanis the 17.0 and 19.0 are still pending, isn't it?

ntsirintanis:
> @ntsirintanis the 17.0 and 19.0 are still pending, isn't it?

yes, that's the case

pedrobaeza:
@ntsirintanis are you going to do such PRs?

ntsirintanis:
> @ntsirintanis are you going to do such PRs?

@pedrobaeza that's not in my immediate plans, and I expect that PRs will appear from other contributors and for  these versions soon-ish. If this does not happen, I'll revisit this myself

pedrobaeza:
OK, thanks for the confirmation. The problem not doing this soon, specially for intermediate versions, is to not have homogenization across versions, and more if this one is pushing a new PAIN version that won't work if migrating from 14+ to 17. 

But don't worry. You have done a lot. I'm doing the PRs now for completing the chain.

ntsirintanis:
> OK, thanks for the confirmation. The problem not doing this soon, specially for intermediate versions, is to not have homogenization across versions, and more if this one is pushing a new PAIN version that won't work if migrating from 14+ to 17.
> 
> But don't worry. You have done a lot. I'm doing the PRs now for completing the chain.

thanks @pedrobaeza for the explanation and your work. Just ping me when there is something reviewable and I'll review asap, in order to get these merged asap too then

PR #1564, opened 2026-02-26
[17.0][FIX] account_*: Adapt tests to odoo new banks flow

christian-ramos-tecnativa:
The commit https://github.com/odoo/odoo/commit/1794fce234735ed174599891435d4e2accc16324 contains the next changes summarized:
- SUPERUSER is not capable of make a bank account allowed (Unless is at module install), now the user needs any of the groups:
  - `account.group_validate_bank_account`
  - `base.group_system`
- Now the `partner_bank_id` can contain non allow_out_payment banks, but the error will raise on post

This PR adapt the tests to the new bank flow
@Tecnativa TT61155
@carlos-lopez-tecnativa @pilarvargas-tecnativa @pedrobaeza @victoralmau

### DIFF
--- a/account_payment_partner/tests/test_account_payment_partner.py
+++ b/account_payment_partner/tests/test_account_payment_partner.py
@@ -69,6 +69,12 @@ class TestAccountPaymentPartner(TransactionCase):
                 "bank_acc_number": "123456",
             }
         )
+        # Use admin user as the SUPERUSER is not allowed to
+        # modify allow_out_payment on bank accounts and this
+        # is needed to avoid error when posting the invoice
+        cls.journal_c1.bank_account_id.with_user(
+            cls.env.ref("base.user_admin")
+        ).allow_out_payment = True
 
         cls.journal_c2 = cls.journal_model.create(
             {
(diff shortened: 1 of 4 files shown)


christian-ramos-tecnativa:
Will add here the fixes for other modules from the repo

carlos-lopez-tecnativa (approved):

account_payment_partner/tests/test_account_payment_partner.py
```diff
@@ -247,7 +253,10 @@ def test_partner_id_changes_compute_partner_bank(self):
         move_form = Form(
             self.env["account.move"].with_context(default_move_type="out_invoice")
         )
-        self.assertFalse(move_form.partner_bank_id)
+        # The partner bank should be set as odoo made posible
```
pedrobaeza:
```suggestion
        # The partner bank should be set as Odoo made possible
```

account_payment_partner/tests/test_account_payment_partner.py
```diff
@@ -247,7 +253,10 @@ def test_partner_id_changes_compute_partner_bank(self):
...
+        # to assign not allow_out_payment bank account
+        # more info https://github.com/odoo/odoo/commit/1794fce234735ed174599891435d4e2accc16324
+        self.assertTrue(move_form.partner_bank_id)
```
pedrobaeza, line 259: Why is this and the next ones changing? We want to not have a bank account assigned. Maybe you need to change the parameter at company level, or test both.

christian-ramos-tecnativa: Because the bank_partner in this case YourCompany has the bank_ids created with demo data https://github.com/OCA/bank-payment/blob/17.0/account_payment_mode/demo/payment_demo.xml#L36, before the change those banks account were not allowed so none of them get selected now it takes de first one https://github.com/odoo/odoo/blob/17.0/addons/account/models/account_move.py#L897

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #1569, opened 2026-03-09
Draft:[18.0][IMP] account_payment_mode: build payment method views on account_payment_method_base

ntsirintanis:
account_payment_method_base now provides the base UI for
account.payment.method (menu, action, form, list, search).

Refactor account_payment_mode views to inherit from
account_payment_method_base while keeping the existing XML IDs
as compatibility anchors for downstream module


pedrobaeza (changes requested):
Changing the dependencies in a stable version is not reasonable, so better to simply keep as they are, and more with the minimum gain obtained.

thomaspaulb:
@pedrobaeza The underlying story here is that we:

1. Have our [SEPA PR](https://github.com/OCA/bank-payment/pull/1553) that adds some fields to the `account.payment` form view by inheriting `account_payment_mode.account_payment_method_form`
2. However, the db also has [account_payment_method_base](https://github.com/OCA/account-payment/blob/18.0/account_payment_method_base/views/account_payment_method.xml) installed, which adds its own menu with reference to another base view, that we do not extend, and so the SEPA fields do not appear in the form view

Besides this dependency change, are there any other ways of addressing that and making sure that the SEPA module works together both with `account_payment_mode` and with `account_banking_payment_base` ? Some kind of conditional perhaps?

pedrobaeza:
I don't know any compatibility trick. That's why we wanted to not make forks due to these reasons...

thomaspaulb:
Hmm. I would then still argue to update `account_payment_mode` in this way and make it a major (breaking) update, because of the extra dependency. Unless we can smoothen that with a migration script, but that seems openupgrade territory.

pedrobaeza:
Sorry, but no, as I don't want to install such module in all our 18.0 deployments, and you can't force everyone to do it just because you want to have both modules installed.

thomaspaulb:
Did some digging around, and seems that `account_payment_unece` and by inheritance some e-invoicing modules depend on `account_payment_method_base`. Seems a bit of a conundrum because it looks to me quite common for people to want to have e-invoicing as well as SEPA payments.

pedrobaeza:
All of them come from the same person that did the fork...

thomaspaulb:
Yes, but the fork is a different story, that was safely parked in another repo clearly marked `alternative` where we can now police for cross-dependencies.

The case for this particular module IMO seems solid, isn't it just good policy to do this if there is no base view defined in Odoo and you want multiple modules to be able to modify such view?

pedrobaeza:
This was derived of the fork, for "sharing some things" between both parts, but as the original modules were already merged at that moment, it was too late for that sharing.

thomaspaulb:
@pedrobaeza To address your concern about an extra module having to be installed, we went for a glue module instead: https://github.com/OCA/bank-payment/pull/1570

pedrobaeza:
Thanks for your consideration.

PR #1570, opened 2026-03-11
[18.0] [ADD] account_payment _method_base_mode

ntsirintanis:
**Overview**

This PR introduces a small glue module account_payment_method_base_mode to align the view hierarchy between account_payment_method_base (OCA/account-payment) and account_payment_mode (OCA/bank-payment).

In Odoo 18, account_payment_method_base introduces the canonical user interface for the model account.payment.method, including:

- menu entry
- action
- form view
- list view
- search view

However, account_payment_mode historically defined its own standalone views for the same model.

As a result, when both modules are installed, two competing base view definitions exist for account.payment.method.

The UI entry point provided by account_payment_method_base becomes the effective base, while downstream modules in bank-payment still inherit view XML IDs from account_payment_mode.

This results in an inconsistent view inheritance chain.

**Solution**

This PR adds a glue module that rewires the account_payment_mode views so they inherit from the base views provided by account_payment_method_base.

The module updates the following view relationships:

account_payment_mode.account_payment_method_form
account_payment_mode.account_payment_method_tree
account_payment_mode.account_payment_method_search

so that they inherit from:

account_payment_method_base.account_payment_method_form
account_payment_method_base.account_payment_method_list
account_payment_method_base.account_payment_method_search

This results in the following view hierarchy:

account_payment_method_base
        ↓
account_payment_mode
        ↓
downstream modules (pain, sepa, etc.)


ntsirintanis:
this requires a little bit more work, as I just saw some elements repeating etc

Kiplangatdan (approved):

account_payment_method_base_mode/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2026 Therp BV
+# <https://therp.nl>. # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
NL66278: Mixup of copyright line and license line.

account_payment_method_base_mode/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "Account Payment Method Base + Payment Mode Glue",
+    "summary": "Glue module for account_payment_mode, "
```
NL66278: Summary should fit on one line.

NL66278 (changes requested):
👍 Thanks for your contribution. Just fix manifest file and it should be good to go.

NL66278 (approved):
👍 LGTM and checked results on runbot

thomaspaulb:
@pedrobaeza Are you alright with this approach?

pedrobaeza:
Thanks, this sounds better, but the repository may not be the best, as the original module that produces this is `account_payment _method_base`, which is in OCA/account-payment, so I think it should be hosted there, don't you think?

ntsirintanis:
> Thanks, this sounds better, but the repository may not be the best, as the original module that produces this is `account_payment _method_base`, which is in OCA/account-payment, so I think it should be hosted there, don't you think?

@pedrobaeza I disagree, and here's why:  this new addon is not extending account_payment_method_base with new features. It is specifically adapting account_payment_mode so its payment_method views sit correctly on top of account_payment_method_base. account_payment_mode belongs in this repository, while account_payment_method_base is already a standalone base addon in account-payment.

pedrobaeza:
OK, no strong opinion on this, so let's merge it.

/ocabot merge nobump

PR #1573, opened 2026-04-08
[16.0][FIX] account_payment_order: allow selecting child partners on payment lines

giarve:
Remove the parent_id domain on partner_id (account.payment.line) and on partner_ids (account.payment.line.create wizard) so contacts/child partners can be freely picked when creating payment lines.

Some companies are just not billed on PO with the commercial entity, does not make sense to hold this restriction here which could be applied by view domain instead (easier to modify in Odoo use-cases).

### DIFF
--- a/account_payment_order/models/account_payment_line.py
+++ b/account_payment_order/models/account_payment_line.py
@@ -56,7 +56,6 @@ class AccountPaymentLine(models.Model):
         comodel_name="res.partner",
         string="Partner",
         required=True,
-        domain=[("parent_id", "=", False)],
         check_company=True,
     )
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
We shouldn't do this, as the journal items will only have parent contacts. If this is for something related with the employees and having defined a children contact for them, then you should switch your strategy in how to defined them.

giarve:
you are totally right and thank you for this. Turns out I have invalid data in the system due to parent_id reassignation after contact creation. Probably will fix it through SQL.


The system is indeed posting in journal always with the commercial entity regardless of the children billed

PR #1579, opened 2026-05-11
[18.0][IMP] account_banking_pain_base: Support PAIN.001.001.09

carlosdauden:
Forward port of https://github.com/OCA/bank-payment/pull/1550

@Tecnativa
ping @sergio-teruel @victoralmau


pedrobaeza (approved):
Please include the migration script for those already on 18.0

carlosdauden:
> Please include the migration script for those already on 18.0

Done

pedrobaeza (approved):

ntsirintanis (changes requested):
Thanks for the contribution. However the one commit in https://github.com/OCA/bank-payment/pull/1550 was not cherry-picked, instead it appears that the functionality was cped and a new commit was made. Please amend that.

pedrobaeza:
I'm now seeing that #1553 is already there, so I think we should switch to that one.

pedrobaeza:
Let's go with the other one, containing the same changes and being older.

PR #1580, opened 2026-05-12
[19.0][MIG] account_payment_partner: Migration to 19.0

kristian-domatix:
This PR migrates the `account_payment_partner` module to Odoo 19.0.

## Changes made

- Bumped version to `19.0.1.0.0`.
- Removed `migrations/` folder (no longer needed after migrating to a new major version).
- Replaced `from odoo import _` with `self.env._` / `cls.env._` (preferred by OCA in recent versions).
- Adapted two tests to changes in Odoo 19 core `_compute_partner_bank_id` behavior:
  - `test_onchange_payment_mode_id`: setting `bank_account_required=False` now triggers the compute which reassigns `partner_bank_id` based on core logic.
  - `test_refund_no_payment_mode_preserves_partner_bank`: the core no longer auto-selects the trusted bank on refunds without payment mode.

## Notes

If a different approach is preferred for the test adaptations (e.g., adapting the module logic instead of the tests), please advise.

All 22 module tests pass locally.


HaraldPanten:
This module is deprecated. Please, see the migration issue:

https://github.com/OCA/bank-payment/issues/1503

PR #1603, opened 2026-06-24
[19.0][FIX] account_payment_order: Don't try to go to draft if already there

pedrobaeza:
Forward-port of #1537 and #1538

When cancelling a payment order in state "Confirmed", the linked payments are yet in draft, so cancelling in that moment the payment order gives an error that only posted moves can be reset to draft.

The solution is to filter which payments to move to draft.

@Tecnativa TT59926


pedrobaeza:
Let's finish the fw-port chain.

/ocabot merge patch

PR #1486, opened 2025-07-17
[16.0][IMP] Optionally use the transfer journal specified in payment mode

enricostano:
This will allow to create the grouped moves in the same journal as the moves associated to the payments.

PR #1605, opened 2026-07-08
[16.0][FIX] account_payment_order_grouped_output: reconcile grouped payments when transfer account differs from outstanding account

eantones:
On 16.0 `reconcile_grouped_payments` looks up the payment lines to reconcile by `payment.outstanding_account_id` only. When the outstanding/efectos account is taken from the journal's outstanding-payment accounts and isn't that exact field, nothing matches, `reconcile()` does nothing, and the invoices stay in `in_payment` with the account left open.

Now it also matches against the journal's inbound/outbound outstanding accounts, like 17.0 and 18.0 do. `payment.outstanding_account_id` came in with #1486 for general-type transfer journals and I kept it, so both setups reconcile.

17.0/18.0 don't carry #1486, so they probably want the same union.

### DIFF
--- a/account_payment_order_grouped_output/models/account_payment_order.py
+++ b/account_payment_order_grouped_output/models/account_payment_order.py
@@ -78,8 +78,14 @@ class AccountPaymentOrder(models.Model):
     def reconcile_grouped_payments(self, move, payments):
         lines_to_rec = move.line_ids[:-1]
         for payment in payments:
+            journal = payment.journal_id
+            outstanding_accounts = (
+                journal._get_journal_inbound_outstanding_payment_accounts()
+                | journal._get_journal_outbound_outstanding_payment_accounts()
+                | payment.outstanding_account_id
+            )
             lines_to_rec += payment.move_id.line_ids.filtered(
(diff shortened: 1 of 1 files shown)


eantones:
The failing `pre-commit` check is unrelated to this PR. CI runs `pre-commit run --all-files`, so it also lints translations in other modules, and it trips on a pre-existing issue in [`account_banking_pain_base/i18n/gl.po`](https://github.com/OCA/bank-payment/blob/0bbe5610370d71a9a8259fe89019baafc3ea34a1/account_banking_pain_base/i18n/gl.po#L907): a Python format placeholder was translated (`{sepa_pattern}` → `{saiba_pattern}`), which `oca-checks-po` rejects (`po-python-parse-format`). It came in via [`ef7ff58`](https://github.com/OCA/bank-payment/commit/ef7ff58192ae2f2a33403fac96b73a60fa3f4894) (Weblate, 2026-05-23) and currently breaks pre-commit on `16.0` for any fresh run. This PR only touches `account_payment_order_grouped_output`; the test jobs pass.

PR #1607, opened 2026-07-21
[15.0][FIX] account_payment_order: unset journal_id when it is not in allowed_journal_ids

carlos-lopez-tecnativa:
Create two payment modes with different allowed bank journals: PM1 with Journal 1 and Journal 2.
PM2 with Journal 3 and Journal 4.
Create a payment order and select PM1 and Journal 1. Change the payment mode to PM2.

Before this commit, Journal 1 remained selected even though it was not available for PM2. After this commit, the journal is cleared when it is not available for the selected payment mode, forcing the user to select a valid journal.

TT63778
@Tecnativa @pedrobaeza @carlosdauden @sergio-teruel @victoralmau could you please review this?


victoralmau (approved):

sergio-teruel (approved):

pedrobaeza:
I think a copier template update is needed.

carlos-lopez-tecnativa:
> I think a copier template update is needed.

Done, but the error was caused by corrupted translations, which I have fixed. Should I fix it here, or does it need to be changed in Weblate?

pedrobaeza (approved):
/ocabot merge patch

PR #1608, opened 2026-07-22
[17.0][FIX] account_payment_sale: We need to set the value if we want to us…

giarve:
…e it as key in grouping

backport https://github.com/OCA/bank-payment/pull/1534

The error only appears if we have 2 records with the same company, customer, same fiscal position and one with payment_mode and one without it. This happens because we are comparing tuples and comparison is done in order:

(1,2,3,1) < (2,2,3,None) # Different company
(1,2,3,1) < (1,3,3,None) # Different partner
(1,2,3,1) < (1,2,4,None) # Different fiscal position (1,2,3,1) < (1,2,3,None)  --> Type error because int(None) cannot be defined.

This is a rare scenario, but might happen

### DIFF
--- a/account_payment_sale/models/sale_order.py
+++ b/account_payment_sale/models/sale_order.py
@@ -25,8 +25,9 @@ class SaleOrder(models.Model):
             ).customer_payment_mode_id
 
     def _get_payment_mode_vals(self, vals):
+        # We need to set it if we want to use it as invoice key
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Let's fast-track it as already merged in upper versions

/ocabot merge patch

PR #1613, opened 2026-07-27
[19.0][IMP] account_banking_sepa_direct_debit: SEPA validation message

adrip-s73:
SEPA Direct Debit validation errors used unsupported {field} placeholders, causing literal placeholders to appear instead of the actual mandate, customer, line reference, or date values

image

### DIFF
--- a/account_banking_sepa_direct_debit/i18n/account_banking_sepa_direct_debit.pot
+++ b/account_banking_sepa_direct_debit/i18n/account_banking_sepa_direct_debit.pot
@@ -285,8 +285,8 @@ msgstr ""
 #. odoo-python
 #: code:addons/account_banking_sepa_direct_debit/models/account_payment_line.py:0
 msgid ""
-"Missing SEPA Direct Debit mandate on the line with partner {partner_name} "
-"(reference {reference})."
+"Missing SEPA Direct Debit mandate on the line with partner %(partner_name)s "
+"(reference %(reference)s)."
 msgstr ""
 
 #. module: account_banking_sepa_direct_debit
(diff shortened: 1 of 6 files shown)


Reyes4711-S73 (approved):
LGTM

pablo-cort-s73 (approved):
LGTM

account_banking_sepa_direct_debit/tests/test_sdd.py
```diff
@@ -376,6 +377,68 @@ def create_invoice(self, partner_id, mandate, price_unit, inv_type="out_invoice"
 
 
 class TestSDD(TestSDDBase):
+    def setUp(self):
```
pedrobaeza: This smells to AI. Please don't add indiscriminate tests, and more to the base test, which they will be repeated across inherited tests. Check if the exception is already triggered in other place of the tests, and if not, reuse an existing test or create just one for doing it, instead of having five new methods.

pedrobaeza (changes requested):

account_banking_sepa_direct_debit/tests/test_sdd.py
adrip-s73: @pedrobaeza okey done !!

pedrobaeza (commented):
It's still done on the base test, and creating all the records, while an existing one can be reused. For getting this, please don't add tests (and it's better to not use AI if it's going to generate slop code).

pedrobaeza (approved):
OK, let's continue without extra tests. FYI, I was talking to add the test in https://github.com/OCA/bank-payment/blob/19.0/account_banking_sepa_direct_debit/tests/test_sdd.py#L378, not in the base

/ocabot merge patch

PR #1614, opened 2026-08-13
[18.0][IMP] account_payment_mode: unique payment mode name per company

marcos-mendez:
## Problem

Nothing prevents the creation of several `account.payment.mode` records sharing the same name in the same company. The model has no uniqueness rule of any kind, and the list view shows little beyond the name, so a user who cannot tell existing records apart just creates yet another one.

This is not theoretical. On a production database I had to clean up last week there were **four active payment modes all named `BOLETO`**, created by the same user on three different dates, with byte-identical configuration (same `payment_type`, same `payment_method_id`, same `bank_account_link`, same `fixed_journal_id`, no distinguishing field whatsoever):

| id | name | created |
|----|------|---------|
| 4  | BOLETO | 2025-01-30 |
| 6  | BOLETO | 2025-02-28 |
| 10 | BOLETO | 2025-04-09 |
| 12 | BOLETO | 2025-04-09 |

By the time it was noticed, the duplicates had spread across 70 invoices, 5 contracts and 1 sale order, and had to be merged back by hand with SQL. Reporting grouped by payment mode was silently wrong for months, because the same real-world payment mode was split across four ids.

The failure mode is self-reinforcing: the harder it is to tell the duplicates apart, the more likely the next user is to create another one.

## Solution

A Python constraint forbidding two **active** payment modes with the same name in the same company.

```python
@api.constrains("name", "company_id", "active")
def _check_name_unique(self):
```

## Design decisions

These are the points I expect reviewers to question, so let me justify them upfront.

### Why not a SQL `UNIQUE` constraint?

`name` is `fields.Char(translate=True)`, which means it is stored as a **`jsonb` column**. A `_sql_constraints` unique index would therefore compare the whole translation dictionary rather than the name the user actually sees. It would be wrong in both directions:

- `{"en_US": "Boleto"}` and `{"en_US": "Boleto", "fr_FR": "Boleto"}` are different jsonb values, so two obvious duplicates would be **accepted**;
- conversely, two records that merely happen to carry the same set of translations would be **rejected** for the wrong reason.

A Python constraint compares the value in the user's current language, which is the value the user is actually looking at when they decide whether a payment mode already exists.

### Why this does not break existing databases

`@api.constrains` is only evaluated on `create`/`write`. Installing or updating the module on a database that **already** contains duplicates does not raise anything — the module updates cleanly and the existing records keep working. Users are only stopped the next time they touch the name, company or active flag of such a record.

A SQL `UNIQUE` index would instead make `-u account_payment_mode` fail outright on every database currently affected by this bug, which is precisely the population that needs the fix. That asymmetry is the main reason for the chosen approach, and it also means no migration script is required.

### Why the comparison ignores case and surrounding blanks

`BOLETO`, `Boleto` and `boleto ` are the same name for a human being, and a rule that let them through would not solve the problem it is meant to solve — the duplicates that actually get created in the field are typed by hand at different times. `str.casefold()` is used rather than `str.lower()` so the comparison behaves correctly for non-ASCII names.

### Why the names are compared in Python rather than in the domain

Comparing with `("name", "=ilike", value)` would treat `%` and `_` inside the name as SQL wildcards, so a payment mode legitimately named `100% prepaid` would collide with unrelated records. Payment modes are a very small table, so the records of the companies concerned are read and compared in Python instead. This also sidesteps the language-fallback semantics of searching on a translated field.

### Why `sudo()`

Uniqueness has to hold company-wide. Without `sudo()`, multi-company record rules can hide part of the existing payment modes from the user performing the write, letting a duplicate through and producing an inconsistency that is even harder to diagnose. The error message names the company so the result is still intelligible to the user.

### Scope of the rule

- **Per company** — the same name remains available in a different company.
- **Archived modes are ignored** — a name freed by archiving a payment mode can be reused, and archived duplicates in existing databases stay untouched.
- **Unarchiving is checked** — `active` is in the `@api.constrains` list, so a conflict cannot be reintroduced by restoring an archived record.

## Tests

Eight tests were added, covering both what must be rejected and what must keep working:

| test | asserts |
|------|---------|
| `test_name_unique_same_company` | duplicate in the same company is rejected |
| `test_name_unique_is_case_and_space_insensitive` | `BOLETO`, `boleto`, `" Boleto "` are all rejected |
| `test_name_unique_within_same_batch` | duplicates inside a single multi-record `create()` are rejected |
| `test_name_unique_on_rename` | renaming onto an existing name is rejected |
| `test_name_unique_on_unarchive` | unarchiving into a conflict is rejected |
| `test_name_unique_per_company` | the same name in another company is **allowed** |
| `test_name_unique_ignores_archived_modes` | a name held by an archived mode is **reusable** |
| `test_rename_keeping_own_name_is_allowed` | writing a record's own name back does **not** self-conflict |

Run against Odoo 18.0 with demo data loaded: **10 tests, 0 failed, 0 errors** (the 8 new ones plus the 2 pre-existing ones).

As a negative control, the constraint was disabled and the suite re-run: the 5 tests that assert rejection fail with `ValidationError not raised`, confirming they are not passing vacuously. The remaining 3 pass either way by design — they are the guards against false positives.

The module demo data was also verified to contain no duplicate names, so it keeps loading unchanged.

`pre-commit` passes on the full set of modified files.


pedrobaeza (commented):
Why not doing it through a SQL constraint - simpler and more efficient -? If you are thinking on those already having duplicated names, that's something to fix in the DB itself IMO.

marcos-mendez:
@pedrobaeza 

I did consider it, but name is translate=True, so since 16.0 it's stored as a jsonb column. A UNIQUE (company_id, name) index would compare the whole translation dictionary, not the name the user actually sees:

{"en_US": "Boleto"}
{"en_US": "Boleto", "pt_BR": "Boleto"}

Those are two different jsonb values, so the index happily accepts them — which is exactly the duplicate we're trying to prevent. A functional index would need to pick one language (name->>'en_US'), which is arbitrary and still misses duplicates created in any other language.

Two more things the SQL constraint can't express:

- Case and blanks. In practice the duplicates users create are BOLETO, Boleto and boleto . The Python constraint compares casefold()ed and stripped names; a plain unique index would let all three through.
- active. Archiving an old payment mode and creating a new one with the same name must keep working, so it would have to be a partial index (WHERE active) on top of the functional expression.

About efficiency: the check only runs on create/write of account.payment.mode, which is a configuration model with a handful of records per company — it's not on any hot path.

---

About the FIX in the DB

Agreed that the data should be fixed, and I'm happy to add a migration script if you prefer. But the failure mode differs: when a SQL constraint can't be applied, Odoo doesn't abort the update — it logs a warning and moves on, leaving the module installed without the constraint. So precisely the databases that have the problem would end up silently unprotected. The Python constraint is evaluated on write, so those databases stay upgradable and get blocked the next time someone tries to add yet another duplicate.

If you still prefer the SQL route I can switch to a partial functional index on en_US plus an openupgrade-style pre-migration, just let me know — but I think it would be strictly weaker here.

pedrobaeza:
OK, it's reasonable what you say. No need of migration scripts. Please check CI.

marcos-mendez:
Pushed a fix for the red CI. The 7 errors were all in sibling modules, and splitting them by root cause:

### 1. `copy()` was broken — 5 of the 7 errors

`account.payment.mode` did not override `copy_data`, so duplicating a payment mode produced a record with the very same name, which the new constraint then rejected. This was not a test problem: **the standard Duplicate action was broken by this PR**, and those tests were simply the first thing to hit it.

Fixed at the source, in `account_payment_mode` itself, so no sibling test had to be touched for this:

```python
def copy_data(self, default=None):
    ...
    vals["name"] = self._get_free_name(mode.name, company_id, taken_keys)
```

The copy falls back to a free `<name> (copy)` name — the way Odoo core does for `account.journal` — and only when the original name is already taken in the target company. Copying to another company where the name is free keeps the name unchanged. Four tests were added for it: renaming, successive duplications (`(copy)`, `(copy 2)`, `(copy 3)`), an explicit `name` in `default`, and the cross-company case.

This is also one more argument for the Python constraint over a SQL `UNIQUE` one, since a unique index would break `copy()` in exactly the same way while additionally making `-u account_payment_mode` fail on every database that already contains duplicates.

### 2. Two tests reused demo data names — the remaining 2 errors

`Direct Debit of customers` and `Direct Debit of suppliers from Société Générale` both exist in `account_payment_mode/demo/payment_demo.xml`. In both tests the name is irrelevant to what is being asserted, so they were renamed.

### Commits

Split per module, as per the guidelines:

- `[IMP] account_payment_mode: unique payment mode name per company` (amended, so the `copy_data` override stays in the same logical change set rather than becoming a review-fixup commit)
- `[FIX] account_payment_partner: don't reuse a demo payment mode name`
- `[FIX] account_payment_sale: don't reuse a demo payment mode name`

No version bump on the last two, since they only touch test files.

### Verification

Ran locally on 18.0 against the whole repo, installing 17 of the 21 modules — the 4 left out (`account_payment_method_base_mode`, `account_payment_order_tier_validation`, `account_payment_order_return`, `account_payment_order_notification`) depend on other OCA repos I do not have locally, and contribute no tests:

```
0 failed, 0 error(s) of 161 tests
```

which is the same test count the CI reports, so the local run covers the same scope. `pre-commit` passes on every modified file.

@pedrobaeza the `copy()` breakage above is probably the most concrete answer to your question about the SQL constraint, so I would be glad to hear whether it changes your view.

pedrobaeza:
Please avoid all the AI text. Concrete and concise answers is the key for getting attention. I'm just wanting a green CI and correct code.

PR #1615, opened 2026-08-13
[15.0][FIX] account_banking_mandate_contact: don't override invoice contact mandate

juancarlosonate-tecnativa:
The mandate set on the invoice contact was ignored when the invoice had a shipping contact without its own contact mandate: valid_mandate_id falls back to the first valid mandate of the commercial partner, so companies with several bank accounts got their debit orders on the wrong one.

Now only an explicit contact_mandate_id on the shipping contact takes precedence, and the same criteria is applied when the mandate is stamped on the invoice for payment methods with mandate_required.

@tecnativa TT64093 
@pedrobaeza @carlos-lopez-tecnativa @carlosdauden


pedrobaeza (approved):
Some comments on the code will serve for explaining the why on that code in future references.

carlos-lopez-tecnativa (approved):

CristianoMafraJunior (approved):
LGTM

pedrobaeza:
Can you forward-port it?

/ocabot merge patch

PR #1616, opened 2026-08-25
[16.0][FIX] account_payment_purchase: purchase order payment mode not copied to the invoice

CristianoMafraJunior:
@Escodoo HT02141

When generating a vendor bill from a purchase order, the "Payment Mode"
field (`payment_mode_id`) was left empty on the invoice, even when the
purchase order had a payment mode set (e.g. Bank Slip / Boleto).

cc @kaynnan @marcelsavegnago @WesleyOliveira98

### DIFF
--- a/account_payment_purchase/models/account_move.py
+++ b/account_payment_purchase/models/account_move.py
@@ -11,10 +11,11 @@ class AccountMove(models.Model):
 
     @api.onchange("purchase_vendor_bill_id", "purchase_id")
     def _onchange_purchase_auto_complete(self):
+        old_mode = self.payment_mode_id.id or False
         new_mode = self.purchase_id.payment_mode_id.id or False
         new_bank = self.purchase_id.supplier_partner_bank_id.id or False
         res = super()._onchange_purchase_auto_complete() or {}
-        if self.payment_mode_id and new_mode and self.payment_mode_id.id != new_mode:
+        if old_mode and new_mode and old_mode != new_mode:
             res["warning"] = {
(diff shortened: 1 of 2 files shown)


kaynnan (approved):

account_payment_purchase/models/account_move.py
```diff
@@ -11,10 +11,11 @@ class AccountMove(models.Model):
 
     @api.onchange("purchase_vendor_bill_id", "purchase_id")
     def _onchange_purchase_auto_complete(self):
+        old_mode = self.payment_mode_id.id or False
```
pedrobaeza, line 14: The change is this file shouldn't be needed. Only the other change.

pedrobaeza (changes requested):

account_payment_purchase/models/account_move.py
CristianoMafraJunior: `account.move's` own `_onchange_purchase_auto_complete` (core `purchase`) does `self.update(invoice_vals)` with everything `_prepare_invoice()` returns. Since `payment_mode_id` is now in that dict, it already gets applied there before this module's mismatch check runs — so the comparison always sees equal values and the "different payment mode" warning stops firing. That's what breaks `test_from_purchase_order_invoicing` if only `purchase_order.py` is changed. The `account_move.py` change snapshots the old value before `super()` runs so the comparison still works. Verified by reverting just that file locally — the test fails without it.

WesleyOliveira98 (approved):

PR #1617, opened 2026-08-31
[18.0][FIX] account_payment_order: fix company currency total aggregation.

Berezi:
total_company_currency uses company_currency_id as its currency field, but the currency field was not included in the list view.
This caused the total aggregation to display "No currency provided".
This change adds company_currency_id as an invisible column so the monetary total can be properly rendered.
Captura desde 2026-08-31 13-04-35


oihane (approved):
LGTM

CristianoMafraJunior (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1618, opened 2026-09-04
[16.0][IMP] account_payment_order: Use allow_out_payment on account_payment_order

adasatorres-tecnativa:
@Tecnativa TT64381
@pedrobaeza 

When enabled:

- For outbound payment orders, ensure`partner_bank_id` has `allow_out_payment` enabled.

Superseed #1340 

Refactor `draft2open` by extracting its logic into several private methods to reduce its cyclomatic complexity and pass the C901 'AccountPaymentOrder.draft2open' is too complex (17) pre-commit check.


pedrobaeza:
This should be mandatory, not something to be chosen at company level IMO.

adasatorres-tecnativa:
> This should be mandatory, not something to be chosen at company level IMO.

So, shouldn't there be any configuration option to disable this validation?

adasatorres-tecnativa:
@pedrobaeza, 

The tests in other addons are failing because the new validation is not taken into account when running the tests, and bank_account_id is not being added to the journal.

I can think of two options: either update the tests in the other addons as part of this PR, using separate commits, or create a bypass to skip this validation.

Which option do you think would be the best one to go with?

cristina-hidalgo-tecnativa (changes requested):
Review code and tested locally

account_payment_order/models/account_move_line.py
```diff
@@ -71,10 +71,10 @@ def _prepare_payment_line_vals(self, payment_order):
             # in this case
         if payment_order.payment_type == "outbound":
             amount_currency *= -1
-        partner_bank_id = self.partner_bank_id.id or first(self.partner_id.bank_ids).id
+        partner_bank = self.partner_bank_id or first(self.partner_id.bank_ids)
```
cristina-hidalgo-tecnativa:
```suggestion
        partner_bank_id = self.partner_bank_id.id or first(self.partner_id.bank_ids).id
```

account_payment_order/models/account_payment_line.py
```diff
@@ -1,8 +1,10 @@
 # © 2015-2016 Akretion - Alexis de Lattre <alexis.delattre@akretion.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, fields, models
+from odoo import _, api, fields, models, modules
```
cristina-hidalgo-tecnativa:
```suggestion
from odoo import _, api, fields, models
```

account_payment_order/models/account_payment_line.py
```diff
@@ -1,8 +1,10 @@
...
 from odoo.exceptions import UserError
+from odoo.fields import first
+from odoo.tools import config
```
cristina-hidalgo-tecnativa:
```suggestion

```

account_payment_order/models/account_payment_line.py
```diff
@@ -249,3 +251,23 @@ def action_open_business_doc(self):
         if not self.move_line_id:
             return False
         return self.move_line_id.action_open_business_doc()
+
+    def _check_bank_allows_out_payments(self):
```
cristina-hidalgo-tecnativa, line 254:
```suggestion
    def _check_bank_allows_out_payments(self):
        for line in self:
            bank = line.partner_bank_id or first(
                line.partner_id.bank_ids.filtered(
                    lambda b: not b.company_id or b.company_id == line.company_id
                )
            )
```

the guard disables the rule for every other suite (dependent modules' tests stop reflecting production) and current_test is `False`/`True` (not a test instance) during module loading / `post_init_hook` / while another module installs → `.test_module` can raise `AttributeError`.

Removing it will probably need small fixes in the dependent modules' test setups (e.g. `"allow_out_payment": True` on the `res.partner.bank` records they create): in a quick local run I saw ~6 failures in `account_banking_sepa_credit_transfer` and none in `account_banking_sepa_direct_debit` (inbound). The `.filtered()` also mirrors core's per-company filter on `bank_ids` (`account_move._compute_partner_bank_id`).

Image

account_payment_order/models/account_payment_order.py
```diff
@@ -6,8 +6,9 @@
 
 import base64
 
-from odoo import _, api, fields, models
+from odoo import _, api, fields, models, modules
```
cristina-hidalgo-tecnativa:
```suggestion
from odoo import _, api, fields, models
```

account_payment_order/models/account_payment_order.py
```diff
@@ -6,8 +6,9 @@
...
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools import config
```
cristina-hidalgo-tecnativa:
```suggestion

```

account_payment_order/models/account_payment_order.py
```diff
@@ -340,17 +324,11 @@ def draft2open(self):
             for payline in order.payment_line_ids:
                 try:
                     payline.draft2open_payment_line_check()
+                    if self.payment_type == "inbound":
```
cristina-hidalgo-tecnativa:
```suggestion
                    if order.payment_type == "outbound":

```

`self.payment_type` is evaluated inside for order in self:, so `draft2open()` on more than one order raises Expected singleton (list action, tier validation) — use order.. And the per-line check of `partner_bank_id` (the account that receives the transfer) belongs to outbound; for an inbound direct debit the mandate is what authorises it, not "Send Money" on the customer account.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -29,9 +29,7 @@ def setUpClass(cls, chart_template_ref=None):
                     (
                         0,
                         0,
-                        {
-                            "acc_number": "TEST-NUMBER",
-                        },
+                        {"acc_number": "TEST-NUMBER", "allow_out_payment": True},
```
cristina-hidalgo-tecnativa, line 32: Image

`allow_out_payment` is `False` by default on every `res.partner.bank`, so every existing payment order will fail draft2open right after the upgrade until someone ticks "Send Money" on each account by hand.

A migration script would avoid breaking existing databases.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -72,13 +70,28 @@ def setUpClass(cls, chart_template_ref=None):
...
         cls.invoice_02 = cls._create_supplier_invoice(cls, "F1243")
         cls.bank_journal = cls.company_data["default_journal_bank"]
+        cls.company_partner_bank = cls.env["res.partner.bank"].create(
+            {
+                "acc_number": "company_1234",
+                "partner_id": cls.company.partner_id.id,
+                "allow_out_payment": True,
+            }
+        )
+        cls.bank_journal.bank_account_id = cls.company_partner_bank.id
```
cristina-hidalgo-tecnativa:
```suggestion

```

With the outbound check moved to the payment lines, company_partner_bank_id is no longer validated, so `cls.company_partner_bank` and the `cls.bank_journal.bank_account_id` assignment are dead setup now — could be dropped.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -227,89 +240,20 @@ def order_creation(self, date_prefered):
...
-        wizard_analytic.populate()
-        self.assertFalse(wizard_analytic.move_line_ids)
-
     def _line_creation(self, outbound_order):
         vals = {
             "order_id": outbound_order.id,
             "partner_id": self.partner.id,
             "currency_id": outbound_order.payment_mode_id.company_id.currency_id.id,
             "amount_currency": 200.38,
             "move_line_id": self.invoice.invoice_line_ids[0].id,
```
cristina-hidalgo-tecnativa, line 247:
```suggestion
            "amount_currency": 200.38,
            "move_line_id": self.invoice.invoice_line_ids[0].id,
            "communication": "TEST-COMM",
```

Without a communication, `draft2open_payment_line_check()` raises "Communication is empty" before the `allow_out_payment` check is reached, so the test below never actually exercises it (and would fail once switched to `assertRaisesRegex(..., "Send Money"))`.

Image

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -643,3 +561,46 @@ def test_action_open_business_document(self):
...
+        payment_order = self.env["account.payment.order"].search(self.domain)
+        payment_order.write(
+            {
+                "journal_id": self.bank_journal.id,
+            }
+        )
+        payment_order.company_partner_bank_id.allow_out_payment = False
+        # Add to payment order using the wizard: error raised
+        with self.assertRaises(UserError):
+            payment_order.draft2open()
```
cristina-hidalgo-tecnativa, line 573:
```suggestion
    def test_check_allow_out_payment(self):
        """Out payments are blocked when "Send Money" is disabled on the
        destination (partner) bank account, and allowed once it is enabled.
        """
        self.invoice.action_post()
        self.partner_bank.allow_out_payment = False
        for bank in self.partner.bank_ids:
            bank.allow_out_payment = False
        self.env["account.invoice.payment.line.multi"].with_context(
            active_model="account.move", active_ids=self.invoice.ids
        ).create({}).run()
        payment_order = self.env["account.payment.order"].search(self.domain)
        payment_order.write({"journal_id": self.bank_journal.id})
        payment_order.payment_line_ids.partner_bank_id = self.partner_bank.id
        with self.assertRaisesRegex(UserError, "Send Money"):
            payment_order.draft2open()
        self.partner_bank.allow_out_payment = True
        payment_order.draft2open()
        self.assertEqual(payment_order.state, "open")
```
The old test disabled `allow_out_payment` on `company_partner_bank_id` and checked the order was blocked — i.e. it gated on the company's own account. Now that the check is on the payment lines, it disables it on the line's `partner_bank_id`, asserts the real message (`assertRaisesRegex`, not any `UserError`), and adds the positive path (enable the flag → `draft2open` works → state `open`).

cristina-hidalgo-tecnativa (changes requested):

adasatorres-tecnativa:
@cristina-hidalgo-tecnativa, since allow_out_payment is a field added by base, I’ll add the migration script directly to the base migration in the OpenUpgrade repository.

account_payment_order/migrations/16.0.1.0.0/post-migration.py
cristina-hidalgo-tecnativa, line 1: Scripts in `migrations/X.Y.Z/` only run when `X.Y.Z` is in (installed version, manifest version]. Existing 16.0 DBs are already at `16.0.1.14.0`, so `16.0.1.0.0` never runs. It has to be `migrations/<new version>/` with the manifest bumped to match (e.g. `16.0.1.15.0`).

account_payment_order/tests/test_payment_order_transfer_journal.py
```diff
@@ -17,6 +17,14 @@ def setUpClass(cls):
...
         )
         cls.bank_journal = cls.company_data["default_journal_bank"]
+        cls.company_partner_bank = cls.env["res.partner.bank"].create(
+            {
+                "acc_number": "company_1234",
+                "partner_id": cls.env.company.partner_id.id,
+                "allow_out_payment": True,
+            }
+        )
+        cls.bank_journal.bank_account_id = cls.company_partner_bank.id
```
cristina-hidalgo-tecnativa:
```suggestion

```

This was added for the old `company_partner_bank_id` check, which is now gone (the check moved to the payment lines). The test passes without it — dead setup.

account_payment_order/tests/test_payment_order_outbound.py
```diff
@@ -643,3 +554,43 @@ def test_action_open_business_document(self):
...
+                "payment_type": "outbound",
+                "payment_mode_id": self.mode.id,
+                "journal_id": self.bank_journal.id,
+                "description": "order with manual line",
+            }
+        )
+        payment_line_1 = self._line_creation(outbound_order)
+        payment_line_1.partner_bank_id = self.partner_bank.id
+        # Add to payment order using the wizard: error raised
+        with self.assertRaises(UserError):
```
cristina-hidalgo-tecnativa:
```suggestion
        with self.assertRaisesRegex(UserError, "Send Money"):

```

`assertRaises(UserError)` passes on any `UserError`. `assertRaisesRegex(..., "Send Money")` pins it to the `allow_out_payment` check, consistent with `test_check_allow_out_payment` above. The `# Add to payment order using the wizard` comment is stale — there's no wizard here, the line is created directly.

cristina-hidalgo-tecnativa (changes requested):

account_payment_order/migrations/16.0.1.15.0/post-migration.py
carlos-lopez-tecnativa: Migration scripts set to 1.0.0 are executed when you migrate from V15, for example. In this case, I noticed two migration scripts with the same content, so please keep only one.

If you expect this code to be executed in existing databases where this module is already installed, keep only the migration script with 16.0.1.15.0, as this is executed when migrating the database from an older version.

However, if you expect the migration script to be executed only during the migration and not on existing databases, keep 16.0.1.0.0.

account_payment_order/migrations/16.0.1.14.0/post-migration.py
carlos-lopez-tecnativa, line 1: This migration script is never executed because you didn't change the version in the manifest. If you want the migration script to be executed on existing databases, increment the version in the manifest explicitly. Otherwise, keep only the `16.0.1.0.0` migration script, based on my previous comment.

carlos-lopez-tecnativa (changes requested):
I noticed that in this commit https://github.com/OCA/bank-payment/pull/1618/changes/30a62a162a268424654d418f7bf6ec9064d630d6, you changed the demo data, which was reverted in this commit: https://github.com/OCA/bank-payment/pull/1618/changes/b89c76977ebeaa5fb13b52007246fb38c6eaa0b4. However, in this commit, you changed data in two modules.

Could you please fix this? If the demo data change is not necessary, please squash the commits.

Try to modify a single module in each commit.

account_payment_order/migrations/16.0.1.14.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
...
+def migrate(env, version):
+    env.cr.execute(
+        """
+        UPDATE res_partner_bank bank
+           SET allow_out_payment = TRUE
+          FROM account_payment payment
+         WHERE payment.partner_bank_id = bank.id
+           AND payment.payment_type = 'outbound'
+           AND payment.payment_order_id IS NOT NULL
+        """
```
carlos-lopez-tecnativa, line 14: A cosmetic change to make it more readable:
```suggestion
        UPDATE res_partner_bank bank
           SET allow_out_payment = TRUE
        FROM account_payment payment
        WHERE payment.partner_bank_id = bank.id
           AND payment.payment_type = 'outbound'
           AND payment.payment_order_id IS NOT NULL
        """
```

carlos-lopez-tecnativa (changes requested):
Please change the version manually in the manifest to `16.0.1.15.0`.

The OCA Bot does this automatically according to the command launched (patch, minor, or major), but when you need a migration script, it is better to set the version explicitly.

adasatorres-tecnativa:
ping @carlos-lopez-tecnativa

account_payment_order/models/account_payment_line.py
```diff
@@ -249,3 +250,22 @@ def action_open_business_doc(self):
...
+                raise UserError(
+                    _(
+                        'The option "Send Money" is not enabled on the bank '
+                        "account %(bank_account)s of partner %(partner)s."
+                    )
+                    % {
+                        "bank_account": bank.acc_number,
+                        "partner": line.partner_id.name,
+                    }
+                )
```
carlos-lopez-tecnativa, line 278: Could we implement a strategy to validate this only when it is in this module's own tests, or when it is not in test mode? What do you think @pedrobaeza

carlos-lopez-tecnativa (commented):
LGTM. However, I think this change could require more adaptations in other repositories where `account_payment_order` is a dependency, especially in tests where `allow_out_payment` is not set. Since you adapted other modules in this repository for the same reason, please keep this in mind if you detect any failures related to this.

account_payment_order/models/account_payment_line.py
pedrobaeza: Yes, a preventive measure in this version (which has been started with because it was the original proposition, but we can go directly to v18 to avoid side effects on old versions), is to put the `if` with the current module testing thing.

adasatorres-tecnativa:
ping @carlos-lopez-tecnativa @pedrobaeza

account_payment_order/models/account_payment_order.py
```diff
@@ -340,17 +324,19 @@ def draft2open(self):
...
+                    if order.payment_type == "outbound":
+                        payline.with_context(
+                            **{
+                                "skip_payment_order_out_payments": (
+                                    config.get("test_enable")
+                                    and modules.module.current_test.test_module
+                                    != "account_payment_order"
+                                )
+                            }
+                        )._check_bank_allows_out_payments()
```
carlos-lopez-tecnativa: Why don't you put this code directly into the `_check_bank_allows_out_payments` method?

carlos-lopez-tecnativa (commented):
Just an existential doubt.

account_payment_order/models/account_payment_order.py
adasatorres-tecnativa: Honestly, I don’t really know. I just decided to pass it through the context and that’s it. There’s no real explanation for it, to be honest. 😅

carlos-lopez-tecnativa (approved):
Thinking out loud: now that the validation is performed only in test mode in this module, are the changes in the other modules really necessary?

Anyway, LGTM.

account_payment_order/models/account_payment_line.py
```diff
@@ -249,3 +251,27 @@ def action_open_business_doc(self):
         if not self.move_line_id:
             return False
         return self.move_line_id.action_open_business_doc()
+
+    def _check_bank_allows_out_payments(self):
+        if (
+            config.get("test_enable")
+            and modules.module.current_test.test_module != "account_payment_order"
+        ):
```
cristina-hidalgo-tecnativa, line 260:
```suggestion
        if (
            config.get("test_enable")
            and getattr(modules.module.current_test, "test_module", None)
            != "account_payment_order"
        ):

```
Out of curiosity, I wanted to check whether the risk of current_test not always being a test instance — and `.test_module` blowing up because of it — was real or just theoretical.

I brought the branch to local, installed the module, and created a plain outbound payment order in draft.

Image

One line, verified supplier account, nothing special, just a real order to call `draft2open()` on.

Started an odoo shell with `test_enable` on, like a process would with `--test-enable` and no test running at that moment, and called `draft2open()`:

Image

`current_test` was `False` (its default), not a test instance, so `.test_module` blows up right there — confirming the risk is real, not theoretical.

Adding this instead of the direct access, and the same scenario no longer fails: the check is skipped as intended, `draft2open()` proceeds normally.

In my opinion it would be worth adding: in a CI/CD pipeline like OCA's own (GitHub Actions, installing modules with `--test-enable`), this is an error that doesn't always show up, and it's hard to catch precisely because it depends on the exact moment `draft2open()` runs relative to whatever test happens to be executing.

To replicate this I used the following script (always assuming you have an outbound payment order in draft state beforehand, and plugging in its id, 24 in my case):

```
docker compose run --rm odoo odoo shell -d devel <<'EOF'
from odoo.tools import config
config["test_enable"] = True
o = env["account.payment.order"].browse(24) # change this id for yours
print("state:", o.state, "| payment_type:", o.payment_type)
o.draft2open()
EOF
```

cristina-hidalgo-tecnativa (changes requested):

account_payment_order/models/account_payment_line.py
adasatorres-tecnativa: It’s a bit of an unusual case, but it’s true that it could cause issues if the Odoo configuration enables test mode by default.

account_payment_order/migrations/16.0.1.14.0/post-migration.py
cristina-hidalgo-tecnativa, line 1: If I understood correctly — there are two different scenarios for a migration folder version: `16.0.1.0.0`-style, meant to fire only on a major-version jump (e.g. 15.0 → 16.0 via OpenUpgrade), or a version above what's currently installed, meant to fire on a normal in-place update within 16.0 for databases that already have the module. Right now it's set to `16.0.1.14.0` for both the manifest and the folder, which doesn't quite land on either: it's not `16.0.1.0.0`, and it's not above what any existing 16.0 install already has, so there's no version gap for an in-place update to cross.

I tested it locally (fresh install at `16.0.1.14.0`, then `-u`) and didn't see it run:

Image

Wouldn't 16.0.1.15.0 be the right version here, for both the manifest and the migration folder, given that the module's current version is 16.0.1.14.0?

Image

cristina-hidalgo-tecnativa (changes requested):

account_payment_order/migrations/16.0.1.14.0/post-migration.py
carlos-lopez-tecnativa: @cristina-hidalgo-tecnativa, you are right. The current version of this module is `16.0.1.14.0`: https://github.com/OCA/bank-payment/blob/c6d2c92d8203a4f5f9fcd1d089b546a00a392b38/account_payment_order/__manifest__.py#L12
I imagine that, since this PR(or the superseded original P) was created, some other PRs had already been merged and the version was incremented.

In this PR, the current outdated version is `16.0.1.13.2`. So, @adasatorres-tecnativa, please rebase the branch and increment the manifest version and the migration folder to `16.0.1.15.0` or `16.0.1.14.1`, depending on how you consider the level of the changes.

cristina-hidalgo-tecnativa (approved):
Code reviewed and tested locally and on runboat. LGTM!

account_payment_order/migrations/16.0.1.15.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env.cr.execute(
+        """
+        UPDATE res_partner_bank bank
+           SET allow_out_payment = TRUE
```
omar7r, line 9: I think that, only should do it in draft payment order, because, a customer could be do a payment for an incorrect account previous, he know it, and now we set this bank account as confiable and the error can be repeated. In an old account the same

PR #1623, opened 2026-09-18
[19.0] [FIX] account_payment_order: fix view

luisDIXMIT:
In the last PR I fix customer invoices view but I make a mistake and vendor bills view was broken. I think this is the best approach, I tried many different options and this is what works.

I suppose that the error comes from that in the orignal view, payment_referend field is duplicated, one is shown for customer invoice view and the other is shown for vendor bill view.


luisDIXMIT:
@pedrobaeza @etobella could you have a look and tell me if it is correct or the code can be improved? Thanks.

pedrobaeza:
Can you put screenshots of before/after of both types of invoices (and check also on regular journal entries)?

luisDIXMIT:
> Can you put screenshots of before/after of both types of invoices (and check also on regular journal entries)?

I add screenshots on the top comment. Thanks!

pedrobaeza (approved):
/ocabot merge patch

